PR #6, opened 2015-08-03
xBRZ 1.4 filter update

intact:


DoctorWho11:
Could you include the wx platform too if possible. If it checks out I'll merge it in. May as well get all supported interfaces included in the xBRZ update

PR #10, opened 2016-04-27
Kludges for ffmpeg 3.0

SwooshyCueb:
FFmpeg 3.0 removed a significant chunk of legacy code VBA-M was relying on. I hammered on it a bit, got it compiling again, and got audio recording working again. Video recording doesn't work with ffmpeg 3.0 still, however.

This is probably not at all the way this should be done, but it at least provides a place to start. None of this code should come into effect if compiled with older versions of ffmpeg.


DoctorWho11:
Thanks, I was working on this but due to recent new job that has been sucking the life out of me haven't been able to do much. I'll merge it in

PR #11, opened 2016-06-07
Add some header and resource files to CMakeLists.txts (Core+WX)

SwooshyCueb:
This makes these files show up in generated CodeBlocks projects, which makes life easier for folks who import into QtCreator.

I've only done this for the WX build and some stuff in the main CMakeLists.txt, as I currently cannot test any changes I make to other builds.


DoctorWho11:
Looks alright, merging in.

PR #14, opened 2016-07-22
Fixing U16 and U8 Parameter Types

MrJman006:
It seems like most of the types are moving to fixed width integer types and this file wasn't fully converted.


DoctorWho11:
Aye, I wasn't quite done. I ended up taking a small break due to life taking over. I'm going to go about fixing more. Will merge and expect to see more work tomorrowish.

PR #15, opened 2016-07-29
Remove typedef'd standard int types

ElementW:
You seem to be struggling with removing `typedef`'d integer types. What tool are you using ?
It literally took me 2 minutes of `find ./ -type f -exec sed -i -e 's/u32/uint32_t/g' {} \;` and half an hour of fixing includes and names wrongly getting caught by the regexp.


DoctorWho11:
I was going through line by line, making sure I was not screwing anything up, there are some oddities in the source tree that needed to be looked at and plus real life has been rather screwing me up a bit, namely work

PR #18, opened 2016-08-08
Small change: make `.clang-format` use C++11 Standard

jayands:
I noticed this very small discrepancy, and so I changed it.  The `CMakeLists` is already using C++11 standard, but the `.clang-format` was not.


DoctorWho11:
Good catch. Thanks.

PR #19, opened 2016-08-09
Fix for T2

jayands:
see: http://dev.vba-m.com/T2

> Every time I launch a server it defaults to the locahost IP instead of my
> actual IP. This is preventing me from linking with friends over Hamachi. When
> they enter my IP it errors on their end and they cannot connect. [report taken
> from <https://sourceforge.net/p/vbam/bugs/219/>]

Also, by chance, added the build folder to `.gitignore` so as to not pollute the
git repo.  I meant to do them as separate commits, but I forgot to do so.

Signed-off-by: Jonathan Sifuentes <jayands2k11@ymail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ src/wx/cmd-evtable.h
 src/wx/cmdhandlers.h
 src/wx/cmdtab.cpp
 src/wx/wxvbam.xrs
+build/
+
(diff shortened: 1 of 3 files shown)


DoctorWho11:
Thanks, merged in. Will close the ticket.

PR #23, opened 2016-09-20
improve OpenAL support on Mac and make it default

rkitover:
Fix the dynamic library path for OpenAL to the framework included with
OS X.

Make option initialization set the AudioAPI to OpenAL on mac by default,
since SDL sound is currently completely broken.

There is a lot of stuttering, will have to look at that next.


DoctorWho11:
Thanks and merged in.

PR #34, opened 2016-11-03
Mac OS X 10.7 (Lion) build and runtime support

rkitover:
Make a custom cstdint.h header file that includes <tr1/cstdint> if
<cstdint> is not available, because the clang 3.x used on Lion does not
have it. Change all references to <cstdint> to use it instead.

Add missing OpenGL header for older OS X in sdl/SDL.cpp .

When calling HiDPI methods, use respondsToSelector: to check if the
methods are available first.

Fix the bundling/linking script to support multiple copies of the same
dylib with different versions. Necessary to include both the Lion system
libpng and the brew libpng. Including the system libpng is necessary
because it is removed in later versions of OS X.

### DIFF
--- a/src/System.h
+++ b/src/System.h
@@ -1,7 +1,7 @@
 #ifndef SYSTEM_H
 #define SYSTEM_H
 
-#include <cstdint>
+#include "common/cstdint.h"
 
 #ifndef __LIBRETRO__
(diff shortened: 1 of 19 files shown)


DoctorWho11:
Looking good. Merged in.

PR #49, opened 2016-12-09
Fix DSYSCONF_INSTALL_DIR (SDL)

Mystro256:
I'm not sure if this was changed in cmake recently, but "SYSCONFDIR" is incorrect. It should be SYSCONF_INSTALL_DIR.

As well CMAKE_INSTALL_PREFIX refers to /usr or /usr/local in linux, so installing it into ${CMAKE_INSTALL_PREFIX}/${SYSCONFDIR} would be incorrect. I've wrapped it in the existing IF( WIN32 ) to preserve the installation for windows, a window dev should fix that if it's incorrect.


DoctorWho11:
Looks good, thanks.

Mystro256:
I missed two files, see pull #50

Issue #42, opened 2016-11-30
Diagonal movement from vertical to horizontal axis doesn't work


autofire372:
Attempting to go directly from a vertical movement to a horizontal one (i.e. going from Up to Up+Right) doesn't work correctly, as the horizontal input doesn't register.

autofire372:
Whoops, just noticed the proper issue tracker.

DoctorWho11:
This is the proper issue tracker now. Source forge issue tracking is shit to be honest

autofire372:
No kidding.

DoctorWho11:
Working on a script to try and export the existing issues from there to here. And then shutting down the sf.net project page once and for all. Anyways thanks for the report. Will look into it asap.

rkitover:
Is this with keyboard or joystick? Which interface? (Wx, GTK, MFC or SDL)

autofire372:
Joystick on Wx.

rkitover:
And of course I forgot to ask, which OS and which type of joystick?

And is this the analog stick or the directional pad?

autofire372:
Xbox One pad on Linux, and the issue is with the controller's d-pad.

rkitover:
Ok thank you, we will try to look at this.

rkitover:
Works fine for me just scrolling the cursor on the map in advance wars, which game are you trying this with?

autofire372:
Tried with Pokemon Blue and with Pokemon Fire Red; both exhibit this incorrect behavior.

DoctorWho11:
I'll try it out when I get home. I have an Xbox one controller. Last I tried using the xbone controller on Linux it was playing Metroid and it was fine for me.

autofire372:
For the record, I just tested, and this issue happens on Windows too.

rkitover:
@autofire372 does #53 (in master) fix your issue?

autofire372:
I can't tell, since master is segfaulting on launch on my end.

autofire372:
After re-fetching the source, I rebuilt the program, and this issue has been resolved; issue #46 is still present, though.

PR #53, opened 2016-12-18
Fix wx joystick hat diagonals

drysyle:
Fixed an issue I was having with my gamepad where VBA wouldn't recognize that a d-pad diagonal was released until the d-pad was released entirely (which I think is the same as issue #42).

The value in ev.GetControlValue() comes from the SDL function function SDL_JoystickGetHat(), which returns a bitmask of the directions currently pressed, so I just used that value directly instead of a bunch of specious logic with the WXJB enums.


DoctorWho11:
Looks good.

PR #60, opened 2016-12-30
Various fixes

Mystro256:
I was cleaning up some gcc warnings and I found a couple other issues. The commit messages should be self-explanatory, but let me know if clarification is needed.

I've only tested building it on Linux, so please let me know if it breaks anything else.


DoctorWho11:
I'll take a look at it when I get home. But so far looks good.

DoctorWho11:
Looking good, tested on mxe and on a windows vm and plays nice enough.

PR #68, opened 2017-01-30
Queue refresh after drawing (required for wayland)

Mystro256:
I noticed that compiling and running on Wayland (an alternative to XOrg/X11 for BSD and Linux), the screen would never refresh automatically unless the window is resized. Requesting a Refresh() following DrawArea() and DrawOSD() does the trick.

Although with that said, Wayland seems to requires drawing with Cairo, as the GUI will segfault without it, so I've wrapped this in ```#ifndef NO_CAIRO``` to avoid any unintended issues with other systems.

Side note, I noticed a comment in ```CairoDrawingPanel::DrawArea()``` about drawing with cairo being very slow for WXMSW. Perhaps this is also fixes that issue, which could be required for cairo to work correctly, but I'm not sure.


rkitover:
@Mystro256 @DoctorWho11 

So according to the docs here:

http://docs.wxwidgets.org/2.8/wx_wxwindow.html#wxwindowrefresh

this queues a repaint event for the window, so what your change does is, in effect, make a loop that sends another paint event after every paint event. This may be **OK** if we can't find a more elegant solution.

What we would ideally like to do is check in `OnIdle()` if the emulator system has a frame ready to be drawn that is different from the previous frame, and only then send a `Refresh()` to the panel.

Mystro256:
I agree; this should work as an interim solution, but doing a check in ```OnIdle()``` seems like the more sane solution.

Mystro256:
Hold on, I'm going to recreate this pull request on a branch...

PR #70, opened 2017-02-11
Some fixes for MSVC.

edorax:
Tested for VS2012 & MXE.


rkitover:
I added a couple of tiny fixes, other than that 👍 from me!

I was going to look at fixing VS cmake generation at some point, so thank you very much for doing this!

I had some questions, since I have only used VS a couple of times before, how do you build a cmake project such as this in VS? Do you download the mingw tarballs for the deps and use the cmake GUI to set paths to them, or some other way? Can we automate this (perhaps by including everything in the dependencies submodule?)

I was going to play with this, essentially full free VS:

https://www.visualstudio.com/vs/community/

rkitover:
@edorax one more thing, I cleaned up the `dependencies/msvc` checking code a little bit, can you please check that everything still works?

edorax:
Your changes are OK, verified for MXE and VS2012.

edorax:
@rkitover
It's a long story about how to build vba-m with cmake & VS.
The most libraries needed was installed(C:\opt is a good place) or built before I met vba-m.
These includes zlib, libpng, SDL2 and wxWidgets 3.0.2.
cmake will check for yasm, I downloaded yasm.exe and put it in the search path(C:\opt\bin),
but VS does not assemble the *.asm files, to works around this, I disabled the ASM optimization.
Manually built wxrc.exe from the source(nmake -f makefile.vc) in Developer Command Prompt for VS and put it in C:\opt\bin .
I found a copy of zip.exe(need by wxrc.exe?) in the installation of the CodeBlocks and put it C:\opt\bin .

Since many other things are not install on my box, I disabled some corresponding options in cmake command line to work around.

The command lien I used to generate the VS solution/projects with cmake:
cmake -G "Visual Studio 11" -DwxWidgets_ROOT_DIR="D:\wdir-svn\MiscTools\wxWidgets\wxWidgets-3.0.2" -DCMAKE_LIBRARY_PATH:STRING="C:\opt\common_libs\lib" -DZLIB_ROOT:STRING="C:\opt\common_libs" -DPNG_PNG_INCLUDE_DIR:STRING="C:\opt\common_libs\include" -DSDL2_PATH:STRING="C:\opt\SDL2" -DSDL2_STATIC:BOOL=ON -DENABLE_LINK:BOOL=OFF -DENABLE_OPENAL:BOOL=OFF -DENABLE_NLS:BOOL=OFF -DENABLE_ASM_CORE:BOOL=OFF -DENABLE_ASM_SCALERS:BOOL=OFF ../vba-m/visualboyadvance-m

rkitover:
@edorax 

Awesome, thank you for the info, if and when I have time, I will try to replicate the process and see if we can automate it a bit more as well.

As for YASM, there is no explicit YASM support in the cmake files yet, I will need to add that.

I told @DoctorWho11 that everything is good to merge.

You can find us on `#vba-m` on freenode if you use irc.

Issue #51, opened 2016-12-10
Emulation is too fast


Mystro256:
I was previously using a personal build based on SVN snapshots but moved to this git version recently. I'm using Fedora (Linux) and I seem to get abnormally really fast emulation speeds, almost as if I was constantly holding down the speed key. Hitting the speed key seems to make things go even faster. It seems like the frame limiter is messed up and games run from 500% to 1000% the normal speed and about 1500% with the speed key hit. I can't reproduce this with OpenGL with WX, but this is always reproducible with GTK, WX+Simple, or WX+Cairo.

I'm going to try to do a git bisect to figure out what happened, but if anyone can point me in the right direction, that may help.

Mystro256:
Found it, it's caused by ef62b1977516a8ea145e9795a434396762d10ce1

I'm thinking it's due to SDL backend not syncing with the audio? I haven't looked too into it. I'll work around the issue for now.

DoctorWho11:
Sorry for the delay. Have you tried with another audio backend like openal?

Mystro256:
I'll check and get back to you

DoctorWho11:
I'll be awaiting.

Mystro256:
Sorry for the delay, I can reproduce this with both SDL and openal.
Manually reverting ef62b19 removes the issue from both as well.

DoctorWho11:
I can't reproduce this issue on my linux setup. But I'll keep trying, even if I have to setup a fedora machine just to see.

ichigo-0:
I'm on debian testing (though I only update occasionally), and have this too with master and Beta-3. However, the version tagged v2.0.0 (1eb768...) seems to work fine.

(In order to compile that version, I had to use newer versions of src/sdl/expr-lex.cpp, expr.cpp, and expr.ypp; or needed at least one of those.)

rkitover:
@ichigo-0 is this with all ROMs you tried or only specific ones? Which ROMs do you use?

rkitover:
@ichigo-0 ah yeah, there is the "throttle" option in vbam.ini, the default setting is `100`.

rkitover:
@Mystro256 and can you also try making sure `throttle=100` is set in your `vbam.ini` and try master?

Either that, or move `~/.vbam` out of the way while testing.

ichigo-0:
[...]

PR #72, opened 2017-02-12
Queue refresh after drawing (required for wayland)

Mystro256:
Recreation of Pull Request #68:

Original description:

> I noticed that compiling and running on Wayland (an alternative to XOrg/X11 for BSD and Linux), the screen would never refresh automatically unless the window is resized. Requesting a Refresh() following DrawArea() and DrawOSD() does the trick.
> 
> Although with that said, Wayland seems to require drawing with Cairo, as the GUI will segfault without it, so I've wrapped this in #ifndef NO_CAIRO to avoid any unintended issues with other systems.
> 
> Side note, I noticed a comment in CairoDrawingPanel::DrawArea() about drawing with cairo being very slow for WXMSW. Perhaps this also fixes that issue, which could be required for cairo to work correctly, but I'm not sure.

Quote from last relevant comment:
@rkitover said:
> So according to the docs here:
> 
> http://docs.wxwidgets.org/2.8/wx_wxwindow.html#wxwindowrefresh
> 
> this queues a repaint event for the window, so what your change does is, in effect, make a loop that sends another paint event after every paint event. This may be OK if we can't find a more elegant solution.
> 
> What we would ideally like to do is check in OnIdle() if the emulator system has a frame ready to be drawn that is different from the previous frame, and only then send a Refresh() to the panel.

### DIFF
--- a/src/wx/panel.cpp
+++ b/src/wx/panel.cpp
@@ -1424,6 +1424,10 @@ void DrawingPanelBase::PaintEv(wxPaintEvent& ev)
 
     DrawArea(dc);
     DrawOSD(dc);
+#ifndef NO_CAIRO
+    if (!paused)
+        GetWindow()->Refresh();
+#endif
 }
 
 void DrawingPanelBase::EraseBackground(wxEraseEvent& ev)


rkitover:
@Mystro256 

This is the issue with OpenGL on Wayland:
http://trac.wxwidgets.org/ticket/17702

rkitover:
@Mystro256 

What happens without this patch, does the window just stay black? Does it stay on the first frame drawn?

Mystro256:
The wxGTK patch or my commit? Without calling refresh(), it just draws the
first frame and does not redraw unless the window is resized.

This wxgtk bug seems unrelated as I'm not getting a segfault, but it would
explain why trying to use OpenGL was crashing for me (opposed to simple or
cairo).

DoctorWho11:
I think the plan is going to replace wxglcanvas with something like egl.

rkitover:
@Mystro256 

I have an idea I used in my useless mac Quartz2D renderer, I will make a branch and link it here.

rkitover:
@Mystro256 can you please try the cairo_wayland branch and see if it fixes the display issue for Cairo on Wayland?

rkitover:
@Mystro256 BTW we are on `#vba-m` on Freenode if you use IRC.

rkitover:
@Mystro256 also wanted to ask, does the Simple renderer work on Wayland? How does the performance of the Simple and Cairo renderers compare?

I'd look at some of this stuff myself, but I don't have a non-VM linux right now, and my VM software only supports xorg.

Mystro256:
Simple and Cairo seems to be the same, but I think this likely because I believe only cairo drawing for WX is supported on wayland, so it's likely all routed via the same code paths in the backend. Disabling the cairo cmake flag seems to segfault vba-m on launch in wayland, which seems consistent with my guess (similar to how the opengl wx bug mentioned above causes sefaults due to hardcored Xorg calls).

I'll give the branch a shot and let you know. As for IRC, I don't use it unfortunately.

Mystro256:
Looks like the fix does nothing. I take back what I said about cairo drawing being required for build time, but simple and cairo still gives similar performances, likely due to most wayland implementations using cairo to draw regardless. Opengl still crashes vba-m in wayland, which is due to the bug you described above.

I've been meaning to bug a WX dev about it, but I'm betting it's not due to a WX bug, as I noticed wx tutorials using refresh() to force re-queuing redraws. I'm assuming this doesn't affect Xorg, as I'm told XOrg takes a more "active" approach to redrawing windows, opposed to wayland which is made to be more efficient. I have no idea how Windows or Quartz works.

rkitover:
@Mystro256 that still leaves the question of why the Simple renderer works and Cairo does not.

I'm going to turn Wayland back on in my Fedora VM and try it out.

Mystro256:
Sorry I may have been confusing on how I phrased it; both simple and Cairo have this issue. I thought it was just Cairo but it seems like it's both. While OpenGL does not work due to the wx bug described above.

rkitover:
I tried to make a Fedora VM with Wayland, but Wayland seems to require kernel modesetting which is only supported by specific video drivers, so no luck. I'll try on my pc laptop.

rkitover:
@Mystro256 I got wayland running on my arch VM with weston + fbdev-backend.so + uvesafb, so I should be able to test this now.

rkitover:
@Mystro256 I cannot reproduce the rendering bug in Weston, both the simple and cairo renderers work fine.

Any chance you could try rawhide and see if it happens there too?

If not, perhaps we should just revisit this when f26 comes out.

rkitover:
@Mystro256 on second thought, it was explained to me in `#wayland` that this is likely a difference between mutter and weston, so I will try f25 on real hardware after all.

Mystro256:
Understood; do you think this is a bug in mutter itself since it works in Weston? I've already found a few bugs already that I've reported to gnome in regards to Wayland.

rkitover:
It's either mutter or some sort of interaction between GTK and mutter, I don't know enough about this to say, also I need to install something to take a look myself.

rkitover:
@Mystro256 I managed to install rawhide in vmware fusion (which has a real drm driver) and run the gnome wayland, and reproduce the issue you are talking about.

rkitover:
@Mystro256 I think I have a fix, please try the cairo_wayland branch again.

You had the right idea about using Refresh(), just had to find the right place to put it.

Basically what I did was, instead of having:

core -> systemDrawScreen() -> panel::DrawArea(frame_buffer) -> panel::DrawArea(device_context)

I changed it to:

core -> systemDrawScreen() -> panel::DrawArea(frame_buffer) -> panel::Refresh() -> panel::PaintEv() -> panel::DrawArea(device_context)

and this puts things into the right state so that the frame actually appears in the window.

There is some other stuff I need to do for Wayland, I need to add some code to check for Wayland at runtime and disable opengl, otherwise you have the situation where the user has opengl in their config and the app segfaults on startup. And unify the OSD code (there are two sets of code right now to draw the OSD and sometimes you even see them both draw text on the frame....)

Mystro256:
Indeed, this seems to have a much cleaner implementation rather than my hacky workaround. Seems to work with Xorg and Wayland just fine. Providing it works with Msys and Mac, commit away.

Thanks for your help, I made a tracker bug for the opengl issue: #76

rkitover:
@Mystro256 please review and try a98f4a1 from the cairo_wayland branch. It also addresses #76.

If all looks good to you will merge to master.

How close are we to a usable fedora package? What are the major blockers aside from this?

rkitover:
@Mystro256 another thing I wanted to mention, when I was trying Weston on my arch VM, I didn't know that the arch wxWidgets is compiled with GTK2 not GTK3, so the app was actually running under XWayland and not Wayland and that's why I didn't see the bug.

Mystro256:
Ah yes, that would make a lot of sense. You would have to use "_wxgtk-gtk3_" from the AUR to test Wayland on Arch.

Looks good, feel free to pull the change in. I think the only other issue is #51.

rkitover:
Merged to master, closing this.

PR #73, opened 2017-02-12
Re-add missing installed icons (for linux)

Mystro256:
Looks like the cleanup for vbam-gtk removed the icon files for Linux.

I've just re-added everything and stuck it in a new folder.


rkitover:
@Mystro256  👍 

Are you sure the basename of the icon files should be "vbam" and not "wxvbam"? When you build this, does the generated menu entry in e.g. Gnome find the icons correctly?

Mystro256:
If you take a look at wxvbam.desktop, the icon is listed as "vbam". If you
would like to change it, that's fine but we might as well change everything
to visualboyadvance-m if that's the case. Please let me know what you think.

Also yes, I did a test build and everything seems good.

rkitover:
@Mystro256 

We are planning a rebranding and name change for the project soon anyway, so let's leave it alone for now if it works.

Merging, thanks!

Mystro256:
Sounds good!

Issue #63, opened 2017-01-03
find cheats not working on gameboy


Cibomatto2002:
There is a bug when trying to make gameboy cheat codes i know how to make them but it's not working on visualboyadvance-m. Will someone please try to make some Infinite Missiles on Metroid 2 so you can see what i'm talking about. On Bizhawk it's easy to make Infinite Missiles. but there RAM search is more advance so maybe that's why it's easier. 

![capture1-4-2017-3 06 39 am](https://cloud.githubusercontent.com/assets/21169038/21637086/fcb05ae0-d22a-11e6-8679-b4698426ae43.jpg)

 

visualboyadvance-m-win32-Beta-3

DoctorWho11:
If you are going to close an issue, please state why you closed it, if not I'm going to reopen.

Cibomatto2002:
I did not think anyone was going to help me with this problem so I was thinking I was wasting you all's time. You can make Cheats on Gameboy color. How do I do a unknown value search? Not all codes can be found by typing in a number.

DoctorWho11:
We will work on it, just at least with myself have day jobs and I also triage issues by importance as well. But your issues will be looked into. Just be patient.

Cibomatto2002:
Thank you. Sorry I was inpatient.

Issue #62, opened 2017-01-02
Cheat Search bug


Cibomatto2002:
How do I do a unknown search? I know how to do it in other emulators but not this one.

Cibomatto2002:
How do I do a unknown search? I know how to do it in other emulators but not this one.

DoctorWho11:
Gonna reopen this because we haven't had time to get to it and even though that's the case it's on the tasks of things to work on and investigate

Cibomatto2002:
Thank you for looking into it.

PR #74, opened 2017-02-12
Fix GCC warnings, incorrect c flag, wrong return type

Mystro256:
Some misc fixes:

- Fix some GCC warnings and added some error cases/checking
- fpermissive is not a C flag, only a c++ flag. I've removed it to silence a warning.
- cheatsImportGSACodeFile incorrectly returns false in non-error case, should be true for success (see src/wx/cmdevents.cpp). Caused an incorrect error code to print.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -315,7 +315,7 @@ IF(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID STREQUAL Clang)
     ENDIF()
 
     # common flags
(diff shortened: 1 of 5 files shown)


rkitover:
@Mystro256 I rebased and added a small commit (-fpermissive is required on Ubuntu 14 or Debian 8, I forget which, I put it in C++ flags only.)

Please check that everything is OK and I will merge.

rkitover:
@Mystro256 also, can you please check if any of your changes fix either #62 or #63.

Mystro256:
Looks good.

As well, I only changed the code when loading cheats and patches from file, while those issues seem to be runtime and GUI enhancement requests.

PR #77, opened 2017-02-18
Tweaks

Mystro256:
First commit looks to be a typo, as the loop iterates one extra time. I would think the check and break should prevent unintentionally dereferencing recent_accel out of bounds (i.e. if i=10), but it could call recent->FindItem once redundantly.

Second commit is using the prefix ++ operator for non-primitive variables with unused return values, as this could yield better performance. Depending on how smart the compiler is, this may not do anything, as when the return value is unused, a smart compiler would optimise out the opt codes that store and return of the variables previous value in the case of a post ++ op.


rkitover:
@Mystro256 sorry this didn't register in my brain when I looked at it last, merging it now.

Issue #84, opened 2017-02-22
Freezes when opening the second game with SDL audio


sergiobenrocha2:
### Actual behavior

Freezing when opening the second ROM. Crashes when opening the second GBA ROM.

### Steps to reproduce the behavior

Choose SDL in audio settings.

Open a GB/GBC game, then try open another, it'll freeze. Open a GBA game, then open another GBA game, it'll crash and you'll get this:

```
Program received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffb3fff700 (LWP 17198)]
0x0000000001558330 in ?? ()
(gdb) bt full
#0  0x0000000001558330 in ?? ()
No symbol table info available.
#1  0x00007ffff772a5fb in SDL_RunAudio (devicep=devicep@entry=0x1556a10) at /build/buildd/libsdl2-2.0.2+dfsg1/src/audio/SDL_audio.c:483
        silence = 0
        device = 0x1556a10
        stream = 0x17cd710 ""
        stream_len = 2048
        udata = 0x17d02b0
        fill = 0x4f2670 <SoundSDL::soundCallback(void*, unsigned char*, int)>
        delay = 11
#2  0x00007ffff778673d in SDL_RunThread (data=0x1558340) at /build/buildd/libsdl2-2.0.2+dfsg1/src/thread/SDL_thread.c:282
        args = 0x1558340
        userfunc = 0x7ffff772a4a0 <SDL_RunAudio>
        userdata = 0x1556a10
        thread = 0x1558000
        statusloc = 0x1558010
#3  0x00007ffff77d1279 in RunThread (data=<optimized out>) at /build/buildd/libsdl2-2.0.2+dfsg1/src/thread/pthread/SDL_systhread.c:72
No locals.
#4  0x00007ffff74fa184 in start_thread (arg=0x7fffb3fff700) at pthread_create.c:312
        __res = <optimized out>
        pd = 0x7fffb3fff700
        now = <optimized out>
        unwind_buf = {cancel_jmp_buf = {{jmp_buf = {140736213284608, -1725036879148407560, 1, 25772416, 140736213285312, 140736213284608, 1724904937141904632, 
                1725052957179979000}, mask_was_saved = 0}}, priv = {pad = {0x0, 0x0, 0x0, 0x0}, data = {prev = 0x0, cleanup = 0x0, canceltype = 0}}}
        not_first_call = <optimized out>
        pagesize_m1 = <optimized out>
[...]

PR #92, opened 2017-03-05
Fixes for SDL throttle

Mystro256:
Looks like the throttle issues for SDL was just caused by the audio device not being closed correctly. This was causing incorrect game speeds and a segfault on quit.
I also changed it so the throttle is applied on init, so the device isn't unnecessarily initialized twice (once for load, another to apply the throttle).

### DIFF
--- a/src/common/SoundSDL.cpp
+++ b/src/common/SoundSDL.cpp
@@ -28,7 +28,7 @@ const float SoundSDL::_delay = 0.032f;
 SoundSDL::SoundSDL():
 	_rbuf(0),
 	_dev(-1),
-	current_rate(0),
+	current_rate(throttle),
 	_initialized(false)
 {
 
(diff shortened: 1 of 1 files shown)


Mystro256:
Note this also fixes #84 on my system

rkitover:
Tested on windows, works perfectly, thanks!

BTW, does SDL sound work for you on Linux? Last time I tried it was all garbled on Linux and Mac, that's why we made OpenAL the default.

rkitover:
@Mystro256 some more good news, throttle now works in all sound drivers as of 7b85964e and OpenAL is now linked at compile time.

Mystro256:
Looks great, thanks! Yeah it looks like I forgot to account for when throttle is 0. As for SDL on Linux, it works on Linux for me after this fix, while before it had problems.

PR #116, opened 2017-05-11
Add BIN_INSTALL_DIR and MAN_INSTALL_DIR

orbea:
With Slackware man pages are installed to `/usr/man` instead of `/usr/share/man` and game binaries are generally installed to `/usr/games` instead of `/usr/bin`. This patch will allow setting these directories with `-DBIN_INSTALL_DIR=/usr/games` and `-DMAN_INSTALL_DIR=/usr/man`.

They now will be installed to the correct directories.
```
usr/games/
usr/games/visualboyadvance-m

usr/man/
usr/man/man6/
usr/man/man6/visualboyadvance-m.6.gz
```
Likewise if they defines are not set, then they install in the default directories that were previously hard-coded.
```
usr/bin/
usr/bin/visualboyadvance-m

usr/share/man/
usr/share/man/man6/
usr/share/man/man6/visualboyadvance-m.6
```


ZachBacon:
Wow, then slack does things nonstandard. I don't think I want to pull this in if it's for one distro. If others do that, then I may merge it in, but for now, this is a no go.

ZachBacon:
Ok, I'm going to merge it in under one condition. Add the appropriate documentation in the readme file. Once you do that, I'll merge it in.

ZachBacon:
Merging it now, thanks for the contribution.

orbea:
Other distros might use this too if they use non-standard locations. Guix and nix come to mind. Additionally the current vbam slackbuild at slackbuilds.org which I do not maintain is accomplishing this with these following sed hacks. I find configuring this with cmake is a far cleaner solution. :)
```
# install binary to /usr/games, since it's got a section 6 man page.
sed -i '/DESTINATION bin/s,bin,games,' CMakeLists.txt

# man pages go in Slack standard location.
sed -i 's,share/man,man,g' CMakeLists.txt
```
https://slackbuilds.org/repository/14.2/games/vbam/

Though I have added it to the README, I took the liberty of adding `DATA_INSTALL_DIRECTORY` too since it was missing, do you have any more suggestions? :)

ZachBacon:
My main realization for adding this in was merely the fact it wasn't hard coded paths, and as much as nonstandard paths are a pain, at least it allows for it, for which I thank you. And as for suggestions, not sure. Nothing on the top of my head

rkitover:
@orbea you should really be using this:

https://cmake.org/cmake/help/v2.8.12/cmake.html#module:GNUInstallDirs

orbea:
@rkitover I only tried to follow the style already in the CMakeLists.txt, anything more is probably beyond the scope of this small patch and should be addressed in another pull request. :)

PR #118, opened 2017-05-12
wx/wxvbam: fix GetAbsolutePath

kwkam:
Makes battery/state directory setting work when set as relative path


ZachBacon:
Thanks, looking good.

Issue #134, opened 2017-08-07
openSUSE Tumbleweed compile errors on latest branch


avindra:
Besides a pile of warnings, getting errors such as:

```c++
error: no matching function for call to 'wxXmlResource::DoGetXRCID(wxString&)'
/home/abuild/rpmbuild/BUILD/visualboyadvance-m-master/src/wx/wxhead.h:113:47: 
 #define XRCCTRL(win, id, type) XRCCTRL_I(win, XRCID(id), type)
                                               ^
```

```c++
error: invalid cast from type 'wxString' to type 'const char*'
/home/abuild/rpmbuild/BUILD/visualboyadvance-m-master/src/wx/wxvbam.cpp:571:45: 
     home = strdup((const char*)wxApp::argv[0]);
                                             ^
```

```c++
error: no matching function for call to 'ToString(wxString&)'
/home/abuild/rpmbuild/BUILD/visualboyadvance-m-master/src/wx/guiinit.cpp:3217:67: 
                 wxWindow* w = SafeXRCCTRL<wxWindow>(d, ToString(pn).c_str());
                                                                   ^
```

You can see the full log from the build here: 

https://pastebin.com/u4DLqTbA

Tumbleweed is currently using gcc 7.

They also have [a note on this specific issue](https://en.opensuse.org/openSUSE:Packaging_wxWidgets):

> The openSUSE wxWidgets is built in Unicode and STL mode. This is the recommended upstream mode. If your wx-using code fails to compile, that is usually because it sucks at strings. The use of wxString::c_str and wxChar[] is an indicator of this. Talk to the upstream authors of the end-user program to rectify it.

Here is the spec file being used to compile `vbam`

https://build.opensuse.org/package/view_file/home:avindra/vbam/vbam.spec?expand=1

Here, you can see some things I tried to do to get the package to build:

https://build.opensuse.org/package/revisions/home:avindra/vbam

----

If all goes well, my goal is to have this submitted to the `Emulators` project in openSUSE ([link](https://build.opensuse.org/project/show/Emulators)), which will enable openSUSE users to easily install a pre-compiled version of vbam for openSUSE. The [current snapshot of vbam](https://build.opensuse.org/package/show/Emulators/vbam) there is 4 years old and is failing to build.

avindra:
Quick update, seems SUSE has 

https://software.opensuse.org/package/wxWidgets-3_0-nostl

to tackle this issue. Will retry with this version of wxWidgets.

[...]

PR #135, opened 2017-08-07
Fix errors found by suses linter

avindra:
Fix the second batch of issues found by suse's linter (see https://github.com/visualboyadvance-m/visualboyadvance-m/issues/134)

*Full disclosure:* I am a total C++ noob, so please review these changes carefully before accepting.

### DIFF
--- a/src/gba/GBALink.cpp
+++ b/src/gba/GBALink.cpp
@@ -1918,7 +1918,8 @@ static void StartRFUSocket(uint16_t value)
                                     if (!ok) {
                                         rfu_curclient = rfu_numclients;
                                         rfu_data.rfu_clientidx[(rfu_masterdata[0] - 0x61f1) >> 3] = rfu_numclients;
-                                        rfu_clientlist[rfu_numclients] = rfu_masterdata[0] | (rfu_numclients++ << 16);
+                                        rfu_clientlist[rfu_numclients] = rfu_masterdata[0] | (rfu_numclients << 16);
+                                        rfu_numclients++;
                                         gbaid = (rfu_masterdata[0] - 0x61f1) >> 3;
                                         rfu_data.rfu_signal[gbaid] = 0xffffffff >> ((3 - (rfu_numclients - 1)) << 3);
                                     }
--- a/src/wx/openal.cpp
+++ b/src/wx/openal.cpp
@@ -356,10 +356,10 @@ bool OpenAL::GetDevices(wxArrayString& names, wxArrayString& ids)
         devs += strlen(devs) + 1;
     }
 
-#else
+#endif
+
     // should work anyway, but must always use default driver
     return true;
-#endif
 }
 
 #endif
--- a/src/wx/panel.cpp
+++ b/src/wx/panel.cpp
@@ -1658,6 +1658,8 @@ public:
             done->Post();
             continue;
         }
+
+        return 0;
     }
 };
 
--- a/src/wx/widgets/wxmisc.cpp
+++ b/src/wx/widgets/wxmisc.cpp
@@ -393,6 +393,7 @@ bool wxPositiveDoubleValidator::TransferToWindow()
         str_val = wxString::Format(wxT("%.1f"), *double_val);
         return wxGenericValidator::TransferToWindow();
     }
+    return true;
 }
 
 bool wxPositiveDoubleValidator::TransferFromWindow()


src/gba/GBALink.cpp
```diff
@@ -1918,7 +1918,8 @@ static void StartRFUSocket(uint16_t value)
                                     if (!ok) {
                                         rfu_curclient = rfu_numclients;
                                         rfu_data.rfu_clientidx[(rfu_masterdata[0] - 0x61f1) >> 3] = rfu_numclients;
-                                        rfu_clientlist[rfu_numclients] = rfu_masterdata[0] | (rfu_numclients++ << 16);
+                                        rfu_clientlist[rfu_numclients] = rfu_masterdata[0] | (rfu_numclients << 16);
+                                        rfu_numclients++;
```
avindra: Assuming the decrement happens after, so moving it here to be explicit.

src/wx/openal.cpp
```diff
@@ -356,10 +356,10 @@ bool OpenAL::GetDevices(wxArrayString& names, wxArrayString& ids)
         devs += strlen(devs) + 1;
     }
 
-#else
+#endif
+
     // should work anyway, but must always use default driver
     return true;
```
avindra: This pattern ensures the compiled code will always have `return true`. Technically both are fine, but the new way won't cause the linter to choke since the logic isnt totally hidden inside the `#` conditionals.

src/wx/panel.cpp
```diff
@@ -1476,7 +1476,7 @@ class FilterThread : public wxThread {
     // if NULL, end thread
     uint8_t* src;
 
-    ExitCode Entry()
```
avindra: What is `ExitCode`?

src/wx/panel.cpp
```diff
@@ -1476,7 +1476,7 @@ class FilterThread : public wxThread {
...
+    int Entry()
```
avindra: I'm assuming changing this to `int` will fix the error

src/wx/widgets/wxmisc.cpp
```diff
@@ -393,6 +393,7 @@ bool wxPositiveDoubleValidator::TransferToWindow()
         str_val = wxString::Format(wxT("%.1f"), *double_val);
         return wxGenericValidator::TransferToWindow();
     }
+    return true;
```
avindra: Don't know what a good default to return here is. I just put `true` for now

src/wx/panel.cpp
ZachBacon: Causes build error.

avindra: Any ideas what the right fix for this one would be?

I assumed the linter was complaining because the type of `ExitCode` doesnt match the integers that are being returned in the function body.

avindra: Seems that it wants `void` . Should I change it to that and remove change all the returns to `return;`?

avindra:
All green ! @ZachBacon can you have another look? Do you think keeping the return 0 would be appropriate? 

I read a bit about void pointers and it seems that they will happily accept any data type.

avindra:
`void*` per the error from ci:

```
C:/msys64/drone/src/github.com/visualboyadvance-m/visualboyadvance-m/src/wx/panel.cpp:1479:9: error: conflicting return type specified for 'virtual int FilterThread::Entry()'
     int Entry()
         ^
In file included from C:/msys64/mingw64/include/wx-3.0/wx/event.h:26:0,
                 from C:/msys64/mingw64/include/wx-3.0/wx/window.h:18,
                 from C:/msys64/mingw64/include/wx-3.0/wx/dcbuffer.h:16,
                 from C:/msys64/drone/src/github.com/visualboyadvance-m/visualboyadvance-m/src/wx/panel.cpp:2:
C:/msys64/mingw64/include/wx-3.0/wx/thread.h:615:19: error:   overriding 'virtual void* wxThread::Entry()'
     virtual void *Entry() = 0;
                   ^
```

src/wx/panel.cpp
rkitover: It's like the return value of a command in the shell, since a thread is essentially a process, 0 is success and anything else is failure.

http://docs.wxwidgets.org/trunk/classwx_thread.html#af50fa7d3ac55002e367e79989b9cbc5d

This is the signature of the pure virtual method, if you change it won't be called. The return type has to match the API.

Also the difference between `void` and `void*` in C and C++ is that the former means no return value, while the latter means 'any kind of pointer'.

This needs to be fixed differently, check the logic of the function and make sure in every logical branch there is a return statement, `0` for success and e.g. `1` for failure.

rkitover:
@avindra see comment above about `FilterThread::Entry`, once that is fixed 👍 from me.

avindra:
@rkitover Thanks! I will restore the `0`s to the `return` statements and see if it passes the post-build checks on SUSE's OBS

Should be able to get around to it after work today, so roughly 7pm EST.

src/wx/panel.cpp
```diff
@@ -1658,6 +1658,8 @@ class FilterThread : public wxThread {
             done->Post();
             continue;
         }
+
+	return 0;
```
avindra: It needed a default return value here, in case the `while` loop never ends.

Is `return 0` a valid default? cc @rkitover @ZachBacon

avindra:
One additional change to call out, which is a default return value for `Entry()`. I set it to `return 0`. Is this valid?

FWIW, it compiled for SUSE and I was able to start and stop emulation cleanly in the built version.

```patch
diff --git a/src/wx/panel.cpp b/src/wx/panel.cpp
index fb4702d..bb9d5a1 100644
--- a/src/wx/panel.cpp
+++ b/src/wx/panel.cpp
@@ -1476,7 +1476,7 @@ public:
     // if NULL, end thread
     uint8_t* src;
 
-    int Entry()
+    void* Entry()
     {
         // This is the band this thread will process
         // threadno == -1 means just do a dummy round on the border line
@@ -1658,6 +1658,8 @@ public:
             done->Post();
             continue;
         }
+
+        return 0;
     }
 };
```

avindra:
I can rebase to squash the commits into one. Not sure what the process you guys use is, lmk

ZachBacon:
We generally prefer commits to be rebased and squashed as per our readme.md file.

avindra:
Totally missed that. Rebased and squashed.

rkitover:
Merged manually in dd91abf7 .

PR #136, opened 2017-08-08
Lint: fix a few classes of warnings

avindra:
While linting for errors, SUSE also reported many warnings, among which were declarations of unused  variables.

There is one specific instance where the change may be breaking and needs more attention (see PR review).


src/wx/viewsupt.cpp
```diff
@@ -250,7 +249,7 @@ void DisList::SetSel()
             start += i;
             end += i + 1;
 #endif
-            tc.SetSelection(start, start + strings[i].size() + 1);
```
avindra: Please note this change

src/wx/viewsupt.cpp
```diff
@@ -250,7 +249,7 @@ void DisList::SetSel()
...
+            tc.SetSelection(start, end);
```
avindra: I think the intent here was to have this argument be `end`, so I changed it to that.

Does that seem right?

ZachBacon:
If you ever have time, stop by the IRC channel on freenode #vba-m

Issue #179, opened 2018-01-18
Build failure with ffmpeg git master


orbea:
### Expected behavior

visulaboyadvance-m should build with ffmpeg.

### Actual behavior

visualboyadvance-m does not build with the latest ffmpeg mater due to API changes.
```
make[2]: *** [CMakeFiles/vbamcore.dir/build.make:231: CMakeFiles/vbamcore.dir/src/common/ffmpeg.cpp.o] Error 1
```
Full build log - https://pastebin.com/4z3bvTzE

### Steps to reproduce the behavior

1. Build ffmpeg from a recent snapshot or their git master
2. Try building visualboyadvance-m with the new ffmpeg.

### Build details

`linux`
`gcc-7.2.0`
`ffmpeg-2018.01.08_3f234a0`

rkitover:
Yeah actually the ffmpeg stuff has been horribly broken anyway, all it could do was record sound clips, the whole thing needs to be fixed.

orbea:
I can't say I used it at all to begin with so I will just use `-DENABLE_FFMPEG=OFF` for now, but I'll leave this issue open since it seems worth fixing.

darktjm:
When I wrote it it worked fine for both video and audio.  I don't want to maintain code that actively violates my coding and UI standards, though (among other things).  When I was still maintaining it, I paid close attention to the change logs and tried to also make it backwards compatible as much as possible, but all patches to the ffmpeg code dropped that as well.  The only reason this even popped up on my radar is that I've stopped caring about my own, working version and just went ahead and installed gentoo's package (but never use; I've pretty much stopped playing emulated games), and noticed that ffmpeg is now masked.

rkitover:
@darktjm if you would be so kind as to share your working version with us, we will not ask you to maintain it.

orbea:
It builds now and should work so I am closing this.

PR #180, opened 2018-01-18
cmake: Fix -DENABLE_FFMPEG=OFF

orbea:
Fixes the cmake configure with `-DENABLE_FFMPEG=OFF` which was failing with this.
```
CMake Error at src/wx/CMakeLists.txt:518 (set_target_properties):
  set_target_properties called with incorrect number of arguments.


-- Configuring incomplete, errors occurred!
```
Related issue. https://github.com/visualboyadvance-m/visualboyadvance-m/issues/179


rkitover:
Thanks! Didn't notice that was broken.

PR #202, opened 2018-03-11
Fixed filter draw bounds

Eleuin:
When render_method is set to simple and filters are enabled, BasicDrawingPanel::DrawArea() makes bad assumptions about draw bounds. Make sure that stuff is actually drawn starting at the edge of the draw area


Eleuin:
there really still needs to be more work done on fixing vertical spacing that shouldn't exist, and some of the code could still be cleaned up

Issue #238, opened 2018-06-03
Cartoon Network Speedway does not display properly.


LofiSR:
### Expected behavior:
Game displays graphics normally.

### Actual behavior:
After turning, random horizontal lines garble the screen where 3D rendering is supposed to occur, making the game unplayable.

### Steps to reproduce the behavior
Simply begin a race and turn. The screen glitches.

### ROM details

Which games are affected by the issue (please state if GBA or GB game):
Cartoon Network Speedway (GBA)

### Option details

Video driver: OpenGL

Sound driver: OpenAL

### Build details

Operating System: Windows 10 64-bit Build 17134.rs4_release.180410-1804

VBA-M version: 2.0.1

Screenshot: https://vgy.me/nDRU1W.png

retro-wertz:
confirmed on linux version too.

ZachBacon:
What happens if the simple driver is used for graphics? Just out of curiosity.

retro-wertz:
im on linux. using simple driver still has this issue

ZachBacon:
Thanks for that, then problem is deeper than just simple graphics on the front end of things

retro-wertz:
im having problems doing a bisect since previous commits past march 1 does not compile on my end for some reasons...

ZachBacon:
Got an error log for me to look at? I know we've been doing successful compiling with the current setup for months even before March.

retro-wertz:
something like this starting from oct 2017 

https://hastebin.com/qojoqunuyi.rb

retro-wertz:
Pushed PR for this...

ZachBacon:
![1260 - cartoon network - speedway u venom](https://user-images.githubusercontent.com/8948676/41011266-000507f0-690a-11e8-97aa-aa33aee7140f.png)
can confirm the fix

ZachBacon:
I'll post a build shortly.

PR #243, opened 2018-06-05
Fix some opcodes in arm

retro-wertz:
Basically this just unrolls MULL and MLAL instructions to remove use of remaining typedefs and should fix issue where an incorrect variable was used.

Fixes issue: https://github.com/visualboyadvance-m/visualboyadvance-m/issues/238


ZachBacon:
lgtm

PR #328, opened 2018-11-28
Fix remote gdb support

### LINKED ISSUE #181: Debugging .gba with arm-none-eabi-gdb: "Cannot execute this command without a live selected thread."
I have tried making debugging in VBA-M work from VS Code with the native debugging extension to debug a GBA project. It did not work, so I tried gdb from the command line, but still no dice. It seems the connection works, but gbd and VBA-M do not work together. I haven't used gbd much, so if I've missed something, please tell me...

### Expected behavior
GDB steps through the executable

### Actual behavior
Error message: "Cannot execute this command without a live selected thread."

- Start VBA-M. Set GBD->"Break on load"
- Load .gba -> waits for connection
- start gdb from devkitPro (arm-none-eabi-gdb) with .elf source of .gba
- "target remote :55555"
- gbd connects, VBA-M connection dialog closes
- try to "step" or "continue" from gbd 

Now gbd shows: "Cannot execute this command without a live selected thread." and nothing happens in VBA-M. The same error message comes up in VS Code, so I hope I'm not doing it so wrong after all... :/

Which games are affected by the issue (please state if GBA or GB game):
GBA rom which otherwise runs fine in VBA-M.

### Build details

Ubuntu 16.04.03 x64

VBA-M from Ubuntu repo, 2.0.1-r201801190125-3e052c9-22~ubuntu16.04.1 amd64

knightsc:
Fixes #181 

See the steps to reproduce on the original issue. This should get remote debugging working again with gdb 8.0 which comes in the latest devkitpro.

The remote query response packets needed to be formatted slightly different and the cpsr register number is changed.

### DIFF
--- a/src/gba/remote.cpp
+++ b/src/gba/remote.cpp
@@ -3700,17 +3700,24 @@ void remoteMemoryRead(char* p)
 void remoteQuery(char* p)
 {
     if (!strncmp(p, "fThreadInfo", 11)) {
-        remotePutPacket("m-1");
+        remotePutPacket("m1");
     } else if (!strncmp(p, "sThreadInfo", 11)) {
         remotePutPacket("l");
     } else if (!strncmp(p, "Supported", 9)) {
         remotePutPacket("PacketSize=1000");
+    } else if (!strncmp(p, "C", 1)) {
+        remotePutPacket("QC1");
+    } else if (!strncmp(p, "Attached", 8)) {
+        remotePutPacket("1");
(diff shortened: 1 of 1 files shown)


knightsc:
@ZachBacon any idea why the tea CI is failing? When I go to that page it just shows up as blank and retrying the CI run doesn't seem to do anything

rkitover:
I think the tea CI is dead.

PR #329, opened 2018-11-30
libretro updates

ghost:
name change, support for multiline cheats for gb/gbc, add turbo buttons


Tatsuya79:
It seems everyone would agree with it being the new VBA-M core.
The main thing missing atm is the additional frame of input lag that Retro-Wertz fixed in the libretro current version:
https://github.com/libretro/vbam-libretro/commit/3031dc53aa5799db40c5207de09160fc3960bf7f

PR #355, opened 2019-01-26
libretro: Add GB color palettes

retro-wertz:


rkitover:
Next time you're doing something like this, consider using a C++ `unordered_map` or something along those lines instead of this horrible C code.

PR #358, opened 2019-01-29
gba save type detection updates, missing save for MMM01 cart for GB and libretro updates

retro-wertz:


rkitover:
@retro-wertz 

I apologize if I offended you.

I was only suggesting that sometimes using C++ features can make code much cleaner and nicer.

If you don't want to, that is fine, your code is always welcome in any case.

PR #359, opened 2019-01-30
GBA: Remove some magic numbers for main pointers and save types size

retro-wertz:
- use labels instead of magic number for known sizes for pointers and save types

feel free to change any labels or if its better to use #define or const <variable type>


rkitover:
This is fantastic thanks! We talked about starting to do this before.

PR #364, opened 2019-02-05
bugfix: crash when loading elf

laqieer:
It reads data from uninitialized memory, converts it to pointer and accesses invalid address 0xCDCDCDCD
[screenshot1](https://imgur.com/a/5vhUOom)
[screenshot2](https://imgur.com/a/UPg77Yl)


ZachBacon:
Looking good, thanks.

Issue #94, opened 2017-03-06
Use XDG specification for config files


sergiobenrocha2:
Right now there's a ~/.vbam directory, which should be located at ~/.config/vbam

Not directly related, but starter directory for search roms should be the home folder (~/), not ~/.vbam

denisfa:
Any updates on this, @ZachBacon ? If it is still open (and just this issue not closed), would you review a PR? I am interested in working on this.

From what I see, there is only a single config file `~/.vbam/vbam.conf`. We can do backward compatibility (check if old dir exists; if it does, use it. Otherwise, use new location), while setting the new default location to `${XDG_CONFIG_HOME:-$HOME/.config}/vbam/vbam.conf` (I don't like the idea of changing the name of the configuration file, although I have seen on other software).

About this file, there is only one thing that we need to discuss: the `[recent]` section.

According to [this](https://wiki.debian.org/XDGBaseDirectorySpecification#state), this type of info leans towards user data, although it is not actually that important to a user. We could split this to a second file, to be put, let's say, on `${XDG_DATA_HOME:-$HOME/.local/share}/vbam/vbam.ini`.

Of course this is just a brainstorming; we can just keep a single file and use the XDG location.

rkitover:
@denisfa 

I last worked on the config code a few years for better mac and windows support. On those platforms the file is in the correct location.

I did not change the location on unix because of backwards compatibility concerns you mentioned which I did not want to deal with at the time, I was only trying to get the program working on mac.

If you are interested I'll give you some hints on where the relevant code is etc..

denisfa:
@rkitover Thanks!

Sure, I could confirm some stuff! I believe I have to work all files using the macro `DOT_DIR`.
So the following files: `./src/common/ConfigManager.cpp`, `./src/sdl/SDL.cpp` and the header `./src/common/ConfigManager.h`.

About the wxWidgets, I saw something like a `FileLayout::FileLayout_XDG`. Is this what you mean? I am not being able to test this now, because the file `/usr/include/wx-3.0/wx/unix/stdpaths.h` from my package distro does not contain this. It is missing some stuff, but I think I am probably looking at the wrong place.

The file for this should be `./src/wx/wxvbam.cpp`.

What are your thoughts about the splitting? Or should we keep a single file for the sake of simplicity?

rkitover:
@denisfa 

[...]

PR #380, opened 2019-02-28
[WIP] Add support for the XDG Base Dir Spec for Linux with legacy support.

denisfa:
We check if `$HOME/.vbam` exists; if it does, then we use
`$HOME/.vbam/vbam.conf`; otherwise, use
`${XDG_CONFIG_HOME:-$HOME/.config}/visualboyadvance-m/vbam.conf`.

Fix for #94.

@rkitover Here it is a first code for this. I have some comments to do about it, so we could check for improvements:

1. I see the config file is `vbam.ini` (first try) or `vbam.cfg` for Windows and MacOS, but the name for Linux is `vbam.conf`. I kept this, since this is what we discussed previously, but maybe we could use the name for Win and Mac with Linux for the XDG Dir. I am not sure how hard this is right now, but with wxWidgets >= 3.1 it should be a piece of cake. 

2. I believe we could also use the time until wxWidgets >= 3.1 became stable as a "adaptation time" for users on Linux. I mean like only using the XDG dir when this happens. This would allow for a good refactoring of this part of the code.

3. Testing. Where do I code for automated testing? I did manually, using `wxWidgets 3.0` on my main and `wxWidgets 2.8` on Debian container. It worked as I expected, but there is never enough testing. I have a template for this, but would like further instructions.

As always, check if the code quality is good enough to be merged. There is always room for improvement! I have some free time right now, so I could take your feedback and code it (or try to, at least).

### DIFF
--- a/src/common/ConfigManager.cpp
+++ b/src/common/ConfigManager.cpp
@@ -718,14 +718,7 @@ const char* FindConfigFile(const char *name)
 void LoadConfigFile()
 {
 #if !defined(_WIN32) && !defined(__APPLE__)
-	// Get home dir
-	char buf[1024];
-	struct stat s;
 	homeDir = getenv("HOME");
-	snprintf(buf, 1024, "%s/%s", homeDir, DOT_DIR);
-	// Make dot dir if not existent
-	if (stat(buf, &s) == -1 || !S_ISDIR(s.st_mode))
-		mkdir(buf, 0755);
 #else
 	homeDir = 0;
 #endif
@@ -746,14 +739,7 @@ void LoadConfigFile()
 void SaveConfigFile()
 {
 #if !defined(_WIN32) && !defined(__APPLE__)
-	// Get home dir
-	char buf[1024];
-	struct stat s;
 	homeDir = getenv("HOME");
-	snprintf(buf, 1024, "%s/%s", homeDir, DOT_DIR);
-	// Make dot dir if not existent
-	if (stat(buf, &s) == -1 || !S_ISDIR(s.st_mode))
-		mkdir(buf, 0755);
 #else
 	homeDir = 0;
 #endif
--- a/src/wx/wxvbam.cpp
+++ b/src/wx/wxvbam.cpp
@@ -32,16 +32,30 @@
 IMPLEMENT_APP(wxvbamApp)
 IMPLEMENT_DYNAMIC_CLASS(MainFrame, wxFrame)
 
+// Get XDG_CONFIG_HOME dir manually
+// only native support for XDG config when wxWidgets >= 3.1
+static wxString get_xdg_user_config_home()
+{
+    wxString path;
+    char *xdg_config_home = getenv("XDG_CONFIG_HOME");
+    // Default for XDG_CONFIG_HOME is '$HOME/.config'
+    if (!xdg_config_home || !*xdg_config_home)
+    {
+	wxString xdg_default(getenv("HOME"));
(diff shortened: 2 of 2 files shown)


rkitover:
@denisfa thank you, this looks great!

Some notes:

* probably best to use `__WXGTK__` instead of `__LINUX__` to check for linux (and freebsd, etc., don't forget they still exist! :smiling_imp: )

* the ConfigManager stuff, we just need to check that the SDL port still works on windows, I will do this and make adjustments if necessary. To build the SDL port pass `-DENABLE_SDL=ON` to cmake.

* instead of checking for `.ini` on linux and `.conf` elsewhere, let's just check for both, check for `.ini` first and rename to `.conf` if it doesn't exist already, would be nice to have more consistency between platforms anyway

* thank you for testing on `2.8`, I put in a lot of work to make `2.8` still work a couple years ago, good to know it still does

* probably a good idea to show a warning about the config file being in the old location and where to move it to, e.g. with `wxLogWarning()`

* we cannot do any automated testing at this time, there are too many memory corruption bugs and I can't even run `--help` in the CI without segfaults

I'm going to merge this now so I can do some testing as well, just send additional PRs if and when you want, thank you!

denisfa:
> @denisfa thank you, this looks great!
>
> Some notes:
> * probably best to use `__WXGTK__` instead of `__LINUX__` to check for linux (and freebsd, etc., don't forget they still exist! smiling_imp )

I see. I focused on Linux too much, but at least this is a quick fix.

> * the ConfigManager stuff, we just need to check that the SDL port still works on windows, I will do this and make adjustments if necessary. To build the SDL port pass `-DENABLE_SDL=ON` to cmake.

The code affected was protected by `#if !defined(_WIN32) && !defined(__APPLE__)`. This should not be a problem for Windows build/run. But I should have tested the SDL for Linux too. I already got a message `Error creating file /home/denisfa/.vbam/loz.sav`. This makes sense since I removed the `.vbam` dir creation. On this topic, should we move this type of file to a XDG dir too? This feels like data for me, and we could use `${XDG_DATA_HOME:-$HOME/local/share}/visualboyadvance-m`. What do you think?

> * instead of checking for `.ini` on linux and `.conf` elsewhere, let's just check for both, check for `.ini` first and rename to `.conf` if it doesn't exist already, would be nice to have more consistency between platforms anyway

So use `.conf` for all platforms? `.ini` and a fallback option `.cfg` are used for Windows and MacOS. `.conf` is for everything else.

> * thank you for testing on `2.8`, I put in a lot of work to make `2.8` still work a couple years ago, good to know it still does

Someone is probably going to complain if it indeed stops, so no worry about that hahaha.

> * probably a good idea to show a warning about the config file being in the old location and where to move it to, e.g. with `wxLogWarning()`

I agree. My original idea was to just put on release notes or something, but we can and should be nicer to users.

> * we cannot do any automated testing at this time, there are too many memory corruption bugs and I can't even run `--help` in the CI without segfaults

About this, when I build the original master branch and ran `./visualboyadvance-m --help` or `./visualboyadvance-m --print-cfg-path`, I got `Segmentation fault (core dumped)`. Is this what you mean? Or is this a possible example?

> I'm going to merge this now so I can do some testing as well, just send additional PRs if and when you want, thank you!

I am going to copy this to the related issue #94  section so we can continue our discussion there. I will also reset my branch (delete and create a new one from your master branch) so I can better manage my branches (I wanted a `denisfa:xdg-spec`, but committed and pushed to `denisfa:master`).

PR #383, opened 2019-03-06
[WIP] XDG Base Dir Spec followup.

denisfa:
@rkitover 

As we previously discussed, these were the next steps to support the spec. So keep This includes:

***What was done***

1. Fix `#ifdef` to support more platforms for the spec.

2. Changing dirs from `DOT_DIR` `.vbam` to a XDG Dir 
`${XDG_DATA_HOME:-~/.local/share}/visualboyadvance-m`. This is for the SDL port and it is mainly for Unix.

3. Migration. From `vbam.conf` to `vbam.ini` automatically. This includes a appropriate error message from `wxWidgets`.

***Testing***

1. As discussed #94.

2. Also more tested by me manually. I build the SDL port, run a game and try to save the different types of files and see if theu go into the XDG dir.

3. Manual testing. There are 4 cases: (1) `vbam.ini` exists and `vbam.conf` doesn't; (2) `vbam.ini` and `vbam.conf` exist; (3) neither file exist; (4) `vbam.conf` exists and `vbam.ini` doesn't. The results were:

(1) No warning message displayed. Configuration loaded as set in `vbam.ini`;
(2) Warning message displayed from `wxWidgets`; uses configuration from `vbam.ini`;
(3) Creates new `vbam.ini`;
(4) `vbam.conf` is renamed to `vbam.ini` and its configuration is loaded.

In case you want to test anything more, clone my repo and branch instead of merging to the main repo.
I had to refactor some coding from previous commits. As usual, keep your eye on code quality! There is always room for improvement!

### DIFF
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -64,6 +64,66 @@ bool FileExists(const char *filename)
 #endif
 }
 
+// Get user-specific config dir manually.
+// apple:   ~/Library/Application Support/
+// windows: %APPDATA%
+// unix:    ${XDG_CONFIG_HOME:-~/.config}
+std::string get_xdg_user_config_home()
+{
+    std::string path;
+#ifdef __APPLE__
+    std::string home(getenv("HOME"));
+    path = home + "/Library/Application Support/";
+#elseif _WIN32
+    std::string app_data(getenv("APPDATA"));
+    path = app_data + '\\';
+#else // Unix
+    char *xdg_var = getenv("XDG_CONFIG_HOME");
+    if (!xdg_var || !*xdg_var)
+    {
+	std::string xdg_default(getenv("HOME"));
+	xdg_default += "/.config";
+	path = xdg_default;
+    }
+    else
+    {
+	path = xdg_var;
+    }
+    path += '/';
+#endif
+    return path;
+}
+
+// Get user-specific data dir manually.
+// apple:   ~/Library/Application Support/
+// windows: %APPDATA%
+// unix:    ${XDG_DATA_HOME:-~/.local/share}
+std::string get_xdg_user_data_home()
+{
+    std::string path;
+#ifdef __APPLE__
+    std::string home(getenv("HOME"));
+    path = home + "/Library/Application Support/";
+#elseif _WIN32
+    std::string app_data(getenv("APPDATA"));
+    path = app_data + '\\';
+#else // Unix
+    char *xdg_var = getenv("XDG_DATA_HOME");
+    if (!xdg_var || !*xdg_var)
+    {
+	std::string xdg_default(getenv("HOME"));
+	xdg_default += "/.local/share";
+	path = xdg_default;
+    }
+    else
+    {
+	path = xdg_var;
+    }
+    path += '/';
+#endif
+    return path;
+}
+
 void utilReadScreenPixels(uint8_t *dest, int w, int h)
 {
         uint8_t *b = dest;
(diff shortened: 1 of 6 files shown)


denisfa:
The CI/CD check failed due to this:
```
Error: An exception occurred within a child process:
  DownloadError: Failed to download resource "giflib"
Download failed: https://downloads.sourceforge.net/project/giflib/giflib-5.1.4.tar.bz2
```

rkitover:
sourceforge download servers go down constantly, this plagues my build system as well

src/Util.cpp
```diff
@@ -64,6 +64,66 @@ bool FileExists(const char *filename)
...
+// windows: %APPDATA%
+// unix:    ${XDG_CONFIG_HOME:-~/.config}
+std::string get_xdg_user_config_home()
+{
+    std::string path;
+#ifdef __APPLE__
+    std::string home(getenv("HOME"));
+    path = home + "/Library/Application Support/";
+#elseif _WIN32
+    std::string app_data(getenv("APPDATA"));
```
rkitover: This should `getenv("LOCALAPPDATA")` to use `Local\` instead of `Roaming\`, we use `Local\` for the GUI as well.

src/Util.cpp
```diff
@@ -64,6 +64,66 @@ bool FileExists(const char *filename)
...
+// windows: %APPDATA%
+// unix:    ${XDG_DATA_HOME:-~/.local/share}
+std::string get_xdg_user_data_home()
+{
+    std::string path;
+#ifdef __APPLE__
+    std::string home(getenv("HOME"));
+    path = home + "/Library/Application Support/";
+#elseif _WIN32
+    std::string app_data(getenv("APPDATA"));
```
rkitover: as above

src/common/ConfigManager.cpp
```diff
@@ -718,7 +718,11 @@ const char* FindConfigFile(const char *name)
 void LoadConfigFile()
 {
 #if !defined(_WIN32) && !defined(__APPLE__)
```
rkitover: Since your functions are cross-platform, remove this `#if` and just use the same code for all platforms.

src/common/ConfigManager.cpp
```diff
@@ -739,7 +743,11 @@ void LoadConfigFile()
 void SaveConfigFile()
 {
 #if !defined(_WIN32) && !defined(__APPLE__)
```
rkitover: as above

src/common/ConfigManager.cpp
rkitover: Also win32 has `stat()`:

https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/stat-functions?view=vs-2017

You may need something like:

```cpp
#ifdef _WIN32
    #define stat _stat
#endif 
```

src/sdl/SDL.cpp
```diff
@@ -1637,12 +1638,23 @@ void handleRewinds()
     }
 }
 
+void SetHomeDataDir()
+{
+    sprintf(homeDataDir, "%s%s", get_xdg_user_data_home().c_str(), DOT_DIR);
+#if !defined(_WIN32) && !defined(__APPLE__)
```
rkitover: See above, this should be perfectly portable `stat()` code.

rkitover:
I left some comments on 724188a, looks great thanks!

denisfa:
@rkitover Coded the requested changes. My only doubt is about the `mkdir` function for windows. Since it only takes a single arg, I had to prepare a macro to handle it. Check if it is ok. It is protected by a `_WIN32`.
```
#define mkdir(X,Y) (_mkdir(X))
```

rkitover:
Alright, looks awesome, merging this now. Thank you!

PR #385, opened 2019-03-09
Add support to save/load geometry options for GUI window.

denisfa:
@rkitover I think I got this time.

I had to do a change regarding calling `update_opts()`. This function writes to disk everytime it is called. For the GUI geometry, this can lead to an unnecessary increase of disk usage and my idea is to only save this type of change when the GUI exits. I also kept the code from `fs-*`.  Not sure yet about the WX functions, but at least this feels a lot better than the other solution.

Keep sending tips if you feel it needs to change.


rkitover:
This looks great:

- don't worry about disk activity, config writes don't use `fsync()`, and the move event is only called when the user finishes moving the window, I don't mind the call to `update_opts()` also being in the destructor just in case, but saving these options frequently would be better because this app can crash a lot

- if position is `(0, 0)` it's best to not set it but let the operating system position the window

- it is best if we initialize `width` and `height` to `0` as well, and not use it if it isn't set, because the default size depends on various things like configured scale etc. and we don't want to hardcode it

rkitover:
I've had some more thoughts about this:

- we want to initialize the `posX`/`posY` to `(-1, -1)` to indicate that they are uninitialized, as distinguished from the top-left corner (you'd want to set the `min` to `-1` when defining the option.)

- for `width`/`height` `(0, 0)` is sufficient, since that is an impossible geometry, they can be uints

- we only want to restore fullscreen if a ROM is loaded, otherwise a full black screen is not going to be very helpful to the user

denisfa:
@rkitover More refactoring here. I coded your suggestions, just had an issue regarding the `x` and `y` starting positions. If they are `(-1, -1)` and we resize the window, these are not updated. Since I only know a function `wxWindow::SetSize` to start the wx window, if these are `-1`, then I cannot set the correct `width`,`height`. Hence my decision to also update these at `::OnSize`.

That's all for now, I think.

denisfa:
> don't worry about disk activity, config writes don't use fsync(), and the move event is only called when the user finishes moving the window, I don't mind the call to update_opts() also being in the destructor just in case, but saving these options frequently would be better because this app can crash a lot

This type of data is not important for the user. It is configuration but much more volatile. It is not a problem if it is not saved, just convenient if it is.

I was also checking for my `${XDG_CONFIG_HOME}/visualboyadvance-m/vbam.ini` while using a `printf` to test how/when the `x`,`y`, `height` and `width` were changing. When I was calling the `update_opts()` on `OnSize` and `OnMove`, the values were always being updated on `vbam.ini`. Not sure what was causing that (if specific from my end), but there was disk activity being registered. The way it is now only updates when the app exits normally, so it feels appropriate for this geometry information.

rkitover:
The way a modern operating system works is that filesystem activity is mostly kept in a RAM cache, there are only flushes to disk if you call fsync() or in idle time and periodically to sync the RAM caches with the physical device.

rkitover:
As a compromise on this point of disagreement, perhaps we can make a timer that calls `update_opts()` every 5 minutes or so. For an example of a timer see `src/wx/widgets/sdljoy.cpp`.

rkitover:
Otherwise your approach to everything else looks fine to me.

denisfa:
So this is something on my end... I will check what is happening. I have been playing with RAM disks lately because of other projects, but It is not obvious to me at all. With that, I agree with the `update_opts()` on both `OnSize` and `OnMove`.

rkitover:
Yeah you will see the change to the config file immediately, but the file is actually in RAM, this is all done by the filesystem caches of the operating system.

This is why `fsync()` exists, if you are writing really important data and want to make sure it is actually committed to disk.

denisfa:
Here it is. All the testing for this part was done manually. I believe it should be its final form.

rkitover:
Wonderful, merging this. Thank you so much for doing this work, it's definitely going to be a "quality of life" improvement for our users.

PR #387, opened 2019-03-13
Overall cleanup.

denisfa:
@rkitover This is a little bit of cleanup. On #94 we discussed about managing a single config file for all platforms (`vbam.ini`), so we needed to also include a migration path for `vbam.cfg` on MacOS and Windows.

We also had some stuff duplicated the `FILE_SEP`, like `${XDG_CONFIG_HOME}//visualboyadvance-m` from the xdg functions. Although it has no problem, since all platforms understand the duplicate `FILE_SEP`, it was a quick fix, so why not?

Finally, it is always a good practice to set a pointer to NULL after deallocating it.  This was missing from the variable `home` that was the cause of that segfault command line issue.

### DIFF
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -66,62 +66,58 @@ bool FileExists(const char *filename)
 
 // Get user-specific config dir manually.
 // apple:   ~/Library/Application Support/
-// windows: %APPDATA%
-// unix:    ${XDG_CONFIG_HOME:-~/.config}
+// windows: %APPDATA%\
+// unix:    ${XDG_CONFIG_HOME:-~/.config}/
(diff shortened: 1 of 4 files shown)


rkitover:
All of this looks fine to me. I'm not sure about a `.cfg` extension on windows, I don't think it ever used that, but that's ok.

Also this change for `home` does not fix the exit code, I have not yet figured out where that comes from, you can test it like this:

```bash
./visualboyadvance-m --help; echo $?
```

Also on windows `/` path separators are fine and understood by the POSIX layer, although using the backslashes is preferred.

Will merge this now.

PR #388, opened 2019-03-14
[WIP] Return non-error code for given command line parameters.

denisfa:
@rkitover I have an idea about how we could handle the return code `255`. But first, I will tell what I think it is happening, so you can check if we are on the same page.

When we run 
`./visualboyadvance-m --help; echo $?`
or 
`./visualboyadvance-m --print-cfg-path; echo $?`

we get the value `255`, which is a non-zero error code. For these 2 cases, we wish we got `0`, since it means the app ran successfully (only printing the asked info and then exiting).

According to my researches, the reason why we get `255` is because we returned `false` on the `OnInit()` function, causing it to exit the program immediately, but in the sense of failing to initialize. Sources [(1)](https://stackoverflow.com/questions/19255020/how-do-i-quit-a-wxwidget-application-when-no-window-was-opened) and [(2)](https://docs.wxwidgets.org/3.0/classwx_app_console.html#a99953775a2fd83fa2456e390779afe15).

Our function `wxvbamApp::OnCmdLineParsed` had several `return false` statements, all of them to parameters that print information (`--print-cfg-path`, for example). We wanted to print information and then exit the app successfully, because we only wanted to see the text from the options.

I searched the WxWidgets docs, and found some interesting stuff. The first one is to handle these parameters that we create; we could make a **console mode**, by overriding the `OnRun` method. The idea is to check if we are in this mode and return our own exit code; otherwise, return the parent function `wxApp::OnRun` [(3)](https://docs.wxwidgets.org/3.0/classwx_app_console.html#ac05a28c7cdb529f2cdfe77b3431c385c). This idea was coded here.

Now, for the `--help`, we have a more specific situation. This is handled by the function `OnCmdLineHelp`[(4)](https://docs.wxwidgets.org/3.0/classwx_app_console.html#ac05a28c7cdb529f2cdfe77b3431c385c). By the doc, this returns `false` for `OnInit()` and, again, in the sense of failing to initialize the app and exiting the application immediately; this is what generates the `255` for this param. If we don't want to consider this as a fail, we need to override `OnCmdLineHelp` and use the same logic that we did for the other params. The counter point to this, though, is that we get other info print alongside it. This is the output for the current code:

```
$ ./visualboyadvance-m -h; echo $?
VisualBoyAdvance-M

Usage: visualboyadvance-m [-h] [--verbose] [--save-xrc <str>] [--save-over <str>] [--print-cfg-path] [-f] [-s] [-o] [ROM file] [<config>=<value>...]
  -h, --help               	show this help message
  --verbose                	generate verbose log messages
  --save-xrc=<str>         	Save built-in XRC file and exit
  --save-over=<str>        	Save built-in vba-over.ini and exit
  --print-cfg-path         	Print configuration path and exit
  -f, --fullscreen         	Start in full-screen mode
  -s, --delete-shared-state	Delete shared link state first, if it exists
  -o, --list-options       	List all settable options and exit
04:37:07 PM: Debug: GetUserLocalDataDir(): /home/denisfa/.visualboyadvance-m
04:37:07 PM: Debug: GetUserDataDir(): /home/denisfa/.visualboyadvance-m
04:37:07 PM: Debug: GetLocalizedResourcesDir(wxGetApp().locale.GetCanonicalName()): /usr/local/share/visualboyadvance-m/en_US
04:37:07 PM: Debug: GetResourcesDir(): /usr/local/share/visualboyadvance-m
04:37:07 PM: Debug: GetDataDir(): /usr/local/share/visualboyadvance-m
04:37:07 PM: Debug: GetLocalDataDir(): /etc/visualboyadvance-m
04:37:07 PM: Debug: plugins_dir: /usr/local/lib/visualboyadvance-m
04:37:07 PM: Debug: XdgConfigDir: /home/denisfa/.config/visualboyadvance-m
0
```

Also, we have to consider wrong or misspelled parameters. Like this:
```
./visualboyadvance-m --print-cfg-paah; echo $?
Unknown long option 'print-cfg-paah'
VisualBoyAdvance-M

Usage: visualboyadvance-m [-h] [--verbose] [--save-xrc <str>] [--save-over <str>] [--print-cfg-path] [-f] [-s] [-o] [ROM file] [<config>=<value>...]
  -h, --help               	show this help message
  --verbose                	generate verbose log messages
  --save-xrc=<str>         	Save built-in XRC file and exit
  --save-over=<str>        	Save built-in vba-over.ini and exit
  --print-cfg-path         	Print configuration path and exit
  -f, --fullscreen         	Start in full-screen mode
  -s, --delete-shared-state	Delete shared link state first, if it exists
  -o, --list-options       	List all settable options and exit
255
```

We would need to define what we want from this. Maybe the message but a success return code? I am not sure... We can work it out by overriding `OnCmdLineError` [(5)](https://docs.wxwidgets.org/3.0/classwx_app_console.html#a1c8bf26b12a280fd03911caac70e3144).

It is really fun to search and learn how to use a library like WxWidgets, but on the other hand, since It is my first contact with it, I could be dropping the ball really hard here, considering you must have been working with it for a longer time and could have a better explanation for what it is happening. But hey, you miss 100% of the shots you don't take, right?

### DIFF
--- a/src/wx/wxvbam.cpp
+++ b/src/wx/wxvbam.cpp
@@ -421,6 +421,26 @@ bool wxvbamApp::OnInit()
     return true;
 }
 
+int wxvbamApp::OnRun()
+{
+    if (console_mode)
+    {
+	// we could check for our own error codes here...
+	return EXIT_SUCCESS;
+    }
+    else
+    {
+	return wxApp::OnRun();
+    }
+}
+
+bool wxvbamApp::OnCmdLineHelp(wxCmdLineParser& parser)
+{
+    wxApp::OnCmdLineHelp(parser);
+    console_mode = true;
+    return true;
+}
+
 void wxvbamApp::OnInitCmdLine(wxCmdLineParser& cl)
 {
     wxApp::OnInitCmdLine(cl);
@@ -505,7 +525,8 @@ bool wxvbamApp::OnCmdLineParsed(wxCmdLineParser& cl)
             s.mb_str());
         tack_full_path(lm);
         wxLogMessage(lm);
-        return false;
+        console_mode = true;
+        return true;
     }
 
     if (cl.Found(wxT("print-cfg-path"))) {
@@ -515,7 +536,8 @@ bool wxvbamApp::OnCmdLineParsed(wxCmdLineParser& cl)
         wxString lm(_("Configuration is read from, in order:"));
         tack_full_path(lm);
         wxLogMessage(lm);
-        return false;
+        console_mode = true;
+        return true;
     }
 
     if (cl.Found(wxT("save-over"), &s)) {
@@ -533,7 +555,8 @@ bool wxvbamApp::OnCmdLineParsed(wxCmdLineParser& cl)
         tack_full_path(lm, oi);
         lm.append(_("\n\tbuilt-in"));
         wxLogMessage(lm);
-        return false;
+        console_mode = true;
+        return true;
     }
 
     if (cl.Found(wxT("f"))) {
@@ -567,7 +590,8 @@ bool wxvbamApp::OnCmdLineParsed(wxCmdLineParser& cl)
         for (int i = 0; i < ncmds; i++)
             wxPrintf(wxT("%s (%s)\n"), cmdtab[i].cmd, cmdtab[i].name);
 
-        return false;
+        console_mode = true;
+        return true;
     }
 
 #if !defined(NO_LINK) && !defined(__WXMSW__)
--- a/src/wx/wxvbam.h
+++ b/src/wx/wxvbam.h
@@ -83,6 +83,8 @@ public:
     {
     }
     virtual bool OnInit();
+    virtual int OnRun();
+    virtual bool OnCmdLineHelp(wxCmdLineParser&);
     virtual bool UsingWayland() { return using_wayland; }
     virtual void OnInitCmdLine(wxCmdLineParser&);
     virtual bool OnCmdLineParsed(wxCmdLineParser&);
@@ -140,6 +142,7 @@ public:
 
 protected:
     bool using_wayland;
+    bool console_mode = false;
 
 private:
     wxPathList config_path;


rkitover:
I tried setting a breakpoint on `exit()` or `_exit()` in `gdb` and did not see anything helpful so I gave up, thank you for looking into this!

In the case of debug output being spewed when e.g. `--help` is given, we can just disable the wx logger in those cases.

Being able to sanity-check the binary in the CI would be wonderful, actual tests would be even better but that's a long ways away.

rkitover:
Testing your code, the `--help` exit code is fixed, and this is wonderful, one annoying thing is that there's a delay before the program exits, but we can fix this later.

denisfa:
So, what do we want for wrong or misspelled parameters? Can it stay this way?

> there's a delay before the program exits

This is because we initialize all the WX structure. The exit happens on the first iteration. Not sure if there is much that we can do here, but I will search the docs for more info.

rkitover:
> So, what do we want for wrong or misspelled parameters? Can it stay this way?

Yes that's fine, it gives an error and shows help, and in this case the runtime is only about a third of a second.

With `--help` the runtime is about 2 seconds, a bit too long. Not a huge deal of course.

denisfa:
@rkitover So should we go back for the quick exit, but `255` code for the `--help`? I need to sync what we decide here with the SDL port `--help` (which is returning `255` now).

Also, do you know what happens if I commit again to this branch of mine? Can I create another PR or will it be waste of time? My approach is to always create a new branch to develop a feature, but I am curious as this case is an immediately follow up work.

rkitover:
> So should we go back for the quick exit, but 255 code for the --help? I need to sync what we decide here with the SDL port --help (which is returning 255 now).

No, we should fix it so it is both an exit code of `0` and runs quickly, but until then I would prefer the exit code of `0` to the quick exit so I can run a binary check on the travis CI.

> So should we go back for the quick exit, but 255 code for the --help? I need to sync what we decide here with the SDL port --help (which is returning 255 now).

In this case I did a rebase and merge, so if you rebase your branch on top of master and push more commits you can use it for a new PR. If I did a squash and merge, that would not work, and you would need to reset the branch to master to reuse it for a PR. E.g.:

```bash
git reset --hard origin/master && git push -f
```

rkitover:
I tried re-enabling the `--help` check on travis, for ubuntu and mac, and they seem to be timing out for some reason (the timeout is 10m) so I disabled them for now.

denisfa:
> In this case I did a rebase and merge, so if you rebase your branch on top of master and push more commits you can use it for a new PR. If I did a squash and merge, that would not work, and you would need to reset the branch to master to reuse it for a PR

Thanks for this info!

> I tried re-enabling the --help check on travis, for ubuntu and mac, and they seem to be timing out for some reason (the timeout is 10m) so I disabled them for now.

I think I found the reason for this. All my testing, even in containers (LXC), had access to the DISPLAY and could launch GUI applications. When I removed this, the container execution never finishes while using the `--help`.
```
$ xvfb-run ./visualboyadvance-m --help
VisualBoyAdvance-M

Usage: visualboyadvance-m [-h] [--verbose] [--save-xrc <str>] [--save-over <str>] [--print-cfg-path] [-f] [-s] [-o] [ROM file] [<config>=<value>...]
  -h, --help               	show this help message
  --verbose                	generate verbose log messages
  --save-xrc=<str>         	Save built-in XRC file and exit
  --save-over=<str>        	Save built-in vba-over.ini and exit
  --print-cfg-path         	Print configuration path and exit
  -f, --fullscreen         	Start in full-screen mode
  -s, --delete-shared-state	Delete shared link state first, if it exists
  -o, --list-options       	List all settable options and exit

(visualboyadvance-m:111): Gtk-WARNING **: 21:05:15.836: gtk_window_present_with_time() should not be called with 0, or GDK_CURRENT_TIME as a timestamp, the timestamp should instead be gathered at the time the user initiated the request for the window to be shown
```

This never stops running. I will try to solve this and create a MR.

denisfa:
Of course, when running without `xvfb-run`, we have:
```
$ ./visualboyadvance-m --help
Unable to init server: Could not connect: Connection refused
21:15:30: Error: Unable to initialize GTK+, is DISPLAY set properly?
```

Issue #318, opened 2018-10-17
VBA-M Start up Message(s)


ghost:
Whenever I launch VBA-M I get a red message and it appears every ten seconds , that happens when i launch a game (any)

How do i disable that ? In an older/different version of VBA had a option of disabling it by clicking 

Options-> Emulator-> "Disable status messages" and that would disable the red message(s) from popping up

rkitover:
What does the message say?

goldenwarriorgw:
Well @rkitover 
![screenshot 2018-10-18 12 43 45](https://user-images.githubusercontent.com/33986489/47170442-05fa7580-d2d4-11e8-863b-41552155a22d.png)

rkitover:
That looks like some kind of unicode decoding issue.

As a temporary workaround, put the rom in a different directory.

ZachBacon:
I'd suggest actually renaming the rom, or better yet, save a copy and then paste it with a new name, That should confirm if it's a unicode decoding issue or not.

theblackwidower:
That doesn't really solve the issue. Because I'm getting the same problem, only it's not an error message, it's a message saying it did a battery save. On some games, this happens quite frequently, and is quite annoying. I'd like to not have a message pop up on the screen every time it undergoes normal operation.

rkitover:
I think we could remove these messages.

rkitover:
@theblackwidower I've removed the "wrote battery" message in master.

aleksrbk:
I got this issue now, saying in big red words loaded battery.
![bug](https://user-images.githubusercontent.com/42532740/108438717-247c2d00-7250-11eb-9269-ab1e61ab40f6.PNG)

Squall-Leonhart:
That's not an issue.

Squall-Leonhart:
>That looks like some kind of unicode decoding issue.
@rkitover

This unicode glitching is just a visual error, the writing error is due to the user having controlled folder access enabled, without whitelisting the emulator to write into this location.

Closing this, because an option has been available for some time

aleksrbk:
Thank you for the help

PR #389, opened 2019-03-16
XDG followup work.

denisfa:
@rkitover This is another followup work I had to do after #94. There was some stuff bugging me at that time, like whatever option to save file I had on GUI (screenshot, recording etc) was having the default option the same path from the configuration file, where `vbam.conf` was. 

This is not an issue for Windows and MacOS, but for Linux the appropriate path would be the `$XDG_DATA_HOME/visualboyadvance-m`, hence what I am doing here.

Another thing that I had to do was to sync these location with the ones from the SDL port. I am 99% sure that these type of `else` (example) were unreachable:
```
if (batteryDir)
...
else if (homeDir)
...
else // <- useless, it would never reach here
```

Now, all these locations are synced with the ones from the WX. The order is:
1. StateDir / BatteryDir / configuration file specific dir;
2. The path of the current loaded game;
3. XDG Base Dir (or equivalent) fallback.

I tested everything on Linux. I will update this if I find any issues on other platforms. My testing was like this: I had the same game on two places: `/opt/loz.zip` and `~/Downloads/loz.zip`.

1. Used `/opt/loz.zip`, with my regular user, on GUI and SDL port. I expected to save stuff on `${XDG_DATA_HOME}/visualboyadvance-m` (OK) and the default option for the gui when asking where to save files to be this location (OK).

2. Used `~/Downloads/loz.zip`, with my regular user, on GUI and SDL port. I expected to save stuff on `~/Downloads/` (OK) and the default option for the gui when asking where to save files to be this location (OK).

So it was very straightforward. Check If I still need to do something and, if I need, do not merge this and let's continue what is missing here before merging.

### DIFF
--- a/src/sdl/SDL.cpp
+++ b/src/sdl/SDL.cpp
@@ -64,6 +64,9 @@
 #include "inputSDL.h"
 #include "text.h"
 
+// from: https://stackoverflow.com/questions/7608714/why-is-my-pointer-not-null-after-free
+#define freeSafe(ptr) free(ptr); ptr = NULL;
+
 #ifndef _WIN32
 #include <unistd.h>
 #define GETCWD getcwd
@@ -222,15 +225,6 @@ uint32_t screenMessageTime = 0;
 #define SOUND_MAX_VOLUME 2.0
 #define SOUND_ECHO 0.2
 #define SOUND_STEREO 0.15
-
-void freeSafe(void *ptr)
-{
-    if (ptr != NULL)
-    {
-	free(ptr);
-	ptr = NULL;
-    }
-}
 
 static void sdlChangeVolume(float d)
 {
--- a/src/sdl/SDL.cpp
+++ b/src/sdl/SDL.cpp
@@ -223,6 +223,15 @@ uint32_t screenMessageTime = 0;
 #define SOUND_ECHO 0.2
 #define SOUND_STEREO 0.15
 
+void freeSafe(void *ptr)
+{
+    if (ptr != NULL)
+    {
+	free(ptr);
+	ptr = NULL;
+    }
+}
+
 static void sdlChangeVolume(float d)
 {
     float oldVolume = soundGetVolume();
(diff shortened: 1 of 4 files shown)


src/wx/panel.cpp
```diff
@@ -461,6 +461,10 @@ void GameArea::recompute_dirs()
         batdir = wxGetApp().GetAbsolutePath(gopts.battery_dir);
     }
 
+    if (!wxIsWritable(batdir)) {
+        batdir = wxGetApp().GetAbsolutePath(get_xdg_user_data_home() + DOT_DIR);
```
rkitover: Can we make a convenience method on the app for this, e.g. `GetConfigDir()` or something like that.

src/wx/wxvbam.cpp
```diff
@@ -750,7 +750,10 @@ wxString MainFrame::GetGamePath(wxString path)
         game_path = wxFileName::GetCwd();
 
     if (!wxIsWritable(game_path))
-        game_path = wxGetApp().GetConfigurationPath();
+    {
+	game_path = wxGetApp().GetAbsolutePath(get_xdg_user_data_home() + DOT_DIR);
```
rkitover: Likewise.

src/sdl/SDL.cpp
```diff
@@ -346,6 +352,32 @@ char* sdlGetFilename(char* name)
     return filebuffer;
 }
 
+char* sdlGetFilePath(char* name)
+{
+    static char filebuffer[2048];
```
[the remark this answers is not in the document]
rkitover: Also this is C++ even though it's written like C, so you can use things like `std::string` etc..

src/sdl/SDL.cpp
```diff
@@ -346,6 +352,32 @@ char* sdlGetFilename(char* name)
...
+
+    while (true) {
+        if (*p == FILE_SEP) {
+            break;
+        }
+        len--;
+        p--;
+        if (len == 0)
+            break;
+    }
```
rkitover: This is the sort of horrible C-style string code I'd prefer to work on removing not adding more of.

Could you use something like this instead:

https://stackoverflow.com/a/8518793/262458

src/sdl/SDL.cpp
```diff
@@ -760,36 +793,39 @@ void sdlWriteBackupStateExchange(int from, int to, int backup)
...
     else
-        sprintf(buffer, "%s.sav", filename);
+        sprintf(buffer, "%s%c%s.sav", homeDataDir, FILE_SEP, gameFile);
 
-    emulator.emuWriteBattery(buffer);
+    bool result = emulator.emuWriteBattery(buffer);
 
-    systemScreenMessage("Wrote battery");
+    if (result)
+	systemScreenMessage("Wrote battery");
```
rkitover: We probably don't want these "wrote battery" messages anymore.

See #318 

We could send them to the console instead perhaps.

src/sdl/SDL.cpp
```diff
@@ -760,36 +793,39 @@ void sdlWriteBackupStateExchange(int from, int to, int backup)
...
-        sprintf(buffer, "%s.sav", filename);
+        sprintf(buffer, "%s%c%s.sav", homeDataDir, FILE_SEP, gameFile);
 
-    bool res = false;
+    bool result = emulator.emuReadBattery(buffer);;
 
-    res = emulator.emuReadBattery(buffer);
-
-    if (res)
+    if (result)
```
rkitover: See above.

src/sdl/SDL.cpp
rkitover: Not a necessity for this PR of course, as it's unrelated.

rkitover:
Thank you, I left some comments on the code.

src/sdl/SDL.cpp
denisfa: I think I got it. I have an idea using only C, though. Check if you like it.

src/sdl/SDL.cpp
denisfa: It is easier to do this now rather than later.

src/sdl/SDL.cpp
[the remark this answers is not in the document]
denisfa: My first proposition for this is still C. Most of this code does not even consider the C++ structures. Refactoring everything is an option, but I feel we could have more discussion about this.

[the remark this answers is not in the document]
mudlord: Guess we could. The code could do with a overhaul.

src/sdl/SDL.cpp
```diff
@@ -223,6 +223,15 @@ uint32_t screenMessageTime = 0;
...
 
+void freeSafe(void *ptr)
+{
+    if (ptr != NULL)
+    {
+	free(ptr);
+	ptr = NULL;
+    }
+}
+
```
rkitover: This is wrong for two reasons:

- you don't need to check for `NULL` when calling `free()`
- to set a pointer variable, you have to use a pointer to a pointer and pass the address of the pointer

So you could do something like this:

```cpp
void freeSafe(void** ptr)
{
    free(*ptr);
    *ptr = NULL;
}
...
...
    freeSafe(&some_ptr);

rkitover:
Other than the one thing I noted everything else looks great, thank you.

src/sdl/SDL.cpp
denisfa: Damm, I always took this for truth... I remember using this code for a long time. Oh well, I guess it is what it is.

denisfa:
Ok, I have an alternative ready. See if it is enough.

rkitover:
Yup, that works!

Issue #79, opened 2017-02-19
Remember window's size/state/position upon restart


msmigiel:
### Expected behavior

It would be nice to have the application remember its size, state (whether maximized or not), and position upon being restarted. Please consider adding this, at least as an option.

### Actual behavior

The application defaults to a specific size and state and a somewhat random position.

### Steps to reproduce the behavior

1.) Load the application.
2.) Change the size/state/position from the default.
3.) Close and reopen it.
4.) Note the default settings are used.

### Build details

Operating System (Windows, Mac, Linux, etc.): Windows

Interface (wx, SDL, default is wx): wx

Version of code (pre-built binary version, or commit ref, or just "master"): 6ecab80.... (released 2/2/17)

Any other relevant build information:

If this is e.g. a joystick or video issue, then the relevant hardware information:

ZachBacon:
You're opening a lot of issues that are related to each other. Open one with everything you find next time if it's related to issues like this.

msmigiel:
I'm sorry. I don't understand how these issues are related and I thought it would be easier to track them separately, but if you would prefer I will open one issue going forward.

ZachBacon:
@msmigiel It's because they are all related to configuration changes a while back which also affect various things you are reporting on. So they are all related.

denisfa:
@rkitover This can be closed too, right?

rkitover:
Yes. Implemented by @denisfa in master.

rkitover:
@denisfa so this doesn't work yet, when you start the emu with a ROM it first switches to the remembered size then shrinks back to the default size, can you please take a look if and when you have a chance?

retro-wertz:
core will  resize based on default magnification size and will not use
remembered size....

rkitover:
Pretty sure this has nothing to do with the core, the core only calls `systemDrawScreen()` see `src/wx/sys.cpp` for the interface to the core.

denisfa:
@rkitover I know how to fix this. I will offer a quick solution in a moment.

PR #401, opened 2019-04-03
Fix for resizing window geometry when loading games.

denisfa:
@rkitover The window was being forced to resize every time a game loads. The idea is to resize only if the minimum size (determined by the zoom configuration) requires a window bigger than the custom one set by the user. In this case, we should set it to the minimum size asked by the zoom.

@retro-wertz I believe this should be enough to deal with the zoom. If I understand it right, the zoom level X means the minimum size for the window is `X * (min_width, min_height)`. Is this correct?

There is still something bothering me here. The window size set by the frame does not include status bar or menu bar. There might be some cases where the window will be either a little bigger than expected (the status bar pixels) or be slight incorrect `X,Y` starting positions. I will try some more tests to see If I can confirm this behaviour.

Fix for #79.

### DIFF
--- a/src/wx/panel.cpp
+++ b/src/wx/panel.cpp
@@ -302,7 +302,8 @@ void GameArea::LoadGame(const wxString& name)
     loaded = t;
     SetFrameTitle();
     SetFocus();
-    AdjustSize(true);
+    // Use custom geometry
+    AdjustSize(false);
     emulating = true;
     was_paused = true;
     MainFrame* mf = wxGetApp().frame;


retro-wertz:
the frame should pad the size of menu bar/status bar. this should avoid the orignal game's frame to break the original ratio, but i think the frame anyways is resizing by ratio (unintentionally probably) frame looks smaller than it should but it does not go out of scale (no stretching when resizing)

rkitover:
This works much better than previously, thank you.

I do notice the size is very slightly different between launches:

```
11:54:34: Debug: GL VIEWPORT: 0, 0, 1444, 963
```

and then on subsequent launch:

```
11:55:09: Debug: GL VIEWPORT: 0, 0, 1445, 996
```

denisfa:
I am trying to understand these and I believe I have a clue. The position `(0,0)` is actually not working for the restore, except for fullscreen or maximized window. 

Whenever I set the game frame out of bounds, the GUI just "fixes" its position on my screen on next run. The GUI also changes if I set a higher zoom level (>= 4.0) and my custom geometry is not big enough. This happens without running a game.

I think the zoom function is adjusting the GUI size somehow, but not only resizing the frame, but also the window (increases the frame size to fit the zoom level, then double it for the window). At least that's the pattern I got from the debug messages.

I will try some stuff and update here, if necessary.

rkitover:
I've worked a lot on this code, if you have any questions about where what is in the code or how anything works let me know.

denisfa:
@rkitover Thanks! I will try some stuff I have planned and then ask what I don't understand.

denisfa:
@rkitover I believe I understand what is happening now. There are 2 issues:

1. the screen is not big enough for the zoom level configured and the app is initialized. This causes a random behaviour when trying to create the window. It does not respect the geometry, but the zoom level requirement. So it places a custom size at any place of the screen.

2. The resolution is not enough to fit the zoom level. I have a 1920x1080 and a menubar that takes the first 64 pixels from the bottom of my screen. When I configure a zoom level >= 6.0 and open the app, the place is random and the width,height are the ones from the zoom level. (the status bar is hidden behind my menubar)

There are also 2 exceptions for this "bug": the maximized and fullscreen modes. If any of them are set, then the zoom level requirement is ignored and we get the desired geometry. This is probably due to the fact that these modes take over the control of the resizing/zooming level and place a custom one instead.

Does any of this sound weird to you? I am not sure what to do here.

rkitover:
I think that's fine because that would rarely happen in practice and is easy for the user to fix.

The issue that I see is that the size does not exactly match on close/restore, see above from the debug log.

rkitover:
Upon further testing, I see that while the subsequent startup geometry may be slightly off, once at the new value it will persist upon subsequent restarts. And the difference is minor enough to not be noticeable to the user.

The fullscreen behavior is also correct.

I think this all works well enough that we can close this now.

denisfa:
@rkitover Can you send me your config, alongside the game you are trying? I am testing this with some games and I can't get this error.

rkitover:
It's not an error, and you have to compile with `-DCMAKE_BUILD_TYPE=Debug` to see the GL viewport messages. But anyway, it really does work well enough now to not bother with anymore for the time being.

Thank you for doing this work.

I will prepare a new release now because we need to get the analog stick fix out ASAP.

denisfa:
I see the GL viewport, but it is not changing its value on subsequent startup for me. This could be something easy to fix. 

> I will prepare a new release now because we need to get the analog stick fix out ASAP.

Release a minor one with the analog stick, since it is important. It is not needed to carry other stuff.

rkitover:
>
> I see the GL viewport, but it is not changing its value on subsequent
> startup for me. This could be something easy to fix.
>
I was testing in Windows, that May have something to do with it, also try
some different window sizes.

Also the difference is very minor and the new value is stable.

>

rkitover:
I think master is in a stable enough state to do a minor release now.

PR #396, opened 2019-03-27
Fix compilation warnings.

denisfa:
@rkitover Hey! I have been busy lately, but I will try to contribute some more!

This is a bunch of code changes regarding getting rid of compilation warnings. This was also an excuse for me to browse the code in the last few days and see how some stuff work. I am not sure if you are interested or not on this, but nevertheless I have a few doubts about where should I go next. 

So these were the commands ran.
```
$ cmake .. -DENABLE_SDL=ON -DENABLE_LIRC=ON -DENABLE_LTO=ON -DENABLE_OPENAL=ON -DENABLE_SSP=ON -DCMAKE_BUILD_TYPE=Debug -DENABLE_ASAN=ON
$ make -j`nproc` &> warnings.txt
```

Next is the output `warnings.txt`. I will ask some questions.
```
Scanning dependencies of target fex
[  0%] Building C object fex/CMakeFiles/fex.dir/7z_C/7zAlloc.c.o
[  1%] Building C object fex/CMakeFiles/fex.dir/7z_C/7zBuf.c.o
[  1%] Building C object fex/CMakeFiles/fex.dir/7z_C/7zCrc.c.o
[  2%] Building C object fex/CMakeFiles/fex.dir/7z_C/7zCrcOpt.c.o
[  2%] Building C object fex/CMakeFiles/fex.dir/7z_C/7zDec.c.o
[  3%] Building C object fex/CMakeFiles/fex.dir/7z_C/7zIn.c.o
[  3%] Building C object fex/CMakeFiles/fex.dir/7z_C/7zStream.c.o
[  4%] Building C object fex/CMakeFiles/fex.dir/7z_C/Bcj2.c.o
[  4%] Building C object fex/CMakeFiles/fex.dir/7z_C/Bra86.c.o
[  5%] Building C object fex/CMakeFiles/fex.dir/7z_C/Bra.c.o
[  5%] Building C object fex/CMakeFiles/fex.dir/7z_C/CpuArch.c.o
[  6%] Building C object fex/CMakeFiles/fex.dir/7z_C/Lzma2Dec.c.o
[  6%] Building C object fex/CMakeFiles/fex.dir/7z_C/LzmaDec.c.o
[  7%] Building C object fex/CMakeFiles/fex.dir/7z_C/Ppmd7.c.o
[  7%] Building C object fex/CMakeFiles/fex.dir/7z_C/Ppmd7Dec.c.o
[...]

PR #402, opened 2019-04-03
Fix compilation warnings for MacOS build.

denisfa:
@rkitover The work we needed to do after #396. I need some checks here.

Some of these files do not include any headers except local ones. That is the reason why I copy & paste the solution for many files. I know there were memory leaks issues, but they came after compilation warnings so I think we can put these two together here.

### DIFF
--- a/src/filters/2xSaI.cpp
+++ b/src/filters/2xSaI.cpp
@@ -157,10 +157,10 @@ static inline uint32_t INTERPOLATE (uint32_t A, uint32_t B)
 
 static inline uint32_t Q_INTERPOLATE (uint32_t A, uint32_t B, uint32_t C, uint32_t D)
 {
-  register uint32_t x = ((A & qcolorMask) >> 2) +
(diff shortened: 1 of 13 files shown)


retro-wertz:
i would take out register keyword entirely

rkitover:
> i would take out register keyword entirely

I'm fine with either solution.

@denisfa I can merge this if you are ready.

denisfa:
I made some checks and it seems that this `register` qualifier is only actually used by old compilers... I also like the idea of removing this code rather than adding. If we can agree on this (use newer compilers), I will do a cleanup here.

rkitover:
It will still work fine on older compilers without it, feel free to remove it.

denisfa:
Alright, now it is done @rkitover.

denisfa:
> If I may give you some more annoying advice, you should start learning the art of proper git commit messages

No problem, I will be working on writing better commit messages from now on.

PR #408, opened 2019-04-07
Fix incorrect starting video showing and geometry.

denisfa:
@rkitover @retro-wertz This is a little trick because we need to decide a behaviour.

The reason such thing happened was  with option `Display/Stretch=1`. Changing on file `panel.cpp`:

```
GetSizer()->Add(w,    frame_priority, gopts.retain_aspect ? (wxSHAPED | wxALIGN_CENTER_HORIZONTAL | wxEXPAND) : wxEXPAND);

to

GetSizer()->Add(w,    frame_priority, gopts.retain_aspect ? (wxALIGN_CENTER_HORIZONTAL | wxEXPAND) : wxEXPAND);
```

The `wxSHAPED` is behind this. From the docs: `wxSHAPED flag tells the window to change it is size proportionally, preserving original aspect ratio`. When we had a default startup, the zoom level decided the window size (by setting the minimal one required for it). Then you could expand however you like it and it would work. When trying a custom size, the window would be a different size from the minimal one, and this is where this weird (for me) issue happen.

If you guys don't like this solution (I don't because it feels like a blob), I would suggest to drop this geometry support entirely. There is a better way, but just requires more time of research and docs hunting.

### DIFF
--- a/src/wx/panel.cpp
+++ b/src/wx/panel.cpp
@@ -1056,6 +1056,8 @@ void GameArea::OnIdle(wxIdleEvent& event)
 #endif
 
         Layout();
+        SendSizeEvent();
+
 
         if (pointer_blanked)
             w->SetCursor(wxCursor(wxCURSOR_BLANK));
--- a/src/wx/panel.cpp
+++ b/src/wx/panel.cpp
@@ -1046,9 +1046,9 @@ void GameArea::OnIdle(wxIdleEvent& event)
         // On windows with the vcpkg version of wxWidgets which is 3.1.2, the
         // wxEXPAND flag throws an XRC error, but it is necessary on earlier versions of wxWidgets
 #if defined(__WXMSW__) && wxCHECK_VERSION(3, 1, 2)
-        GetSizer()->Add(w,    frame_priority, gopts.retain_aspect ? (wxSHAPED | wxALIGN_CENTER_HORIZONTAL) : wxEXPAND);
+        GetSizer()->Add(w,    frame_priority, gopts.retain_aspect ? (wxSHAPED | wxALIGN_CENTER) : wxEXPAND);
 #else
-        GetSizer()->Add(w,    frame_priority, gopts.retain_aspect ? (wxSHAPED | wxALIGN_CENTER_HORIZONTAL | wxEXPAND) : wxEXPAND);
+        GetSizer()->Add(w,    frame_priority, gopts.retain_aspect ? (wxSHAPED | wxALIGN_CENTER | wxEXPAND) : wxEXPAND);
 #endif
 
 #if wxCHECK_VERSION(2, 9, 0)
--- a/src/wx/wxvbam.cpp
+++ b/src/wx/wxvbam.cpp
@@ -763,22 +763,23 @@ void MainFrame::OnMenu(wxContextMenuEvent& event)
 void MainFrame::OnMove(wxMoveEvent& event)
 {
     (void)event; // unused params
-    wxRect pos = GetRect();
-    int x = pos.GetX(), y = pos.GetY();
+    wxPoint pos = GetScreenPosition();
+    int x = pos.x, y = pos.y;
     if (x >= 0 && y >= 0 && !IsFullScreen())
     {
 	windowPositionX = x;
 	windowPositionY = y;
-	update_opts();
     }
+    update_opts();
 }
 
 void MainFrame::OnSize(wxSizeEvent& event)
 {
     wxFrame::OnSize(event);
     wxRect pos = GetRect();
+    wxPoint windowPos = GetScreenPosition();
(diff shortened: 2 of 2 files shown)


retro-wertz:
everything is now stretched with that regardless of "retain aspect ratio"
in runtime. cant you just force re-update the size once when a game is
loaded to similate changing window size so it can update based on the
current settings? at least for now instead of dropping the whole thing
entirely? dunno a thing about wx but on libretro we refresh size when
something in geometry is changed from its last config during runtime.

denisfa:
@retro-wertz Do you mind testing the new code? I am on Linux x64 and it fixes for me.

> cant you just force re-update the size once when a game is
loaded to similate changing window size so it can update based on the
current settings? at least for now instead of dropping the whole thing
entirely? dunno a thing about wx but on libretro we refresh size when
something in geometry is changed from its last config during runtime.

I found a dummy function to trigger the resize.

> update:
minimizing windows then restoring the window fixes the issue. looks like this is calling OnSize during this minimize/restore or window, maybe calling this OnSize once during the mainloop (or where emusys->emuMain(emusys->emuCount) is) would work as well.

This is an issue that I am being tilted about it. For me (Linux), this did not work. I can't properly debug this on other platforms. So It can work for me, but not for other. Likewise, it can work for others, but not for me.

> update2:
changing display filters also causes this stretching..

On current master? This is intended, right? The display filter has to take over to do what it is supposed to do.

denisfa:
@rkitover I will rebase this and write a proper commit message, don't worry!

retro-wertz:
no more stretching so far. but image is not centered vertically:
[image: Screenshot_2019-04-09_01-28-22.png]
- load game
- increase vertical size
- if image stays centered, switch to fullscreen, then go back to window
mode

tested on linux, aspect_ratio ON

denisfa:
@retro-wertz Thanks for the test, but I believe this should not be an issue caused by the geometry change. Check the original code on the second commit, it has not been changed by me. Anyway, since it looks easy to fix, here it is.

retro-wertz:
image still stays at the bottom. i know its not something geometry-related, but positioning still an issue. anyways ill let someone with better understanding in wx and programming than me decide this. im ok with it so far than the last behaviour. just let me know if i have something to test...

Screenshot_2019-04-09_02-30-35

denisfa:
@retro-wertz This is actually a bigger issue than what it looks like. When I go full screen and come back, the image stays at the bottom, but not only for the current game, also for the entire app run. It is not fixed when switching game. There must be a custom behaviour somewhere. I will investigate this.

Do you remember if this was an issue before the geometry change?

denisfa:
Alright, I have my report now.

Version `2.1.1` had this exactly same issue about the position (test case described by @retro-wertz). So the geometry feature did not introduce it. That is a big relief for me. I investigated what is the cause of this issue. It is this line on `panel.cpp` (from ` 2.1.1` source code) :

```
GetSizer()->Add(w,    frame_priority, gopts.retain_aspect ? (wxSHAPED | wxALIGN_CENTER_HORIZONTAL | wxEXPAND) : wxEXPAND);
```

Here the interaction between `wxSHAPED` and `wxEXPAND` is the root of all evil. While not fullscreen, the app is resized keeping the proportions (`wxSHAPED`). When fullscreen, the whole display is filled with the game area; there is no black space draw (`wxEXPAND`). After we quit fullscreen, the app is then resized to keep `wxALIGN_CENTER_HORIZONTAL`, but not the vertical one. When trying to add `wxALIGN_CENTER_VERTICAL` or `wxALIGN_CENTER`, the app just ignores it and we have our issue again.

I believe we have some options about this:

1. Keep the issue, since it is niche case (although it persists between games in a single execution);
2. Remove `wxEXPAND` from `(wxSHAPED | wxALIGN_CENTER_HORIZONTAL | wxEXPAND)`. The proportions are kept in fullscreen, but there is empty space and it depends on the size of the game area screen when you hit fullscreen (from maximized it is bigger than any other size);
3. Add our customized code, where ideally we would use `wxEXPAND` when fullscreen and `wxSHAPED | wxALIGN_CENTER` everywhere else. I have no idea how to do this.

What do you think @retro-wertz , @rkitover?

rkitover:
This line of code has already caused me a good deal of grief when I was doing visual studio support, because it throws an XRC error dialog about `wxEXPAND` not being allowed with that combination of flags.

But taking out the `wxEXPAND` does not work when using `gtk3` at least, then the game area never expands with the window.

I will try to look at all of this over the next few days and figure out some solution.

Perhaps I can just make a custom sizer subclass for this.

rkitover:
Also the `wxEXPAND` is not present on windows because we are using wxWidgets `3.1.2` in the mingw build script.

rkitover:
These changes look good though, so I will merge this now and look at remaining issues later.

Thanks @denisfa!

Issue #390, opened 2019-03-17
Weird Sticky Keys Issue


sorin255:
### Expected behavior
When movement keys are released, that key's input will stop being read.

### Actual behavior
When a movement key is pressed, often after being held for a bit, movement in that direction will continue until the key is pressed again.

### Steps to reproduce the behavior
First noticed in Fire Red. Tried changing keys to see if it was just an issue with my directional keys but it was the same experience. Made sure autofire and turbo were cleared and turned off.
Tested in other games: Sonic Advance, then Donkey Kong Land (GB) and was having the same issues there.
Didn't seem to be having the issue with my joypad.

### Crash bugs

If this is a crash bug, and you have the time and inclination, please provide a
symblic backtrace.

Instructions for doing so are
[here](https://github.com/visualboyadvance-m/visualboyadvance-m/blob/master/README.md#reporting-crash-bugs).

### ROM details

Which games are affected by the issue (please state if GBA or GB game): Pokmeon Fire Red (GBA), Sonic Advance (GBA), Donkey Kong Land (GB)

Which games, if any, are NOT affected by the issue (please state if GBA or GB game): none that I can tell

Include screenshots of ROM Information (File > ROM Information):

### Option details

Video driver (simple, opengl): OpenGL

Sound driver (directsound, xaudio, openal, SDL): OpenAL

### Build details

Operating System (Windows, Mac, Linux (state distribution), etc.): Linux (Ubuntu 18.04)

If using Linux, specify if using xorg or Wayland: xorg

Version of code (pre-built binary version, or commit ref, or just "master"):

### Hardware details:

If your joystick or gamepad isn't working in Linux, please try to troubleshoot
using the information on this page:
https://wiki.archlinux.org/index.php/Gamepad **BEFORE** filing an issue. We use
SDL for joystick support, any problems will generally be an issue with SDL and
**NOT** the emulator.

If relevant, the joystick you are using and the drivers you are using for it: xbox one controller 2017 firmware, just default xpad driver

rkitover:
Which keys did you map for movement? arrows?

sorin255:
Arrows initially and then WASD when replicating the issue.

rkitover:
Thanks, I'll see if I can reproduce this.

shibatan:
Wanted to chime in that I'm experiencing this (also in Fire Red) but exclusively while using Throttle.

rkitover:
Please try latest nightly.

[...]

PR #427, opened 2019-05-22
Fix #142 and #390.

### LINKED ISSUE #142: Button presses getting stuck when using keyboard
### Expected behavior

Once a key on the keyboard which is configured to a certain button is released, it shouldn't be registered as released and not being pressed/held anymore.

### Actual behavior

Once releasing a certain key (say, left directional, which is configured to the left button), at most times the corresponding button is also registered as not being pressed anymore. However, at certain times (which seem to be random), even after being released, the button corresponding to the key which was pressed still registers as if it was being pressed/hold. If the key is pressed and released again, its release is registered and everything comes back to normal.

I've tried reproducing the problem with a Dualshock 3 (aka PS3 controller) but it seems to not happen with it. Also, something curious is that if I press and release the button on the DS3 which corresponds in the input options to the same button which got "stuck" when using the keyboard, it doesn't get registered as released, as opposed to when doing the same in the keyboard itself.

Either press a key a certain number of times with some pauses between the presses to check whether its input got "stuck" or alternate between buttons until one of them gets stuck.  It seems to be easier to reproduce by pressing directional keys randomly on a menu screen with a lot of options which can be switched horizontally and vertically (such as a naming screen in a game, for example). Turbo mode may also help.

### ROM details

I experienced this with Harvest Moon: Friends of Mineral Town (US version). I was able to reproduce the problem in both the initial naming screen and while controlling the character at random. I also tried it with the english version of Golden Sun, in the initial naming screen, and was able to reproduce the problem, so I assume it's not a problem limited to a certain ROM.

### Option details

Video driver (simple, opengl or cairo): OpenGL

Sound driver (directsound, xaudio, openal, SDL): Problem happens with both OpenAL/SDL

### Build details

Commit 4d7e102e2650d9d66ac0ef49fac3869dd29df8da, built and tested on Arch Linux. Using xorg.

denisfa:
@rkitover I copied my commit message here. See if you can understand.

```
Fix continuous sprinting when using a modifier key as a mapping key (#142).

When holding a key(W,A,S,D) and then pressing a modifier(SHIFT,CTRL,ALT), we get two sets {MOD,MOD,0} and {KEY,MOD,0} with MOD > 0 after a {KEY,0,0}. When releasing the initial key while holding the modifier, we search for {KEY,MOD,0} instead of {KEY,0,0} to be removed.

Finally, when releasing the modifier, we remove {MOD,MOD,0} from our stack. This means the {KEY,0,0} stays "pressed", although no input is being taken from the keyboard.

Full example using WASD as movement keys and L_SHIFT as our B button.
1. Press and hold "A" --> add {A,0,0} to stack;
2. Press "L_SHIFT" --> add {L_SHIFT,4,0} and {A,4,0} to stack;
3. Release "A" --> remove {A,4,0} from stack (no effect);
4. Release "L_SHIFT" --> remove {L_SHIFT,4,0} from stack;
5. {A,0,0} still on stack.

No keyboard button is being pressed anymore, but there is a continuous movement towards the direction mapped by "A" {A,0,0}. It is only fixed when "A" is pressed two times in succession, because we don't add duplicate sets to our stack, but try to remove them anyway.
```

There is one big implication here that we should no longer support key combinations as input mappings. Not sure this is too much to give away (I don't use it, but there might be some use case that I am not aware about it -- using stuff as "SHIFT+W" to do something). I believe this should not be that big of deal because we would not need to check for conflict with stuff such as "SHIFT+F1".

Check if you like it.

### DIFF
--- a/src/wx/panel.cpp
+++ b/src/wx/panel.cpp
@@ -1183,6 +1183,9 @@ static bool process_key_press(bool down, int key, int mod, int joy = 0)
             mod |= wxMOD_RAW_CONTROL;
             break;
 #endif
+        default:
+            mod = 0;
+            break;
     }
 
     // check if key is already pressed


rkitover:
@denisfa this looks like a completely reasonable change for the time being, thank you!

Just one gripe for the future, please wrap the title of your commit messages to 50 columns and the text to 72 columns.

See if your editor has a profile for git commit messages to make this easier, vim does this automatically. The actual text of the message is very good, no complaints there.

I've been busy with stuff, got a contract on upwork, but I hope to do some more work on this repo too soon.

rkitover:
I will build a nightly for this now.

denisfa:
@rkitover The commit message has this formatting. [Check here](https://github.com/visualboyadvance-m/visualboyadvance-m/pull/427/commits/cfbfec5f1cc038b69caadcdfea03432ba2f1ef7e). I don't know why it appeared like this here.

Good luck on the contract! Can I keep tagging you to check for this MR? Also, you self-assigned to most of the tasks for `2.1.4`. Can I take a look at some of that? I want to have something else useful to work alongside the fix-recording stuff, since that is much more time consuming (as boring) than I thought.

rkitover:
Yeah I see now the wrapping is good on the message, title line should be 50 characters though.

Yeah of course, and reassign as needed.

Work on whatever is interesting for you, this is a volunteer project and we are under no obligation to anyone obviously.

PR #429, opened 2019-05-28
Fix #142.

### LINKED ISSUE #142: Button presses getting stuck when using keyboard

denisfa:
@rkitover I would like to know if the design is good. Take a look at where I am calling the functions and check if it needs to be changed. It has been some time since I worked on the frontend, so probably some code there could be improved.

I tested on a `ubuntu 18.04` on a VM and managed to consistently run into `stuck button after pause` and #142 issues. The solution fixed both for me and I got a positive feedback from one user.

One thing I am disliking is the static function created. There is some refactoring to be done. I will be taking a look at this.

### DIFF
--- a/src/wx/panel.cpp
+++ b/src/wx/panel.cpp
@@ -16,6 +16,9 @@
 #include "filters.h"
 #include "wxvbam.h"
 
+// release all buttons currently pressed
+static void clear_input_press();
+
 int emulating;
 
 IMPLEMENT_DYNAMIC_CLASS(GameArea, wxPanel)
@@ -919,6 +922,12 @@ GameArea::~GameArea()
     }
 }
 
+void GameArea::OnKillFocus(wxFocusEvent& ev)
+{
+    clear_input_press();
+    ev.Skip(true);
+}
+
 void GameArea::Pause()
 {
     if (paused)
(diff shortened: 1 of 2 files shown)


denisfa:
@rkitover It is good to go!

PR #435, opened 2019-06-05
[WIP] Enhance geometry.

### LINKED ISSUE #79: Remember window's size/state/position upon restart

denisfa:
@rkitover This is a follow up for #79 that I wanted to do, but it was very low priority. There were some problems that annoyed me a lot, but I was not sure about the solution.

The commit message is self explanatory; I would like to expand on something weird from `wxWidgets`. I wanted to also have a maximized state for the window, but in a more intuitive way. Let's say, I have starting coordinates `(100,100)` and size `(800,600)`. When I maximize (press the button) the window, the framework executes the events in the following order:

`EVT_MOVE` --> `EVT_MAXIMIZED` --> `EVT_SIZE`.

On the first event, the framework moves the window to position `(0,0)` while `IsMaximized()` returns `false`. This is saved in our configuration. So the position `(0,0)` is our last point, although it is very clear it isn't `(100,100)`. From the second event forward, we can query the `IsMaximized() == true` correctly. This is a particular case for the maximize effect, since outside of this we can correctly query on our function `OnMove`.

When you try to restore the window, only two events are played: `EVT_MOVE`  --> `EVT_SIZE`.

This means that if you close the app while maximized, your starting point will be `(0,0)` for the next one. I did not want that, but actually the position before the window is maximized. The size will be the correct one, since we do check on our `OnSize` if the window is fullscreen/maximized before updating values.

I will ask some questions on the code. After this is done, we should finally close #79.

As an extra, there is a [way to do this](https://docs.wxwidgets.org/trunk/overview_persistence.html) using `wxWidgets`, although only from `2.9.0` on. So this will be a candidate for refactoring when we drop support for `2.8`.


src/wx/opts.cpp
```diff
@@ -277,6 +277,7 @@ opt_desc opts[] = {
 
     /// Geometry
     INTOPT("geometry/fullScreen", "Fullscreen", wxTRANSLATE("Enter fullscreen mode at startup"), fullScreen, 0, 1),
+    INTOPT("geometry/isMaximized", "Maximized", wxTRANSLATE("Window maximized"), windowMaximized, 0, 1),
```
denisfa: How do we handle the new option? Does it need to be translated somewhere? I believe this should not appear anywhere but in `vbam.ini`.

src/wx/wxvbam.cpp
```diff
@@ -761,10 +766,20 @@ void MainFrame::OnMove(wxMoveEvent& event)
     (void)event; // unused params
     wxPoint pos = GetScreenPosition();
     int x = pos.x, y = pos.y;
-    if (x >= 0 && y >= 0 && !IsFullScreen())
+    if (!IsFullScreen() && !IsMaximized())
```
denisfa: This is the same check twice on Mac, but I don't think we need to have a macro check here to avoid one.

src/common/ConfigManager.h
```diff
@@ -133,8 +133,9 @@ extern int videoOption;
 extern int vsync;
 extern int wasPaused;
 extern uint32_t windowHeight;
-extern int windowPositionX;
-extern int windowPositionY;
+extern int windowMaximized;
+extern int windowPositionX, bkpPosX;
+extern int windowPositionY, bkpPosY;
```
denisfa: The `bkpPos{X,Y}` are meant to be one from last position of window. This achieves what I want, but I am not 100% sure if I really need them.

src/wx/opts.cpp
rkitover: You could also add it to `ConfigManager.{cpp,h}` but it's fine if it's not.

PR #439, opened 2019-06-11
Key bindings and shortcuts improvement.

denisfa:
@rkitover This is the fix for everything from that deleted user (whose issues were also erased). I believe each commit explains itself, so I will just post some comments on the code.

You should probably check the design, since I also had to do some `FIXME` from the code.

I will see if there is something related on our issues section.

### DIFF
--- a/src/wx/opts.cpp
+++ b/src/wx/opts.cpp
@@ -135,18 +135,22 @@ const wxString joynames[NUM_KEYS] = {
     wxT("Speed"), wxT("Capture"), wxT("GS")
 };
 
-wxJoyKeyBinding defkeys[NUM_KEYS * 2] = {
-    WJKB(WXK_UP), WJKB(1, WXJB_AXIS_MINUS, 1), WJKB(WXK_DOWN), WJKB(1, WXJB_AXIS_PLUS, 1),
-    WJKB(WXK_LEFT), WJKB(0, WXJB_AXIS_MINUS, 1), WJKB(WXK_RIGHT), WJKB(0, WXJB_AXIS_PLUS, 1),
-    WJKB(wxT('X')), WJKB(0, WXJB_BUTTON, 1), WJKB(wxT('Z')), WJKB(1, WXJB_BUTTON, 1),
-    WJKB(wxT('A')), WJKB(2, WXJB_BUTTON, 1), WJKB(wxT('S')), WJKB(3, WXJB_BUTTON, 1),
-    WJKB(WXK_BACK), WJKB(4, WXJB_BUTTON, 1), WJKB(WXK_RETURN), WJKB(5, WXJB_BUTTON, 1),
-    WJKB(WXK_NUMPAD_UP), WJKB(0), WJKB(WXK_NUMPAD_DOWN), WJKB(0),
-    WJKB(WXK_NUMPAD_LEFT), WJKB(0), WJKB(WXK_NUMPAD_RIGHT), WJKB(0),
-    WJKB(WXK_NUMPAD_PAGEUP), WJKB(0), WJKB(WXK_NUMPAD_PAGEDOWN), WJKB(0),
-    WJKB(wxT('W')), WJKB(0), WJKB(wxT('Q')), WJKB(0),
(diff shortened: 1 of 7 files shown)


src/wx/widgets/keyedit.cpp
```diff
@@ -124,10 +124,12 @@ wxString wxKeyTextCtrl::ToString(int mod, int key)
 
 #endif
 
-    if (s.empty() || (key != wxT('-') && s[s.size() - 1] == wxT('-'))
-                  || (key != wxT('+') && s[s.size() - 1] == wxT('+')))
+    if (s.empty() || (key != wxT('-') && s[s.size() - 1] == wxT('-') && s != wxT("Num -"))
+                  || (key != wxT('+') && s[s.size() - 1] == wxT('+') && s != wxT("Num +")))
```
denisfa: In case you are wondering, this was breaking `WXK_NUMPAD_SUBTRACT` and `WXK_NUMPAD_SUBTRACT`   bindings we had. Check how the it was being done before and how it was translated in `keyedit.h`.

src/wx/wxvbam.cpp
```diff
@@ -1192,24 +1212,9 @@ void MainFrame::IdentifyRom()
 // a few keys (e.g. only ctrl-x works for exit, but not esc & ctrl-q;
 // ctrl-w does not work for close).  It's possible another entity is
 // grabbing those keys, but I can't track it down.
```
denisfa: This is not Windows only... On Linux, I was trying to setup the `L` (no modifier) to open files, but it was being ignored. It seems to be something related to `wxGTK`. No idea what was causing this on Windows.

src/wx/opts.cpp
```diff
@@ -759,9 +762,11 @@ void update_opts()
     if (gopts.accels.size())
         cfg->SetPath(wxT("/Keyboard"));
 
+    int cmd_id = -1;
```
denisfa: Not sure if `-1` is the best value here... I needed to be something like "empty" and that should always be set on the first iteration of the loop. It works for now.

src/wx/opts.cpp
```diff
@@ -135,18 +135,22 @@ const wxString joynames[NUM_KEYS] = {
...
+    WJKB(WXK_SPACE), WJKB(0), WJKB(0)
+};
+
+wxJoyKeyBinding defkeys_joystick[NUM_KEYS] = {
+    WJKB(1, WXJB_AXIS_MINUS, 1), WJKB(1, WXJB_AXIS_PLUS, 1), WJKB(0, WXJB_AXIS_MINUS, 1), WJKB(0, WXJB_AXIS_PLUS, 1),
+    WJKB(0, WXJB_BUTTON, 1), WJKB(1, WXJB_BUTTON, 1), WJKB(2, WXJB_BUTTON, 1), WJKB(3, WXJB_BUTTON, 1),
+    WJKB(4, WXJB_BUTTON, 1), WJKB(5, WXJB_BUTTON, 1),
+    WJKB(0), WJKB(0), WJKB(0), WJKB(0),
+    WJKB(0), WJKB(0), WJKB(0), WJKB(0),
+    WJKB(0), WJKB(0), WJKB(0)
```
denisfa: I forgot to mention, but this removed some mappings. Basically, we keep the binding only for the standard buttons (the one you have on an actual GBA). Everything else was set to empty.

ZachBacon:
Before @rkitover signs off on this, I want to say, thanks for contributing to the project. Lately with my job as is, I haven't been able to put much work in and having people like you helping out, makes me glad I made the right call to move the project to github. You have my thanks for your contributions

denisfa:
@ZachBacon It's my pleasure. It has been a really good learning experience so far. I will try to contribute whenever I can.

src/wx/opts.cpp
rkitover: Is this the mapping for the xbox 360 controller? Removing the unnecessary bindings is good.

denisfa: This is the same from before for controller. I don't have one to check if these match.

PR #446, opened 2019-06-20
[WIP] Fix for unicode related issues.

denisfa:
@rkitover this is a quick fix, considering there still are some possible solutions available.

At first, when this issue first happened, I thought it very strange for something to break like this suddenly. Then I remembered that from `2.1.1.` to `2.1.2` there was that compilation warnings commit. I reviewed what I changed (already considering that as the cause), but failed to find the answer.

So I went to your `unicode-filenames` branch to test some stuff, but the compilation was failing with weird errors (that do not happen for current master). That's why I created this new branch and cherry-picked your commit from there. 

I will comment some of the code, but it should be fairly straightforward about why this works. It is probably not the best solution, though.

As far as testing goes, I tried opening all the files from the current zip on `Windows 10 x64`. They are all variations from a custom ROM with different names. I believe this is enough for the open issues.

[testing.zip](https://github.com/visualboyadvance-m/visualboyadvance-m/files/3309353/testing.zip)

### DIFF
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -569,7 +569,8 @@ IMAGE_TYPE utilFindType(const char *file, char (&buffer)[2048])
                 return IMAGE_UNKNOWN;
         }
         MultiByteToWideChar(CP_ACP, 0, file, -1, pwText, dwNum);
-        char *file_conv = fex_wide_to_path(pwText);
+        //char *file_conv = fex_wide_to_path(file);
+        char *file_conv = (char *)file;
         //	if ( !utilIsImage( file_conv ) ) // TODO: utilIsArchive() instead?
         //	{
         fex_t *fe = scan_arc(file_conv, utilIsImage, buffer);
@@ -578,7 +579,7 @@ IMAGE_TYPE utilFindType(const char *file, char (&buffer)[2048])
         fex_close(fe);
         file = buffer;
         //	}
-        free(file_conv);
+        //free(file_conv);
 #else
         //	if ( !utilIsImage( file ) ) // TODO: utilIsArchive() instead?
         //	{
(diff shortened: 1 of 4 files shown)


src/wx/panel.cpp
```diff
@@ -80,13 +80,13 @@ void GameArea::LoadGame(const wxString& name)
 
     // auto-conversion of wxCharBuffer to const char * seems broken
     // so save underlying wxCharBuffer (or create one of none is used)
-    wxCharBuffer fnb(fnfn.GetFullPath().mb_fn_str());
+    wxCharBuffer fnb(fnfn.GetFullPath().mb_str(wxConvUTF8));
```
denisfa: If I understand the situation correctly, we have an issue of ping-ponging a string between UTF-8, UTF-16 and whatever else Windows sends in our way. `wxString` can deal with these, but we needed to use across all files, which is not true since we have `const char *` and `wchar_t *` on multiple files. This is a way that I find on [docs](https://docs.wxwidgets.org/3.0/classwx_string.html#adcfd12e6d0765b1d74bccc3d63d02e98) to get a `UTF-8` string converted automatically and use as an input for `utilFindType`.

src/Util.cpp
```diff
@@ -569,7 +569,8 @@ IMAGE_TYPE utilFindType(const char *file, char (&buffer)[2048])
                 return IMAGE_UNKNOWN;
         }
         MultiByteToWideChar(CP_ACP, 0, file, -1, pwText, dwNum);
-        char *file_conv = fex_wide_to_path(pwText);
+        //char *file_conv = fex_wide_to_path(file);
+        char *file_conv = (char *)file;
```
denisfa: We already have a UTF-8 string here. The steps above seems unnecessary. Before, we had a multibyte string (UTF-16), that we get a `wchar_t *` from and then we convert into a `char *` UTF-8 string.

The key idea here is that the first string we have is `wxString`. We can use its functions to get the desired format. So when the code runs here, `const char *file` is already what we want(?).

denisfa:
Since this is a WIP, I did not write a proper commit message or tagged the issues reported. I will if this goes forward.

I think I have a good idea about the issue, but would like to understand more about it (my job is pretty much backends and server tools, so all this string codification problems are new to me).

rkitover:
The cause of this breakage is probably some refactoring I did to restore wX 2.8 compat.

Issue #411, opened 2019-04-10
add hotkeys for 10 save states


SLaks:
### Expected behavior

Load / Save State submenus in Key Shortcuts dialog should have entries for `1` through `10`

### Actual behavior

![image](https://user-images.githubusercontent.com/313894/55905706-896c9380-5ba0-11e9-9c5c-8543aa6951c4.png)

On Windows 10

RunisDauphin:
It does not matter which version of Windows you use: it is still present (in 2.1.3) in Windows 8.1 since 2.0.x, when vba-m.com looked completely different.

PR #450, opened 2019-06-24
Fix key shortcuts for save/load states.

denisfa:
@rkitover 

We show the wrong description for these shortcuts, although it works
correctly for the keybinding.

Original:
original

New:
new

I will explain why this happens on the code. Fix for #411.

### DIFF
--- a/src/wx/guiinit.cpp
+++ b/src/wx/guiinit.cpp
@@ -2244,13 +2244,28 @@ void MainFrame::add_menu_accels(wxTreeCtrl* tc, wxTreeItemId& parent, wxMenu* me
             // but then if the user overides main menu, that req. is broken..
             wxString txt = (*mi)->GetItemLabelText();
 
-            // It is not necessary unless we want the full description.
-            //for (int i = 0; i < 10; i++)
-            //    if (*mi == loadst_mi[i] || *mi == savest_mi[i]) {
-            //        txt = cmdtab[i].name;
-            //        break;
-            //    }
-
+            // we could probably have a global hashmap:
+            // cmdtab[i].cmd -> cmdtab[i]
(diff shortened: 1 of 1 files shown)


src/wx/guiinit.cpp
```diff
@@ -2244,11 +2244,12 @@ void MainFrame::add_menu_accels(wxTreeCtrl* tc, wxTreeItemId& parent, wxMenu* me
             // but then if the user overides main menu, that req. is broken..
             wxString txt = (*mi)->GetItemLabelText();
 
-            for (int i = 0; i < 10; i++)
-                if (*mi == loadst_mi[i] || *mi == savest_mi[i]) {
-                    txt = cmdtab[i].name;
```
denisfa: Here. We are using the wrong index. `cmdtab[i]` is not where the save/load state is. We needed to get the label (`LoadGame01`, for example), then go into the table and check which command has this name and then get the label.

denisfa:
The fact that it is only the number (not the actual description, like `Load saved state 1`) feels very intuitive for me. I know how to do to get this if you prefer, but I would like to see if you like this too.

rkitover:
I think I'd prefer a description like "Load save state in slot 1" and "Save state in slot 1", otherwise no one will know what this means.

denisfa:
@rkitover Here it is. I had a feeling that maybe the previous was enough because the numbers are inside the `Load State` menu. Same for the `Save State`.

new2

Just to confirm this, the text shown is the label from the `*.xrc` files, except the save/load states. This was being done before any of this, and continues to do so. The names are very similar to the labels (the label is usually shorter). Everything is working as expected.

rkitover:
Oh sorry, I missed the part that they were in the `Load state` menu, but this is much better anyway.

PR #456, opened 2019-06-29
Add hotkeys for changing sizes (1x ... 6x).

### LINKED ISSUE #415: Add a hotkey for changing sizes, such as mGBA and previous versions of VBA do
Have a hotkey for changing window size (for example, if want the window 4x the real size, I can press key "4" and it automatically changes, but if I want it later 2x the real size, simply I press "2" then and it reverts back).

### Actual behavior

You have to change it manually in Options -> Video -> Configure -> Zoom -> Default Magnification. Also, for some reason, there are like 40 pixels beneath the "game screen" (my vbam.ini file says window is 741 pixels tall) that are black.

denisfa:
Use a accelerator to quickly provide a way to change to/from multipliers
of the original size.

- Fix #415.

The only thing I would like to add here is some stuff is being done the way it is so we can also allow the users to change these keybindings via `Options` > `Key Shortcuts`. This explains whey this appears on the main menu.

### DIFF
--- a/src/wx/cmdevents.cpp
+++ b/src/wx/cmdevents.cpp
@@ -2102,6 +2102,42 @@ EVT_HANDLER(GameBoyConfigure, "Game Boy options...")
     update_opts();
 }
 
+EVT_HANDLER(SetSize1x, "1x")
+{
+    gopts.video_scale = 1;
+    panel->AdjustSize(true);
+}
+
+EVT_HANDLER(SetSize2x, "2x")
+{
+    gopts.video_scale = 2;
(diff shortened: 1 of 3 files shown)


retro-wertz:
does this have UI access too? i believe majority of people familiar with the older vba has been reporting these and some mission UI quick access to functions (resize for one etc)

denisfa:
@retro-wertz What do you mean? They are available in `Options` > `Video` > `Quick resize`. The submenu has `1x`,`2x`,...,`6x`.

Issue #437, opened 2019-06-08
Key Maps stay empty and cannot be set


simsemax:
### Expected behavior

Defined Key Mapping works while playing and is saved when VBA is restarted.

### Actual behavior

Manually configured key maps stay empty, even default mappings are not saved.
The first time I opened VBA everything worked fine, but the second and following times this keeps happening.

### Steps to reproduce the behavior
Start VBA and configure the keys maps in the input menu, save them but in the game they won't work. If you open up the input menu again, you will see why: the definitions of Up, Down, Left, Right, ABLR, Select and Start are empty except for the special ones, which are always (MotionUp=KP_UP, MotionDown=KP_DOWN, MotionLeft=KP_LEFT, MotionRight=KP_RIGHT, MotionIn=KP_PAGEUP, MotionOut=KP_PAGEDOWN, AutoA=W, AutoB=Q, Speed=SPACE).
Due to the fact that I use a laptop with no keypad, I can only use AutoA, AutoB and Speed.
Configs in vbam.ini can be reset manually but if I open VBA again, vbam.ini gets initially reset and overwritten with the key maps described above and the same problem occurs.

### Crash bugs

no crashes occurred.

### ROM details

Only tested "Legend of Zelda, The - The Minish Cap (Europe) (En,Fr,De,Es,It).gba" so far but this problem occurs even if no rom is loaded.

### Option details

Video driver: OpenGL

Sound driver: SDL

### Build details
Operating System : Mac OS X 10.10.5

Version of code (pre-built binary version, or commit ref, or just "master"): I tried the VBA versions 2.1.3, 2.1.2 and 2.1.1

### Hardware details:
Computer: MacBookPro9,2, Intel Core i7, 2,9 GHz, 16 GB RAM, Intel HD Graphics 4000 (1024GB VRAM)

Joystick: Gravis Gamepad Pro USB Model #4211, don't know the driver, Plug'n'Play just worked fine.

denisfa:
@simsemax Thanks for the report. I believe this is similar to another open issue. Could you provide what language are you using on your Mac? We have a translation issue currently reported with french.

simsemax:
Hi @denisfa,
thanks for the reply! My Mac OS runs in german and so does VBA. I have found the threads #110 and #126 after I posted this but I haven't figured out yet how to change the VBA language to english on my mac. I suppose the first time I ran VBA it was in English and everything worked.

denisfa:
[...]

Issue #273, opened 2018-07-02
Key binding not saved properly.


UynGH:
### Expected behavior

Saved key mapping with the french translation.

### Actual behavior

Nothing is saved when I close VBA-M.

### Steps to reproduce the behavior

Try using VBA-M with the french translation and configure any joypad. Every key that's translated will not be saved.

### ROM details

Any ROM.

### Option details

Nothing specific. Tried on every first launch everytime.

### Build details

Latest stable. 2.1.0 x64.

Operating System (Windows, Mac, Linux (state distribution), etc.):

Windows 10.

### Hardware details:

X360 gamepad.

---

First launch key mapping looks like this: 
![1](https://user-images.githubusercontent.com/40763842/42165732-3c8d4f7a-7e09-11e8-8567-c3298ed0340b.png)

Clicked OK then launched any ROM and nothing worked in game. Checked the vbam.ini file then that's what was saved: 
![3](https://user-images.githubusercontent.com/40763842/42165807-69ffc14a-7e09-11e8-8147-4f839589b3f9.png)

Pretty sure it's the translation's fault because when I removed the wxvbam.mo file from my folder, everything worked perfectly in english.

Workaround: I had to put every key manually in the vbam.ini file (written in english, not in french): 
![4](https://user-images.githubusercontent.com/40763842/42165880-a46bf538-7e09-11e8-8daf-1d91b065ecc3.png)
And everything worked.

I guess it's because we translated everything in french on Transifex (ie. button = bouton, Left=Joy1-Hat0W = Left=Joy1-Hat0O, etc.). I just updated the translation yesterday but I can't try it since I don't know how to compile the wxvbam.mo file needed.

Do we need to let every key in english and don't translate them?

Thanks in advance and great job guys. (:

PR #458, opened 2019-07-01
Translation issues.

### LINKED ISSUE #441: Emulator does not read rom
### Expected behavior
Open the game when you select the rom

### Actual behavior
It says that it is not possible to read the rom, even if it is in the format .gba, .gbc or .gb and neither the place where it is

Open the emulator and select a rom

### ROM details
All GBA, GBC and GB games in any format

### Build details
Windows 8

Version 2.1.3

Edit:
This occurs in the default language (I do not know if translating yet the error occurred because I could not translate)

It seems that the bug happens after you change version

In version 1.8.0 and 2.1.0, I selected a folder, now in 2.1.3 it does not read the folder any more, only if it changes the game or the folder of the place

Sem título

Sem título 1

denisfa:
@rkitover  I had to fix some spelling for the french translation and create a function to initialize the regex properly. If we try like was before, the regex compiles the english words (don't use gettext to translate it). This is why the configuration was not set.

I have to update the transifex too, but first want your feedback regarding this. I suggest we drop all translation for buttons/axes/joy and use the english text only. The labels will remain translated, but these may cause a minor inconvenience when switching languages, as the app will complain about weird strings (set as french, then use en_US ==> warning about wrong configuration).

Fix #441, #437 and #273. I tested for french, german and portuguese. Everything worked as expected.

After your feedback I will update transifex accordingly. Then we can merge this.

### DIFF
--- a/src/wx/widgets/joyedit.cpp
+++ b/src/wx/widgets/joyedit.cpp
@@ -101,52 +101,52 @@ wxString wxJoyKeyTextCtrl::ToString(int mod, int key, int joy)
     wxString s;
     // Note: wx translates unconditionally (2.8.12, 2.9.1)!
     // So any strings added below must also be translated unconditionally
-    s.Printf(_("Joy%d-"), joy);
+    s.Printf(("Joy%d-"), joy);
     wxString mk;
 
     switch (mod) {
     case WXJB_AXIS_PLUS:
-        mk.Printf(_("Axis%d+"), key);
+        mk.Printf(("Axis%d+"), key);
         break;
 
     case WXJB_AXIS_MINUS:
-        mk.Printf(_("Axis%d-"), key);
+        mk.Printf(("Axis%d-"), key);
         break;
 
     case WXJB_BUTTON:
-        mk.Printf(_("Button%d"), key);
+        mk.Printf(("Button%d"), key);
         break;
 
     case WXJB_HAT_N:
-        mk.Printf(_("Hat%dN"), key);
+        mk.Printf(("Hat%dN"), key);
         break;
 
     case WXJB_HAT_S:
-        mk.Printf(_("Hat%dS"), key);
+        mk.Printf(("Hat%dS"), key);
         break;
 
     case WXJB_HAT_W:
-        mk.Printf(_("Hat%dW"), key);
+        mk.Printf(("Hat%dW"), key);
         break;
 
     case WXJB_HAT_E:
-        mk.Printf(_("Hat%dE"), key);
+        mk.Printf(("Hat%dE"), key);
         break;
 
     case WXJB_HAT_NW:
-        mk.Printf(_("Hat%dNW"), key);
+        mk.Printf(("Hat%dNW"), key);
         break;
 
     case WXJB_HAT_NE:
-        mk.Printf(_("Hat%dNE"), key);
+        mk.Printf(("Hat%dNE"), key);
         break;
 
     case WXJB_HAT_SW:
-        mk.Printf(_("Hat%dSW"), key);
+        mk.Printf(("Hat%dSW"), key);
         break;
 
     case WXJB_HAT_SE:
-        mk.Printf(_("Hat%dSE"), key);
+        mk.Printf(("Hat%dSE"), key);
         break;
     }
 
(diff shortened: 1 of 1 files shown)


rkitover:
I don't think we should depend in any way on a translation being correct.

The question is, where is the correct translated string coming from if there is a mismatch, the code should simply use whatever the translation is.

denisfa:
> I don't think we should depend in any way on a translation being correct.
> The question is, where is the correct translated string coming from if there is a mismatch, the code should simply use whatever the translation is.

The current translation is split on components. For example, with `fr_FR.utf8`, `Joy` goes to `Manette` but `Joy[0-9]+` stays with `Joy[0-9]+` instead of `Manette[0-9]+`. This is what I am calling spelling error. There are some minor ones when abbreviating directions. Check what was done on the `po` files.

rkitover:
I understand, in this case we should simply have only one string being translated and that is what the regexes should be compiled with.

There should be no changes to the `.po` files.

rkitover:
BTW I think I fixed the translation issues for the keyboard side in 30b92724.

rkitover:
What is mean is, instead of translating both the strings `Joy` and `Joy[0-9]+` just translate the string `Joy` and concatenate it with whatever is necessary.

In this way the code will not depend on any parts of the translation.

denisfa:
> BTW I think I fixed the translation issues for the keyboard side in 30b9272.

What was the issue? For translated keys?

> What is mean is, instead of translating both the strings Joy and Joy[0-9]+ just translate the string Joy and concatenate it with whatever is necessary.

Do you think it is necessary to translate the buttons and joystick binds? If we work with english only, we would not need to update/create translations for these. The labels and descriptions I would keep the way it is now. There is a possible issue that if you set these with a language, and then open vbam using another one, the user would lose all its input because they would not be recognized (with a warning message).

rkitover:
Yes I think it doesn't hurt, see how in `keyedit.cpp` we use both e.g. `_("CTRL")` and `wxT("CTRL")`, we can do the same here.

In the regex, compile it as e.g.:

```
"(Joy|" + _("Joy") + ")"
```

adjust the match group number. Since `wxRegex` is POSIX regex it doesn't have non-capturing parens.

But the key issue is that only one translated string be used, we do not want to depend on the translation files in any way.

denisfa:
> But the key issue is that only one translated string be used, we do not want to depend on the translation files in any way.

What do you mean? We don't depend on translation for it to work, but the strings that are not translated yet will keep the original text (english). This does not affect us. Unless you mean to show the translated text, if possible, and use all english under the hood. When the config is read, it will be translated to whatever language is set; likewise, when the config is saved, it will be with english text.

rkitover:
Like I said, you are translating the whole regex when you should just translate the string "Joy". See above.

denisfa:
> Like I said, you are translating the whole regex when you should just translate the string "Joy". See above.

Sorry, my head is all over the places now. I understand what you are doing here, but I did not give one information of mine. If we check the po files, we actually do not have the strings `Axis`, `Joy` and friends. For instance,

```
$ fgrep -RnwI ./po -e "Axis" -A 1
...
./po/wxvbam/tr.po:1131:msgid "Axis%d+"
./po/wxvbam/tr.po-1132-msgstr "Eksen%d+"
--
./po/wxvbam/tr.po:1136:msgid "Axis%d-"
./po/wxvbam/tr.po-1137-msgstr "Eksen%d-"
--
./po/wxvbam/tr.po:1189:msgid "Axis([0-9]+)([+-])"
./po/wxvbam/tr.po-1190-msgstr "Eksen([0-9]+)([+-])"
...
```
```
$ fgrep -RnwI ./po -e "Joy" -A 1
...
./po/wxvbam/pt_BR.po:1130:msgid "Joy%d-"
./po/wxvbam/pt_BR.po-1131-msgstr "Controle%d-"
--
./po/wxvbam/pt_BR.po:1189:msgid "^Joy([0-9]+)[-+]"
./po/wxvbam/pt_BR.po-1190-msgstr "^Controle([0-9]+)[-+]"
...
``` 

We only have the regex translation. This is why I did not want to do this split, and offered the idea of english only for these. If we are going to do something, we have to decide what we should do with the translation.

As far as I can tell, we can either:
1. stick with the current solution minus the po editing (that should be done on transifex);
2. only show the translated strings, but use english under the hood;
3. remove the regex from all po files (transifex), and replace them with the single keywords, then use your idea about setting up the regex to be english + current locale;
4. double the regex on `CompileRegex`, the same way you are suggesting to do for 3, eg
 `"(Joy%d-|" + _("Joy%d-") + ")"` etc.

I believe all of these are viable, given enough time to do it.  `1`, `3` and `4` have this hypothetical problem that if the user saves it Input on one language, then tries to run with other, he will be given a warning and its Input will be emptied. Very like what is happening with the stable version + warnings.

Hope I made more clear now why this is taking so long to close.

rkitover:
I'm fine with just using english for this and not using the translated strings anymore.

denisfa:
@rkitover That should be it. Every time I look at this I feel a bit of wasted potential, but it is true that this should allow us to focus on other stuff.

rkitover:
We cannot remove the translation from keyboard keys, wx translates them unconditionally.

rkitover:
Rebase your work into one commit that just removes the translation from joystick keys and don't do anything else.

rkitover:
Also I fixed all the translation issues with the keyboard keys anyway.

rkitover:
If you do want to translate the joystick strings, just use the actual word concatenated with other characters, once the string is pushed to transifex someone will translate it again.

denisfa:
> We cannot remove the translation from keyboard keys, wx translates them unconditionally.

We can use `ToRawString()` to get the non-translated one. It is on the trick that you did to change the key into an accelerator to get the string.

> Rebase your work into one commit that just removes the translation from joystick keys and don't do anything else.

No problem, before I finish this I will check the translations on transifex. It is for me to remember.

> If you do want to translate the joystick strings, just use the actual word concatenated with other characters, once the string is pushed to transifex someone will translate it again.

The thing is here is that I do believe there are some good ideas about how to tackle this and I would not like to waste anything that we already have. But it is not up to me to decide this, and your solution is good enough for me. I will go with removing the translation of joystick keys. By next push here it will be done.

> Also I fixed all the translation issues with the keyboard keys anyway.

What issue? The keyboard being translated at the moment is also saved translated on `vbam.ini`. It works for that locale only. If it is not translated, then english text is used, so no actual issues minus partial translation.

rkitover:
> What issue? The keyboard being translated at the moment is also saved translated on vbam.ini. It works
> for that locale only. If it is not translated, then english text is used, so no actual issues minus partial
> translation.

There were issues with this years ago for users of translations, I don't remember anymore, but I'm pretty sure I fixed them with 30b92724 and d585a610.

Once we merge this, we can do a new release, I will post instructions for you on the CHAT page for how to do the release.

denisfa:
@rkitover Done. I still believe we are going to go back to this, since someone will complain that the joystick is not translated like the keys, but for now it should work.

Issue #65, opened 2017-01-16
Emulator crashes when trying to load a save state


autofire372:
### Expected behavior
The previously saved state is loaded.
### Actual behavior
The emulator crashes to desktop.
### Steps to reproduce the behavior
Open a game. Make a save state, and then attempt to load it (i.e. hit Shift+F1 and then F1). The program should then crash.
### Build details

Operating System (Windows, Mac, Linux, etc.):
Windows 10 x64
Interface (wx, SDL, default is wx):
wx
Version of code (pre-built binary version, or commit ref, or just "master"):
master
Any other relevant build information:

If this is e.g. a joystick or video issue, then the relevant hardware information:

ichigo-0:
In my case, it is inconsistent. Sometimes it crashes when I try to load a saved state, but if I relaunch and try again, it will often load that state successfully. I don't think I've yet found a save that couldn't be loaded (but initially I assumed the save states were corrupt and after a crash would try to load a different state).

I'm also using master (fcc34394c139), though I'm on Linux.

Possibly related are crashes when saving state. These crashes seems to happen after the save completes (because I can load it successfully later).

TheDauthi:
I'm seeing the same behavior - inconsistent crashes to desktop on save file loading. (I also see the same problem with saves).

Upon regenerating the ini file, the first time I loaded a save state, the emulator crashed. Upon reopening, it no longer crashes.

rkitover:
Apparently there's an issue with the current save code in that it incorrectly determines the flash size for some ROMs, this is known to be a bug with normal saves, but it may be affecting state saves as well.

Squall-Leonhart:
it would not be affecting save states, the issue at large is because vba-m writes to the config file when a state is loaded.

kvaak:
64-bit Win10, latest GitHub release. I'm not sure whether it's related to this issue at all, but since it involves broken save states I'm going to assume so. My save states load just fine, problem is they seem to consist of glitchy garbage that makes Sinclair Spectrum loading screens sound like an opera. ROM is European version of FF6 Advance with a sound restoration patch glued on top.

**SOUND WARNING. SERIOUSLY, TAKE OFF YOUR HEADPHONES.**

https://www.youtube.com/watch?v=Rpiric_J16U

[...]

PR #459, opened 2019-07-01
[WIP] Fix load state failing to load with Simple output module.

denisfa:
@rkitover This is a very weird behaviour for this output module. When I checked #65, I got several types of errors (like first `double free`, then `segfault`, then `free(): corrupted unsorted chunks`). With the help of @retro-wertz, we used ASAN to find a specific line on our code, just before the error's party begun.

I will point out the code, hoping you can have a better understanding about this.

### DIFF
--- a/src/wx/panel.cpp
+++ b/src/wx/panel.cpp
@@ -31,6 +31,7 @@ GameArea::GameArea()
     , rewind_time(0)
     , do_rewind(false)
     , rewind_mem(0)
+    , num_rewind_states(0)
     , loaded(IMAGE_UNKNOWN)
     , basic_width(GBAWidth)
     , basic_height(GBAHeight)
@@ -1860,6 +1861,7 @@ void DrawingPanelBase::DrawArea(uint8_t** data)
                     showSpeedTransparent);
 
                 free(buf);
+                buf = NULL;
             } else
                 panel->osdtext.clear();
         }
@@ -1978,9 +1980,11 @@ void DrawingPanelBase::OnSize(wxSizeEvent& ev)
 DrawingPanelBase::~DrawingPanelBase()
 {
     // pixbuf1 freed by emulator
-    if (pixbuf2)
+    if (pixbuf1 != pixbuf2 && pixbuf2)
+    {
         free(pixbuf2);
-
+        pixbuf2 = NULL;
+    }
     InterframeCleanup();
 
     if (nthreads) {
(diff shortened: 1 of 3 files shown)


src/wx/panel.cpp
```diff
@@ -2004,7 +2008,7 @@ BasicDrawingPanel::BasicDrawingPanel(wxWindow* parent, int _width, int _height)
     // 16 or 32, though
     if (gopts.filter == FF_NONE && gopts.ifb == IFB_NONE)
         // changing from 32 to 24 does not require regenerating color tables
-        systemColorDepth = 24;
+        systemColorDepth = 32;
```
denisfa: @rkitover @retro-wertz When you use `24`, the code for the Simple output seems to not allocated enough memory for everything. Then, we wait for whenever there is a wrong memory access when loading a state (can happen after {1,3,11,...} try(ies)). Depending where it lands, we get a different error message.

denisfa:
On the code `./src/wx/panel.cpp`, we have this scope:
```
void DrawingPanelBase::DrawArea(uint8_t** data)
{
    // double-buffer buffer:
    //   if filtering, this is filter output, retained for redraws
    //   if not filtering, we still retain current image for redraws

     ### IF WE HAVE systemColorDepth == {16,32}, IT WORKS FLAWLESS
    int outbpp = out_16 ? 2 : systemColorDepth == 24 ? 3 : 4;
    int outrb = systemColorDepth == 24 ? 0 : 4;
    int outstride = std::ceil(width * outbpp * scale) + outrb;

    if (!pixbuf2) {
        int allocstride = outstride, alloch = height;

        // gb may write borders, so allocate enough for them
        if (width == GameArea::GBWidth && height == GameArea::GBHeight) {
            allocstride = std::ceil(GameArea::SGBWidth * outbpp * scale) + outrb;
            alloch = GameArea::SGBHeight;
        }

        ### ERROR LINE BELOW
        pixbuf2 = (uint8_t*)calloc(allocstride, std::ceil((alloch + 2) * scale));
    }
```

Following this, we have `
void BasicDrawingPanel::DrawArea(wxWindowDC& dc)` also with a special case for when `systemColorDepth == 24` dealt as the first `if` there. I believe there should be some type of code change here doing something similar to what happens to other cases, eg, some type of scale(=1).

As a result, if you use any filter with Simple, it works as expected.

Issue #416, opened 2019-04-18
Clear button on Input Configuration works weird


RunisDauphin:
I expect it should clean only the button configuration you are in.

### Actual behavior

It cleans all the buttons, and there is no button to eliminate only one button.
The only way I can do it is by manually change them in vbam.ini and search for the wanted command.

rkitover:
you can press backspace to delete a button

rkitover:
I could add a button next to each box to clear it, and rename the bottom clear button "reset".

RunisDauphin:
That would be great. By now, this is the only emulator I found that does this kind of things.

Thanks for your attention, M. Kitover.

Issue #362, opened 2019-02-02
Default Game Keys


rkitover:
I want to get some feedback from users about your preferred game keys (and any other keys you use) to see if changing the defaults makes sense.

Here is what I propose:

| keys | GBA keys |
|------|-----------|
| WASD | direction
| `[]` | a / b |
| P\ | l / r |
| enter | start |
| backspace | select |

other default keys like speedup would be somewhere in the `[]` area as well.

Also want to change the default joystick binds to match the 360/xbone controller.

Please let me know what you think.

Rambin:
It depends on what game I'm playing.
For example, I use WSAD for direction and [K][J][L][;] for ABLR when playing Metroid, and [J][K][U][I] for ABLR when playing Gyakuten Saiban.
Is it possible to save key setting seperately for a game?

rkitover:
That sounds like a nice feature to add, or perhaps configuration profiles of some sort.

RunisDauphin:
As far as I know, previous versions of VBA-M, the original VBA and VBA-RR allow to configure up to 4 different sets of keys, leaving you the chance of changing between games or even setting a local multiplayer.

I have never tested it with a multiplayer, but I did some times set two different sets: one with my keyboard and the other one with my beloved Black European Gamecube controller (because we gotta remember that VBA and some of the forks of that time did not allow to assign more than one command to any button), so I could change it whenever I wanted it. Including this feature for this characteristic (and local multiplayer) would be pretty pointless, since you can assign several commands to a single button in the newer versions of VBA-M, but for using different sets for different games, sounds great.

Thanks for hearing me, and much welcomed for providing some help and ideas. If someone wants to tell Endrift about this, feel free to do.

P.S.: If someone wants a configuration for a Gamecube controller through a vJoy device, this is what I recommend (always preceded by "Joy1-", where Joynumber is your number of port, being 1 the one at your right):
A: Button0 (A)
B: Button1 (B)
L: Button6 (L)
R: Button5 (R)
START: Button7 (START)
SELECT: Button4 (Z)
Up: Button8 (D-Up) and/or Axis1- (A-Up)
Down: Button9 (D-Down) and/or Axis1+ (A-Down)
Left: Button10 (D-Left) and/or Axis0- (A-Left)
Right: Button11 (D-Right) and/or Axis0+ (A-Right)

denisfa:
We could have a profile per game, that could be a `ini` file with the name of the game.
[...]

PR #467, opened 2019-07-11
General enhancements.

denisfa:
@rkitover Here I have some enhancements that I have been working on. They finish some issues, as per commit messages (the 3 currently open for project `v2.1.7`), but are mostly some code that I wanted to study.

Feel free to comment about whatever you want of the code.


src/wx/guiinit.cpp
```diff
@@ -3330,10 +3330,36 @@ bool MainFrame::BindControls()
...
         {
+#ifndef NO_ONLINEUPDATES
             // Online Auto Update check frequency
             getrbi("UpdateNever", gopts.onlineupdates, 0);
             getrbi("UpdateDaily", gopts.onlineupdates, 1);
             getrbi("UpdateWeekly", gopts.onlineupdates, 7);
+#else
+            wxWindowList &children = d->GetChildren();
+            std::vector<wxWindow*> forDeletion;
+            for (wxWindowList::Node *node = children.GetFirst(); node; node = node->GetNext())
```
rkitover: This is cool, but is this really necessary? Not displaying the controls would probably be enough. Fine with me either way though.

rkitover:
Other than that, everything looks awesome!

src/wx/guiinit.cpp
denisfa: Is there a way to hide all elements from a `<object class="sizeritem">` or `<object class="wxBoxSizer">`? I did not find a easy way to do this, like I am doing at the moment (by selecting something with label/name equals to what I want to remove).

Since I was already doing the entire selection, I decided to delete these controls.

denisfa:
> Honestly I'd much prefer WASD for direction, and ENTER for START and BACKSPACE for SELECT. Unless you have strong objections.
> 
> Other keys can be:
> 
> A: DELETE, `:`
> B: END, `"`
> L: INSERT, P
> R: HOME, `[`
> 
> This way we support both full and compact layouts comfortably.
> 
> What do you think?

What keyboard layout do you have in mind here? I agree with changing; this feels better for notebooks indeed. Another thing, about the XBOX 360 controller: the D-PAD or the analog for the keys? At the moment, I set the left analog. But thinking again, It feels like the D-PAD would be more intuitive.

About the current choice, It was very subjective; I use the PT-BR ABNT2 layout.

retro-wertz:
just incase my suggestion still matters... https://github.com/visualboyadvance-m/visualboyadvance-m/issues/362#issuecomment-510693715

denisfa:
@rkitover Here it is. There are some extras, like #416. It was almost automatically since I had to check the XRC file for input configuration. I will just squash the WIP's and it is good to go.

rkitover:
All of this looks good, and I like your choices for the keybinds. Are you ready for merging?

rkitover:
Also, if it's possible it would be good to use both the joystick hat and the left analog stick, some people prefer the stick, but the hat should definitely be bound by default.

denisfa:
> Also, if it's possible it would be good to use both the joystick hat and the left analog stick, some people prefer the stick, but the hat should definitely be bound by default.

@rkitover Ok, I will do that. I was talking about those 2 WIP between commits. I forgot about them. I will signal here when it is ready to merge.

rkitover:
@denisfa is this ready for merging? Looks good to me.

denisfa:
@rkitover Sorry, I was working on the vid recording and completely forgot about this. Yeah, it is ready.

There is one thing that I wrote down: the shortcut for opening recently played games (`CTRL + F{1,10}`) is not working on Linux; it is not reaching the app. I was going to suggest to replace these with `CTRL + {1,10}`, but wanted to further investigate if this is a Linux thing only or something else; but this can be done in a separate merge request.

rkitover:
I will look into that, I've done some stuff with the accelerators.

I do not recommend CTRL+1 because we wanted to reserve this for features like geometry shortcuts.

rkitover:
Also I really appreciate the nice git history, nice work!

rkitover:
I have a gentoo + KDE setup and the CTRL + F1 accelerator works fine for me.

Going to test on windows.

What are you using?

rkitover:
Works fine on windows too.

denisfa:
> I have a gentoo + KDE setup

Ah, I see you are a man of culture as well.

I use XFCE over here, but have quite some stuff from Xorg customized. It is probably one of them. 

Just a final question, if you don't mind: when you load a new game, does the accelerator for the new game works too, as in without needing to close the app first?

rkitover:
@denisfa I checked that the accelerators (ctrl+f1, ctrl+f2) work with another game loaded at the time, either via command line or via menu, on both linux and windows.

denisfa:
@rkitover Thank you! I will figure it out my setup now.

rkitover:
@denisfa The `xev` program is good for debugging these things, it shows exactly what X events are generated when you press a specific key.

PR #474, opened 2019-07-25
Just a minor issue with clang and snprintf on windows

ZachBacon:
this fixes the build process but shouldn't interfere with gcc


rkitover:
Just remove that `#define`, it's wrong and not being used anywhere.

ZachBacon:
Pushed

Issue #476, opened 2019-07-26
[Debugger] Logging showing erroneous values


retro-wertz:
### Expected behavior

aside from it has memory leaks, it should work as expected. and this used to log to a text file but not anymore .

### Actual behavior

returns erroneous values, no logging.

### Steps to reproduce the behavior

load gba game. goto tools > Logging. check on anything that would show an output. compare to source code to compare formatting..

![Screenshot_2019-07-26_23-27-24](https://user-images.githubusercontent.com/23315382/61963336-ee3be480-affd-11e9-8886-749e8dac0784.png)

you can also use the the log() function and output something like log("%02x %04x %08x\n", 0xff, 0xffff, 0xffffffff); this should show the formatting and values as specified.

![Screenshot-2](https://user-images.githubusercontent.com/23315382/61962983-5d650900-affd-11e9-81cc-96dd696f7048.png)

here is what is expected to be shown:
![Screenshot-3](https://user-images.githubusercontent.com/23315382/61963099-884f5d00-affd-11e9-9b20-88c967a873f5.png)

an error box also appears at runtime:
![Screenshot_2019-07-25_23-37-50](https://user-images.githubusercontent.com/23315382/61962889-2bec3d80-affd-11e9-9f56-90aadbdee8a1.png)

### Crash bugs

If this is a crash bug, and you have the time and inclination, please provide a
symblic backtrace.

Instructions for doing so are
[here](https://github.com/visualboyadvance-m/visualboyadvance-m/blob/master/README.md#reporting-crash-bugs).

(see bottom screenshots)

### ROM details

Which games are affected by the issue (please state if GBA or GB game): all

Which games, if any, are NOT affected by the issue (please state if GBA or GB game): non

Include screenshots of ROM Information (File > ROM Information): no rom specific

If you are using a BIOS file, please say which one.

### Option details

Video driver (simple, opengl): 

Sound driver (directsound, xaudio, openal, SDL):

### Build details

Operating System (Windows, Mac, Linux (state distribution), etc.): arch

If using Linux, specify if using xorg or Wayland: xorg

Version of code (pre-built binary version, or commit ref, or just "master"):

### Hardware details:

If your joystick or gamepad isn't working in Linux, please try to troubleshoot
using the information on this page:
https://wiki.archlinux.org/index.php/Gamepad **BEFORE** filing an issue. We use
SDL for joystick support, any problems will generally be an issue with SDL and
**NOT** the emulator.

[...]

PR #477, opened 2019-07-26
Fix wxWidgets assertion error on `Logging...` menu.

denisfa:
@rkitover This is a regression needed for #476 (it will not close it, but partially fix it).

I also did the test of `log("%02x %04x %08x\n", 0xff, 0xffff, 0xffffffff);` on a random scope to check its return and it seems to be working.

The memory leak is a bit more complicated... There are some things mixed there and It is probably needed to do something more towards refactoring. It will be on the list of memory issues, nevertheless.

### DIFF
--- a/src/wx/sys.cpp
+++ b/src/wx/sys.cpp
@@ -1294,8 +1294,10 @@ bool debugWaitSocket()
 void log(const char* defaultMsg, ...)
 {
     va_list valist;
+    char buf[2048];
     va_start(valist, defaultMsg);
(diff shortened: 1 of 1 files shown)


rkitover:
Actually, take a look at what I did with `wxLogDebug()` in `wxvbam.h`, maybe we can combine these somehow.

denisfa:
> Actually, take a look at what I did with wxLogDebug() in wxvbam.h, maybe we can combine these somehow.

@rkitover Sorry about missing this comment. I will check and see what can be done.

Issue #124, opened 2017-06-18
Don't work rec video: Error in video recording


slacknk:
OS-Distro: slackware64-14.2 
Version visualboyadvance-m (wx): 20170603_throttle (from git origin/master)

Build keys: 
```
  SLKCFLAGS="-O2 -fPIC"
  LIBDIRSUFFIX="64"
  ASM=OFF
```
```
 -DCMAKE_C_FLAGS:STRING="$SLKCFLAGS" \
 -DCMAKE_CXX_FLAGS:STRING="$SLKCFLAGS" \
 \
 -DENABLE_ASM_CORE=${ASM} \
 -DENABLE_ASM_SCALERS=${ASM} \
 \
 -DENABLE_WX=ON \
 -DwxWidgets_CONFIG_EXECUTABLE=/usr/lib$LIBDIRSUFFIX/wx/config/gtk2-unicode-3.0 \
 -DwxWidgets_USE_DEBUG=OFF \
 \
 -DENABLE_FFMPEG=ON \
 -DENABLE_SDL=ON \
 -DENABLE_NLS=ON \
 -DENABLE_CAIRO=ON \
 -DENABLE_OPENAL=ON
```

Open anybody ROM, Menu: File - Record - Start video recording...
folder, example: /tmp - filename without spaces - x264/mkv/ogv/avi - save 
and opened window with error:

> Error in video recording (error writing to output file); aborting

Log from terminal:
```
$ visualboyadvance-m 
[libmp3lame @ 0x3cfd400] Channel layout not specified
[avi @ 0x3d00e00] Using AVStream.codec.time_base as a timebase hint to the muxer is deprecated. Set AVStream.time_base instead.
[avi @ 0x3d00e00] Using AVStream.codec to pass codec parameters to muxers is deprecated, use AVStream.codecpar instead.
[avi @ 0x3d00e00] Using AVStream.codec.time_base as a timebase hint to the muxer is deprecated. Set AVStream.time_base instead.
[avi @ 0x3d00e00] Using AVStream.codec to pass codec parameters to muxers is deprecated, use AVStream.codecpar instead.
[swscaler @ 0x3d0a400] Warning: data is not aligned! This can lead to a speedloss
[avi @ 0x3d00e00] Timestamps are unset in a packet for stream 0. This is deprecated and will stop working in the future. Fix your code to set the timestamps properly
[avi @ 0x3d00e00] Encoder did not produce proper pts, making some up.
[...]

Issue #404, opened 2019-04-04
throttle/turbo/speedup related bugs/glitches/issues


rkitover:
Related threads:

#374 #339 #212 

Please discuss throttle, speedup button (called turbo button sometimes) and the `Speefdup / Turbo` config dialog related issues here.

## some background

There are three possible states when the turbo button is pressed, throttle, speedup and speedup with frame skip.

If you select zero throttle and zero frame skip, then speedup is in effect, the game runs at full speed without skipping frames.

If you select a number of frames to skip, then speedup is in effect, the game runs at full speed and ALSO skips that number of frames, the default is/was 9 frames.

If you select throttle, speedup is NOT in effect, and the sound driver syncs the game to that percentage of game speed, the game speed is always synced by the sound driver regardless.

rkitover:
@LazyBunny

The "multithread filter" option never actually did anything, except maybe in the MFC builds. I have removed it in master.

Now that I see that things are mostly working as expected, minus glitches, we need to start isolating the glitches.

So what would help me is any particular glitch you can find and give me the process to reproduce exactly, and anyone else of course.

LazyBunny:
Pretty much all my results for (475%?)500% turbo carried over into 2.1.12 from that 2.1.11 that i tested.
I also tested a couple of the weird things on my xeon x5670@4ghz 6C/12T desktop, to rule out my GPD-Win2's performance limitations.
I know, that was a LOT of test results to pick through. If you want me to test something, i can. But i'll only post the combinations that're broken.

rkitover:
@LazyBunny thank you, what I really would like is for things that are broken, the steps to reliably reproduce the brokenness, because only then would I be able to fix them. There are lots of glitches in this codebase that's for sure.

PR #481, opened 2019-08-01
[WIP] Fix video/audio recording.

denisfa:
@rkitover @ZachBacon 

```
We create a namespace to deal with most of our recording solution.

Besides that, we also add some functions to remove the need of
including libavutil headers on other part of the code. This is meant to
isolate most of recording solution components on the proper files.

We will start with a limited number of codecs supported; slowly we
should add them as they are tested (the previous one did not work for
most codecs listed).

This should support `ffmpeg 4.1` and further, including removing
all compilation warnings related to versions discrepancy.
```

This is my promised solution for #124, and it includes a heavy refactoring from the former one. I tried to put as much work as possible on this (not because it is hard, but I needed a continuous amount of time to do it; which I did not have thanks to my work schedule from the recent past). I did some basic testing with my games and it worked; more testing is always good, though.

Feel free to comment and ask for any type of adjustment relevant here. I believe I learned a lot about `FFmpeg`, so I should be able to answer some questions.

Also, I have a followup commit about adding new formats for the output (both audio/video), but I need you guys approve the structure overall before doing so. It might take some time to review this, but it will be worth it!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -280,7 +280,7 @@ set(
 if(ENABLE_FFMPEG)
     find_package(PkgConfig REQUIRED)
 
-    pkg_check_modules(FFMPEG REQUIRED libavcodec libavformat libswscale libavutil)
+    pkg_check_modules(FFMPEG REQUIRED libavcodec libavformat libswscale libavutil libswresample)
 
     if(FFMPEG_STATIC)
         set(FFMPEG_LIBRARIES ${FFMPEG_STATIC_LIBRARIES})
@@ -625,7 +625,7 @@ set(
 )
 
 if(MSVC)
-	set(SRC_MAIN ${SRC_MAIN} "dependencies/msvc/getopt.c")
+        set(SRC_MAIN ${SRC_MAIN} "dependencies/msvc/getopt.c")
 endif()
 
 set(
@@ -643,7 +643,7 @@ set(
 )
 
 if(MSVC)
-	set(HDR_MAIN ${HDR_MAIN} "dependencies/msvc/getopt.h")
+        set(HDR_MAIN ${HDR_MAIN} "dependencies/msvc/getopt.h")
 endif()
 
 if(ENABLE_FFMPEG)
--- a/src/common/ffmpeg.cpp
+++ b/src/common/ffmpeg.cpp
@@ -106,15 +106,21 @@ recording::MediaRet recording::MediaRecorder::setup_audio_stream()
     aenc->sample_fmt = acodec->sample_fmts ? acodec->sample_fmts[0] : AV_SAMPLE_FMT_FLTP;
     aenc->bit_rate = 128000; // mp3
     aenc->sample_rate = sampleRate;
+    // this might be useful to check if the codec suports the
+    // sample rate, but it is not strictly needed for now
+    bool isSupported = false;
     if (acodec->supported_samplerates)
     {
-        aenc->sample_rate = acodec->supported_samplerates[0];
         for (int i = 0; acodec->supported_samplerates[i]; ++i)
         {
-            if (acodec->supported_samplerates[i] == 44100)
-                aenc->sample_rate = 44100;
+            if (acodec->supported_samplerates[i] == sampleRate)
+            {
+                isSupported = true;
+                break;
+            }
         }
     }
+    if (!isSupported) return MRET_ERR_NOCODEC;
     aenc->channels = av_get_channel_layout_nb_channels(aenc->channel_layout);
     aenc->channel_layout = AV_CH_LAYOUT_STEREO;
     if (acodec->channel_layouts)
--- a/src/common/ffmpeg.cpp
+++ b/src/common/ffmpeg.cpp
@@ -450,6 +450,7 @@ void recording::MediaRecorder::Stop()
     npts = 0;
     if (oc)
     {
+        flush_frames();
         // close the output file
         if (!(fmt->flags & AVFMT_NOFILE))
         {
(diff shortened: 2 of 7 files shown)


rkitover:
I rebased on top of master and fixed conflicts.

rkitover:
In case you didn't know, to update a force pushed branch do:

```bash
git fetch --all --prune
git reset --hard origin/fix-recording
```

denisfa:
> I rebased on top of master and fixed conflicts.

Thank you very much!

> In case you didn't know, to update a force pushed branch do:

Thanks for the tip! I believe what happened here was mostly because I had to work on this across multiple days and it was not a linear work. I wanted to do several commits, but I started mixing a lot of stuff and hence ended up with a single commit. If you check the log, you will see a bunch of WIP commit's message with some PT-BR comments in the code.

> I honestly didn't look at the ffmpeg specific details

There is something I wanted to ask: what happens when you turbo/speed up the game? How this is done on the core? If you try to record audio with the current solution (the former one too), we get the audio recording as withing game time, not real time. The video recording works in this case. Try to run a test and video record a game with turbo or speed up. You will understand what I mean.

I believe it was something related to the sample rate, but I failed to confirm it.

rkitover:
Yeah, when you work in a branch on a long-term project, you want to sync with master occasionally, using something like:

```bash
git fetch --all --prune
git rebase origin/master
```

rkitover:
Yeah I worked on the sound drivers before, including rewriting the SDL one. The SDL one is our simplest and most stable sound driver, see `src/common/SoundSDL.cpp`.

Changing sound rate is in fact how turbo works, the game is always synced to the sound driver and the sound driver controls the speed of the game.

I will take a look as well and see if it's something simple.

denisfa:
I see this line on `src/common/SoundSDL.cpp`:
```
audio.freq     = current_rate ? sampleRate * (current_rate / 100.0) : sampleRate;
```

I need this info for recording... But this does not seem common for all audio drivers. What can we do here? Also, how does the sample rate works for the emulation in general? What is supposed to happen when you change from 44100Hz to 22000Hz or something else? I see no diff for the audio output, at least in the games I am trying. This might be important for the recording, I think.

denisfa:
> I need this info for recording... But this does not seem common for all audio drivers. What can we do here? Also, how does the sample rate works for the emulation in general? What is supposed to happen when you change from 44100Hz to 22000Hz or something else? I see no diff for the audio output, at least in the games I am trying. This might be important for the recording, I think.

I did my own research and figured this out. I just needed some time to test more stuff.

I pushed an idea about how to tackle the turbo for recording. It revolves around not allowing to skip frames and saving everything as in-game time, that will be converted to real time. The result is a video whose length is bigger than what you would expect, but with all the frames instead.

I believe I know the diff now: we skip video frames and don't wait to read all samples from audio frames before playing it with turbo. The actual speedup is made by the core, but it is passed to the sound driver. Another solution is probably making the `ffmpeg` code less independent from the rest, to make use of global variables and such to guide the recording. This is weird because we skip video frames, but not audio ones; instead we play a small part of each of them (related to frame skip variables).

This was also an issue from the previous sound recording, so maybe it is much more complicated than it looks.

If you can, test and tell me what you think.

rkitover:
It seems I didn't implement the `speedup_frame_skip` stuff fully, but we should definitely fix it in a separate PR or commit.

rkitover:
We need to be careful here because the legacy 9 frame skip speedup code is used by libretro, while the new code is used by our GUI.

retro-wertz:
go on ahead and fix whatever is needed. frameskipping is not used in libretro-port (or at least not yet). it would be more beneficial if the actual core would only do audio frames and video frames when needed (like an actual skip is needed) and let frontend deal instead with whatever frameskip loops they want to do to with the vid and audio frames

denisfa:
We can protect it just like we do for [GBA](https://github.com/visualboyadvance-m/visualboyadvance-m/blob/master/src/gba/GBA.cpp#L3758) counterpart(?). This way we can keep this ready for libretro, if it ever needs it.

I don't think the current idea is the best solution, but I am struggling to find another route for this. I need to think outside the box.

If I understood this right, we set a constant frame skip value and ignore what was set by the user, at least for this GB scope under these conditions.

As far as what I am aiming for, I am using #404 as a reference:

 - [X] If you select zero throttle and zero frame skip, then speedup is in effect, the game runs at full speed without skipping frames.

Video is longer than expected, but it works.

- [ ] If you select a number of frames to skip, then speedup is in effect, the game runs at full speed and ALSO skips that number of frames, the default is/was 9 frames.

The result for default value is 6 (video) FPS for recording; but we still have 60 audio FPS, hence a desync here. Not sure what can I do.

- [X] If you select throttle, speedup is NOT in effect, and the sound driver syncs the game to that percentage of game speed, the game speed is always synced by the sound driver regardless.

Video is longer than expected, but it works.

> It seems I didn't implement the speedup_frame_skip stuff fully, but we should definitely fix it in a separate PR or commit.

If you can, open an issue with the details of what is missing. This way we could track this better.

rkitover:
> If you can, open an issue with the details of what is missing. This way we could track this better.

I don't know to be honest, it might be implemented for GBA not GB, you must have had a reason for commenting out those lines.

Anyway, just remove that part of the code and let's merge this and deal with turbo mode recording later. This is likely not going to be trivial to fix.

denisfa:
@rkitover because it was overriding the value of frame skips. I don't know if we expect this somewhere, so I did not want to remove it. If we can delete those lines, then we can do something a little better. I will push another idea.

denisfa:
> Anyway, just remove that part of the code and let's merge this and deal with turbo mode recording later. This is likely not going to be trivial to fix.

Which lines? The commented ones?

> If we can delete those lines, then we can do something a little better. I will push another idea.

If you meant the commented line, then forget about this. I would prefer to just remove them too.

rkitover:
Well it doesn't really matter anyway, I'll just merge this now.

PR #486, opened 2019-08-10
Fix behaviour of `Tools` > `Memory Viewer`.

denisfa:
@rkitover This is like one of those dumb segfaults that I mentioned earlier. As soon as you click this item, it appears a wxWidgets error message. It has been like this since my pr to remove most of the compilation warnings... 

@retro-wertz You are interested on this, at least at the time of the issue reporting. Check if this is what you want and test it, please. I know I can see rows of memory maps, but I don't know if they are correct or not. The GUI behaviour I could test myself, but we need to be sure that I am not going to break something else with this solution.

Feel free to comment whatever you want. If there is some type of improvement desired, leave a note. I will be off from now until next wednesday (14.8.19) to focus on personal stuff. When I am back, I will see what I can do for this.

### DIFF
--- a/src/wx/viewsupt.cpp
+++ b/src/wx/viewsupt.cpp
@@ -136,7 +136,7 @@ void DisList::MoveSB()
         pos = (topaddr - 100) / 10 + 100;
     else if (topaddr >= maxaddr - 1100)
         pos = (topaddr - maxaddr + 1100) / 10 + 300;
-    else
+    else // FIXME this pos is very likely wrong... but I cannot trigger it
         pos = 250;
 
     sb.SetScrollbar(pos, 20, 500, 20);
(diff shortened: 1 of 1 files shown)


retro-wertz:
nope. this does not fix the problem. this does allow to scroll but memory locations are invalid. Actually only the 0 - 0x2B0 is correct and the rest are invalid now.
Screenshot_2019-08-10_18-10-43

it even loops back to 0x0 once you go past 0xFFFFFFFF (which invalid to begin with since usable memory is only up to 0xE00FFFF)
Screenshot_2019-08-10_18-15-50

denisfa:
@rkitover Would you mind merging this? This is not the definitive solution, but at least restores the old behaviour with the corrected values (not giving an error message when we try to open it via menu).

I will keep the original issue open. I cannot tackle this at the moment because I don't quite understand what I have to see in the memory rows.

rkitover:
Of course.

PR #490, opened 2019-08-20
Auto hold buttons.

denisfa:
@rkitover All the issues opened by this user are fun too code. I tested this on `Super Mario Land` (GB).

This is based on `vba-rr` (as in done the same way).

I also had to do some fixes, but attached them here because only discovered them while coding this.


rkitover:
This is fantastic, just please try to write more descriptive commit messages (after the top summary line.)

Describe what you did and how.

denisfa:
@rkitover Ok, I will be more descriptive next time.

PR #491, opened 2019-08-21
Allow UPS patch for GBA games.

### LINKED ISSUE #487: UPS soft-patching fails to work, runs the rom unpatched
### Expected behavior

Soft-patching should just works in most cases unless patch itself is damaged.

### Actual behavior

No softpatching happens when using .ups file

in this example, ill be using .ups named "Pokemon Adventure : Red Chapter 14.5", which itself is a rom patch for Pokemon Fire Red Version 1.0.

1. Download the .ups patch from here: 
2. Rename the file accordingly with the same name of the original rom (aka Pokemon FireRed 1.0)
3. Load the game. If patching works, you should be seeing this title screen instead or the original.

Screenshot_2019-08-15_09-20-33

The failure is because before a patch is applied, the rom size is padded to 0x2000000. UPS patching checks hash for the patch file, the target file and the expected output file after patching, but since the rom size is padded, the hash check portion will cause patching to fail. 

### ROM details

Which games are affected by the issue (please state if GBA or GB game): any using .ups patching

Which games, if any, are NOT affected by the issue (please state if GBA or GB game):

Include screenshots of ROM Information (File > ROM Information):

If you are using a BIOS file, please say which one.

### Option details

Video driver (simple, opengl):
gl

Sound driver (directsound, xaudio, openal, SDL):
sdl

### Build details

archlinux

If using Linux, specify if using xorg or Wayland:
xorg

built from master branch 6b18c1f9b886a7241c726518665e12373c5c3942 

### Hardware details:

If your joystick or gamepad isn't working in Linux, please try to troubleshoot
using the information on this page:
https://wiki.archlinux.org/index.php/Gamepad **BEFORE** filing an issue. We use
SDL for joystick support, any problems will generally be an issue with SDL and
**NOT** the emulator.

If relevant, the joystick you are using and the drivers you are using for it:

denisfa:
From my very superficial understanding, the `rom[]` resize happens only when `rom_size` is greater than `0x2000000`.

We should probably study this better for an actual explanation.

- Fix #487.

This is a little shady for me... As I don't quite understand what is the problem of resizing `rom[]`. This solution is very simple, though. It should still protect the case for when the resize could happen. This is only for GBA ROMs, it seems (because they are bigger, probably);

### DIFF
--- a/src/wx/panel.cpp
+++ b/src/wx/panel.cpp
@@ -216,7 +216,7 @@ void GameArea::LoadGame(const wxString& name)
         if (loadpatch) {
             // don't use real rom size or it might try to resize rom[]
             // instead, use known size of rom[]
-            int size = 0x2000000;
+            int size = 0x2000000 < rom_size ? 0x2000000 : rom_size;
             applyPatch(pfn.GetFullPath().mb_str(), &rom, &size);
             // that means we no longer really know rom_size either <sigh>
         }


rkitover:
If it works and doesn't break anything I have no problem with it. I won't claim that I understand how a lot of this code works either.

PR #494, opened 2019-08-28
Update to xBRZ `1.7`.

### LINKED ISSUE #164: Xbrz 1.5 released
Xbrz 1.5 has been released. This is not an issue, but i think Vba-m can be faster with this plugin with the latest improvements.

denisfa:
@rkitover 
```
Get files from here `https://sourceforge.net/projects/xbrz/files/xBRZ/`.

Then, update the src to handle pitch params. For our case, the pitch
is necessary because we deal with borders (top and right) of our source
image. Normally, we would want to scale without it, and therefore we
need to adjust the pointers to skip the borders.

If we have a `width + 1 pixel border` per line, the we need to scale
a image with `width` line size, but advancing the pointer for each new
line processed including the border on the count.

Also, since our output pointer also allocates for the border, we need
to adjust the output moving pointer for each line in a custom way.
(output border in this case)

- Fix #164.
```

The best way to review this is to get the source code from the latest xBRZ and compare how the files were updated. The changes I did are very minimal and do not interfere anyhow with the core from these filters.

This is a also a case where I know what is happening, but might fail to explain it in english. Check if you can understand. The diagonal line you saw when trying to update to `1.6` was due the filter trying to scale a `width + border` image pointer, but only advancing `width` memory per line processed. The output scaled was doing the same thing.

Moving with this, we can also remove the branch `xbrz1.6`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -814,8 +814,9 @@ set(
     src/filters/hq2x.h
     src/filters/interp.h
     src/filters/lq2x.h
-    src/filters/xBRZ/config.h
+    src/filters/xBRZ/xbrz_config.h
     src/filters/xBRZ/xbrz.h
+    src/filters/xBRZ/xbrz_tools.h
 )
 
 set(
--- a/src/filters/xBRZ/xbrz.cpp
+++ b/src/filters/xBRZ/xbrz.cpp
@@ -1,13 +1,14 @@
 // ****************************************************************************
-// * This file is part of the HqMAME project. It is distributed under         *
-// * GNU General Public License: http://www.gnu.org/licenses/gpl-3.0          *
+// * This file is part of the xBRZ project. It is distributed under           *
+// * GNU General Public License: https://www.gnu.org/licenses/gpl-3.0         *
 // * Copyright (C) Zenju (zenju AT gmx DOT de) - All Rights Reserved          *
 // *                                                                          *
 // * Additionally and as a special exception, the author gives permission     *
-// * to link the code of this program with the MAME library (or with modified *
-// * versions of MAME that use the same license as MAME), and distribute      *
-// * linked combinations including the two. You must obey the GNU General     *
-// * Public License in all respects for all of the code used other than MAME. *
+// * to link the code of this program with the following libraries            *
+// * (or with modified versions that use the same licenses), and distribute   *
+// * linked combinations including the two: MAME, FreeFileSync, Snes9x, ePSXe *
+// * You must obey the GNU General Public License in all respects for all of  *
+// * the code used other than MAME, FreeFileSync, Snes9x, ePSXe.              *
 // * If you modify this file, you may extend this exception to your version   *
 // * of the file, but you are not obligated to do so. If you do not wish to   *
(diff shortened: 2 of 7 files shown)


rkitover:
Nice work dude, I wanted to do this for like 2 years now.

rkitover:
There are some C++ compatibility issues you can see in the CI builds, I will take care of that later.

Two things for you:

- Does the distribution come with an assembly implementation? Please check the the asm xbrz still works with `-DENABLE_ASM=ON` for 32 bit builds, if not remove the asm version. That reminds me I need to turn on that option for our 32 bit release builds. If there is a 64 bit asm version, I can do something to add it into our build system.

- Please use `rebase -i` to squash these two commits together.

rkitover:
Also please check that the change you made to `panel.cpp` works with the other filters too.

denisfa:
> Does the distribution come with an assembly implementation? Please check the the asm xbrz still works with -DENABLE_ASM=ON for 32 bit builds, if not remove the asm version. That reminds me I need to turn on that option for our 32 bit release builds. If there is a 64 bit asm version, I can do something to add it into our build system.

There is only ASM code commented out.
```
//inline
//double fastSqrt(double n)
//{
//    __asm //speeds up xBRZ by about 9% compared to std::sqrt which internally uses the same assembler instructions but adds some "fluff"
//    {
//        fld n
//        fsqrt
//    }
//}
//
```

If I want to use this, what should I examine? What kind of ASM code breaks our builds?

> Please use rebase -i to squash these two commits together.

Ok, I forgot about my printf helpers.

> Also please check that the change you made to panel.cpp works with the other filters too.

They work. I will explain the changes on the code. They should be easy to understand.

rkitover:
Makes sense, but things like this should really be in the commit message.

Please ignore what I said about asm filters, it looks like the only asm filters we have are `2xSaImmx` and `hq`.

rkitover:
Ok I am merging this now and I will take care of the build related issues later today.

Thank you very much for doing this!

denisfa:
@rkitover what about the `_asm` code? Can I enable it and use? The comment says there will be better performance...

rkitover:
Yeah we could, but it's not that simple:

- it would have to be enabled using the appropriate `#ifdef`s **ONLY** on `x86` and `x86-64` (remember people use this on things like raspberry pis etc. which are arm.)

- you'd want to make sure it works on **BOTH** `x86` and `x86-64`, or possibly use different versions for each

- the inline assembly syntax varies between compilers, you may need different versions on gcc/clang and visual studio, fortunately we have appveyor now for easy checking of visual studio compatibility.

Given all of this, it may not be worth the effort, but up to you, I could also look at this later if you want.

rkitover:
Also, I still occasionally play a game of advance wars with xbrz6 on my 43" monitor, so this is an improvement in my life, thank you :D

denisfa:
Most of this for me is a learning experience. So I will take a look. At the very least I want to make it work for `x86_64` so I can use it.

> Also, I still occasionally play a game of advance wars with xbrz6 on my 43" monitor, so this is an improvement in my life, thank you :D

You are welcome!

PR #498, opened 2019-09-02
Check ASAN support when `ENABLE_ASAN=ON`.

### LINKED ISSUE #409: cmake: add compiler check for ASAN
I need to compile VisualBoyAdvance-M with Network Link, because it's not enabled for some reason in the official build of 2.1.1, however going to compile it complains that the compiler can't find SFML or SSP (gcc-4.8) and GTK (gtk-2.0/gtk-3.0) even though they exist, dev packages and all.

```
/usr/include/SFML
[...]
/usr/share/SFML
/usr/share/SFML/cmake
/usr/share/SFML/cmake/Modules
/usr/share/SFML/cmake/Modules/FindSFML.cmake
/usr/include/gtk-2.0
[...]
/usr/include/gtk-3.0
```

```
$ cmake ..
Could NOT find SFML (missing:  SFML_NETWORK_LIBRARY SFML_SYSTEM_LIBRARY)
CMake Warning at CMakeLists.txt:232 (message):
  SFML not found, LINK will be disabled


-- Found gcc-ar: /usr/bin/gcc-ar
-- Found gcc-nm: /usr/bin/gcc-nm
-- Found gcc-ranlib: /usr/bin/gcc-ranlib
-- Could NOT find SSP (missing: SSP_LIBRARY) 
CMake Error at src/wx/CMakeLists.txt:538 (message):
  Could not find gtk


-- Configuring incomplete, errors occurred!
```

Looks to me that it's not looking in the right places to find them or using the wrong syntax in the code to search for them, the proper directory is `/usr/include/` is where it's suppose to search for the libraries it needs for compiling, and SSP is has since been included in GCC since 4.1. The error message for "Could not find gtk" is too vague of a message, what are you expecting? libgtk-2.0(-dev)? gtk-3.0(-dev)?

[CMakeOutput.log](https://gist.github.com/Alkaris/c46f6ede5ea281cfac6b4e20dfd83db4#file-cmakeoutput-log)
[CMakeError.log](https://gist.github.com/Alkaris/2f47cda1db096e55c48b496babe488ea#file-cmakeerror-log)

```
Linux Mint 19.1 4.15.0-46-generic #49-Ubuntu x86_64 GNU/Linux
```

denisfa:
```
We check if the compiler has asan support when the option is enabled.
If it does not, then we stop compiling with the proper message.

To test this, we use `gcc (GCC) 4.4.7` (no asan support) and
`gcc (GCC) 9.1.0` (asan is supported). Call cmake like this:

`cmake .. -DCMAKE_BUILD_TYPE=Debug -DENABLE_ASAN=On` and
`CXX=g++-4.4 cmake .. -DCMAKE_BUILD_TYPE=Debug -DENABLE_ASAN=On`

we then check that it fails for the older version and it works for
the newer gcc, like expected.

- Fix #409.
```

@rkitover  I know this PR has weird timing, but I try to match whatever tool I am handling at work with the closest thing we have here. This way I can double practice whenever I have to do stuff. Check if this is enough.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -480,6 +480,8 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID STREQUAL Clang)
         set(MY_C_DBG_FLAGS -g -fno-omit-frame-pointer)
     endif()
 
+    include(CheckCXXCompilerFlag)
+
     if(ENABLE_ASAN)
         string(TOLOWER ${CMAKE_BUILD_TYPE} build)
         if(NOT build STREQUAL "debug")
@@ -490,7 +492,13 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID STREQUAL Clang)
         if(SANITIZER STREQUAL "on" OR SANITIZER STREQUAL "true")
             set(SANITIZER address)
         endif()
-        set(MY_C_DBG_FLAGS ${MY_C_DBG_FLAGS} -fsanitize=${SANITIZER} -lasan)
+        set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} "-lasan")
+        check_cxx_compiler_flag("-fsanitize=${SANITIZER} -lasan" ASAN_SUPPORT_FLAG)
+        if(${ASAN_SUPPORT_FLAG})
+            set(MY_C_DBG_FLAGS ${MY_C_DBG_FLAGS} -fsanitize=${SANITIZER} -lasan)
+        else()
+            message(FATAL_ERROR "asan not available to compiler.")
+        endif()
     endif()
 
     # common flags
@@ -499,8 +507,6 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID STREQUAL Clang)
     if(NOT (WIN32 OR X86_32)) # inline asm is not allowed with -fPIC
         set(MY_C_FLAGS ${MY_C_FLAGS} -fPIC)
     endif()
-
-    include(CheckCXXCompilerFlag)
 
     # check if ssp flags are supported for this version of gcc
     if(CMAKE_COMPILER_IS_GNUCXX)


CMakeLists.txt
```diff
@@ -490,7 +492,13 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID STREQUAL Clang)
         if(SANITIZER STREQUAL "on" OR SANITIZER STREQUAL "true")
             set(SANITIZER address)
         endif()
-        set(MY_C_DBG_FLAGS ${MY_C_DBG_FLAGS} -fsanitize=${SANITIZER} -lasan)
+        set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} "-lasan")
+        check_cxx_compiler_flag("-fsanitize=${SANITIZER} -lasan" ASAN_SUPPORT_FLAG)
+        if(${ASAN_SUPPORT_FLAG})
```
rkitover: When checking if a variable is true, it's not necessary to use the value,

```cmake
if(ASAN_SUPPORT_FLAG)
```

will work fine.

CMakeLists.txt
```diff
@@ -490,7 +492,13 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID STREQUAL Clang)
...
+            set(MY_C_DBG_FLAGS ${MY_C_DBG_FLAGS} -fsanitize=${SANITIZER} -lasan)
+        else()
+            message(FATAL_ERROR "asan not available to compiler.")
```
rkitover: Better phrasing would be something like "your version of gcc does not support the -fsanitize option" or something.

CMakeLists.txt
```diff
@@ -490,7 +492,13 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID STREQUAL Clang)
         if(SANITIZER STREQUAL "on" OR SANITIZER STREQUAL "true")
             set(SANITIZER address)
         endif()
-        set(MY_C_DBG_FLAGS ${MY_C_DBG_FLAGS} -fsanitize=${SANITIZER} -lasan)
+        set(CMAKE_REQUIRED_LIBRARIES ${CMAKE_REQUIRED_LIBRARIES} "-lasan")
+        check_cxx_compiler_flag("-fsanitize=${SANITIZER} -lasan" ASAN_SUPPORT_FLAG)
```
rkitover: Since you added `-lasan` to `CMAKE_REQUIRED_LIBRARIES` it's not necessary to pass it into the `check_cxx_compiler_flag()` call, but harmless.

denisfa:
@rkitover Thanks for the feedback. I will keep track of your refactoring to learn more.

PR #501, opened 2019-09-07
Languages

### LINKED ISSUE #483: Two words not translated into Brazilian Portuguese (pt_BR)
### Expected behavior
Putting the translation, the emulator (less a few words, because there is no direct translation) would be translated

### Actual behavior
Only two words are not translated, which are Open (GBA) and Quit
Open=Abrir
Quit=Sair

Put the translation in Brazilian Portuguese

Sem título

denisfa:
@rkitover This is a bunch of stuff regarding the translation issues:

9784766 : these were never being translated or appeared on the po files because they were not on any XRC and strings from `cmdevents` are processed by cmake syntax.  So we could not use `_()` for gettext.

0b6bcdc : wxWidgets has some issues when translating internal strings (such as the buttons you would expect to be translated automatically). This is due how there is a difference between user and system locale (easy example on commit message about `pt_BR` and `en_US` on 837a48c).

The two above should fix #483 (I did the `pt_BR` translation entirely, and the user of hungarian localization can use transifex to solve the ones he complains there).

3c2acf9 : new operations regarding save states.

A bit unrelated but discovered while working on this:

a572a6a and 5de33cd : just some work that was missing when using key shortcuts for these options (they would not work before, much like the pause issue we had some time ago).

### DIFF
--- a/src/wx/cmdevents.cpp
+++ b/src/wx/cmdevents.cpp
@@ -2803,8 +2803,10 @@ EVT_HANDLER(ApplyPatches, "Apply IPS/UPS/IPF patches if found")
 
 EVT_HANDLER(MMX, "Enable MMX")
 {
+#ifdef MMX
     GetMenuOptionInt("MMX", disableMMX, 1);
     update_opts();
+#endif
 }
 
 EVT_HANDLER(KeepOnTop, "Keep window on top")
--- a/src/wx/cmdevents.cpp
+++ b/src/wx/cmdevents.cpp
@@ -1749,12 +1749,20 @@ EVT_HANDLER_MASK(SaveGameSlot, "Save current state slot", CMDEN_GB | CMDEN_GBA)
 EVT_HANDLER_MASK(IncrGameSlot, "Increase state slot number", CMDEN_GB | CMDEN_GBA)
 {
     state_slot = (state_slot + 1) % 10;
+
+    wxString msg;
+    msg.Printf(_("Current state slot #%d"), state_slot);
+    systemScreenMessage(msg);
 }
 
 // new
 EVT_HANDLER_MASK(DecrGameSlot, "Decrease state slot number", CMDEN_GB | CMDEN_GBA)
 {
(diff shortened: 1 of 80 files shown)


src/wx/cmdevents.cpp
```diff
@@ -2783,8 +2872,10 @@ EVT_HANDLER(ApplyPatches, "Apply IPS/UPS/IPF patches if found")
 
 EVT_HANDLER(MMX, "Enable MMX")
 {
+#ifdef MMX
```
rkitover: I'm not sure we should bother with this, every CPU for more than like two decades has MMX, we should just remove all references to this, but this can be done later.

src/wx/guiinit.cpp
```diff
@@ -2830,6 +2839,17 @@ bool MainFrame::BindControls()
                 continue;
 	    }
 #endif
+#ifndef MMX
```
rkitover: See above.

rkitover:
I cleaned up the history and merged it.

denisfa:
@rkitover Thank you! I was playing the free weekend of KCD and completely forgot about this... Sorry!

src/wx/cmdevents.cpp
denisfa: This is enabled only for x32 builds. Can we enable for all? I believe there should not be any issues with x64, but what about ARM and others? I have no idea.

rkitover: Yes x64 definitely has MMX, assuming there is no assembly involved, 32 bit assembly doesn't always work for x64 builds.

rkitover:
Not a problem, that looks like a fun game! I think I might try it.

src/wx/cmdevents.cpp
rkitover: And if there is assembly involved, it may not be better than what the compiler generates with the appropriate `-march` etc. flags, which we need to revisit at some point.

PR #506, opened 2019-09-16
Leftovers.

denisfa:
@rkitover These are some leftovers from a couple of issues.

ebe1be7: Some UTF-8 messages were being displayed wrong.

387a1c8: the `xBRZ` optimization for some platforms. The only one missing from what I wanted is the `Windows x64` because it does not accept inline ASM. It is too much work to make it worth for visual studio.

a6f5848: I am studying `WinSparkle` for a definitive solution. So, although it fixes the crash, I will let the issue open for now.

4998a5d: From
menulink-old

to

menulink-new

### DIFF
--- a/src/filters/xBRZ/xbrz.cpp
+++ b/src/filters/xBRZ/xbrz.cpp
@@ -64,16 +64,23 @@ uint32_t gradientARGB(uint32_t pixFront, uint32_t pixBack) //find intermediate c
 }
 
 
-//inline
-//double fastSqrt(double n)
-//{
-//    __asm //speeds up xBRZ by about 9% compared to std::sqrt which internally uses the same assembler instructions but adds some "fluff"
-//    {
-//        fld n
-//        fsqrt
-//    }
-//}
-//
+inline double fastSqrt(double n)
+{
+#ifdef __GNUC__ || __clang__ || __MINGW64_VERSION_MAJOR || __MINGW32_MAJOR_VERSION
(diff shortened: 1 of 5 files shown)


ZachBacon:
Looking at some of the changes and it looks good on my end. As for the update checker, winsparkle would allow a notification for users of a new version and offer to update their app. Based off of the sparkle framework for macOS. I think it runs it's own check each time it detects (and this is entirely from memory) a change in a specific file on a server that has a different version than the one it's on currently. So this idea would basically rip out the current system in place and drop in a new one. Not sure if it could be adapted in its current form.

denisfa:
@ZachBacon We will need a website for it to pull information about updates. We already have one right? We will need to add some files (xml, nevertheless) for the download and check.

I believe it should be easy to code, the issue is that it is windows only and it has been a long time since I had to do something for this OS.

denisfa:
> So this idea would basically rip out the current system in place and drop in a new one. Not sure if it could be adapted in its current form.

For windows. I guess we could go forward with that idea of disabling the update for Linux (and ARM), use `winsparkle` for windows and then doing something specific to macOS.

ZachBacon:
@denisfa the XML can be stored in git to be honest, whenever we do tag release or a nightly, we adjust the file and update the url with the executables location.

rkitover:
Awesome, thank you!

@ZachBacon is just finishing up updating the builder scripts for ffmpeg etc. and then we should be good to go for a release.

PR #508, opened 2019-09-25
SDL misc.

denisfa:
This is meant to remove one of the branches (`arthur/sdl2`).

Also, there was a weird behaviour when using the command line of the sdl port. There was an overriding between `LoadConfig()` and `ReadOpts()`. We get the `--config` on `ReadOpts`, but only load the default options or `vbam.ini`. The command line options were also getting ignored. Now we should have order priority:

`$ ./vbam -O 2 -f 10 -I 2 -c vbam.ini ~/Downloads/loz/loz.zip` uses the `vbam.ini` options over the others, meanwhile
`$ ./vbam -c vbam.ini -O 2 -f 10 -I 2  ~/Downloads/loz/loz.zip` uses the command line over `vbam.ini`

This was somewhat annoying to dealt with it.

### DIFF
--- a/src/common/ConfigManager.cpp
+++ b/src/common/ConfigManager.cpp
@@ -898,6 +898,7 @@ int ReadOpts(int argc, char ** argv)
 			preferences = NULL;
 			OpenPreferences(optarg);
 			fclose(f);
+			LoadConfig();
 		}
 		break;
 		case 'd':
--- a/src/sdl/SDL.cpp
+++ b/src/sdl/SDL.cpp
@@ -494,27 +494,9 @@ void sdlOpenGLInit(int w, int h)
 {
     (void)w; // unused params
     (void)h; // unused params
-#if 0
-  float screenAspect = (float) sizeX / sizeY,
-        windowAspect = (float) w / h;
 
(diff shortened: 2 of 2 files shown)


ZachBacon:
Looked at it myself, everything looks good, I did a test compile on my end and it looks pretty good on my end too. Things still work, the config loading works on my end based on your scenario for me.

denisfa:
@ZachBacon Thank you for reviewing this. I tag @rkitover for everything, but feel free to also comment whatever you wish about the code of any pr. If there is something more specific to you than him, please let me know.

The SDL port is really interesting, but it is in this weird position where it feels like a completely different app... Even some key names for its config are different from the wx port. Some are ignored (as in no effect whatsoever). Not sure about the possibilities, but it will likely need an overhaul at some point.

ZachBacon:
@denisfa yeah no worries, he's better than me any way. I just try my best to manage things, set goals etc, been trying to be more active. But as for the SDL port, that's been around well since the 1.7 days when forgetten was still developing the original. so... if I had to wager, 15 years or so it's been around.

PR #514, opened 2019-10-02
smarter ver on win build

ArtiiP:


ArtiiP:
for now version.h don't have numbers in good format for rc, ie
we need 
2,1,4,0 and "2.1.4.0"

ZachBacon:
we could probably do a few changes to make it work. Perhaps a macro or something.

ArtiiP:
btw,  it's any chance, to configure  downloadable artifacts with appveyor

rkitover:
@ArtiiP thank you this is perfect.

I hate to be an ass, but could you please write a commit message?

This is how to write git commit messages:

https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

you may need to configure your editor to do correct line wrapping.

ArtiiP:
that's hard part ;)
i think reusing this pullreq title:
_smarter version on win build_
or something like this:
_automatic generation version number for rc (win build)_

with one?

ZachBacon:
the title would be more along the lines of [Platform] Platform being in this case windows with what follows "automatic generation version number for rc" with a small summary of what you did, based on what @rkitover link he posted.

rkitover:
@ArtiiP I will look into making appveyor artifacts available etc.. But first I need to do more work on the visual studio support.

I want to change it to static vcpkg builds and use ninja instead of msbuild, and get debug builds to work, lots to do.

PR #515, opened 2019-10-03
[WIP - DO NOT MERGE] Update checker for windows using Winsparkle.

denisfa:
@rkitover @ZachBacon I need a little help setting up the new lib. I am using [winsparkle 0.6.0](https://github.com/vslavik/winsparkle/releases), but I don't know how to do the proper procedure to enable it only for windows builds. There are 2 headers needed.

They have it available for `x86` and `x64` (both `.lib` and `.dll`). Once this is solved, the following steps are much easier: decide where we need to stock an XML file (`appcast.xml`) and it is pretty much done.

I will refactor this, both code and commits messages, but really need to be able to compile without doing all the stuff I had to do these days.

At least something works:
l

This is a test I did, so ignore the obvious error of version mismatch. Clicking `Get Update` opens the browser for the new version download.

Can you help me?

EDIT: None of the compilation tests for Windows will go through at the moment, because I needed to do a lot of manual operations to get it to compile. So ignore these for now.


data/appcast.xml
```diff
@@ -0,0 +1,33 @@
...
+          sparkle:os="windows-x86"
+          sparkle:version="2.1.4"
+          length="0"
+          type="application/octet-stream"
+      />
+    </item>
+    <item>
+      <link>https://github.com/visualboyadvance-m/visualboyadvance-m/releases/download/v2.1.4/visualboyadvance-m-Mac-64bit.zip</link>
+      <enclosure
+          sparkle:os="macos"
```
denisfa: @ZachBacon There is the `sparkle` lib for macos, which shares the same fields here. We could use this for macos too. What do you think? I know nothing about updating apps on macos, but I can tell that it should be easy to do it for us this way.

src/wx/wxvbam.cpp
```diff
@@ -185,6 +185,27 @@ wxString wxvbamApp::GetAbsolutePath(wxString path)
...
+{
+#ifdef __WXMSW__
+    std::wstring version;
+#ifndef FINAL_BUILD
+    version = _(VERSION);
+#else
+    version = wxT("0");
+#endif // FINAL_BUILD
+    //win_sparkle_set_appcast_url("https://github.com/visualboyadvance-m/visualboyadvance-m/data/appcast.xml");
+    win_sparkle_set_appcast_url("https://github.com/visualboyadvance-m/visualboyadvance-m/tree/update-checker/data/appcast.xml");
```
denisfa: @ZachBacon This is for testing only, but where should we save this file for us?

data/appcast.xml
ZachBacon: well winsparkle is actually inspired directly from sparkle, which is partly why I was suggesting winsparkle

src/wx/wxvbam.cpp
ZachBacon: visualboyadvance-m/data/appcast.xml seems alright with me, but so does the current location, as long as the updater can find it really.

denisfa:
@rkitover  @ZachBacon  So, just as a FYI, `Winsparkle` can also download a `setup.exe` and run it automatically after download. This is not so interesting for us because we run as a portable exe on windows. There is no such option to just replace the current exe with a downloaded one (at least for now). 

Creating a installer, at least according to what I researched, is an overkill for what we want. So I decided to go with the `link-to-download`option.

ZachBacon:
> 
> 
> @rkitover @ZachBacon So, just as a FYI, `Winsparkle` can also download a `setup.exe` and run it automatically after download. This is not so interesting for us because we run as a portable exe on windows. There is no such option to just replace the current exe with a downloaded one (at least for now).
> 
> Creating a installer, at least according to what I researched, is an overkill for what we want. So I decided to go with the `link-to-download`option.

The only thing I can think of is just a self extracting archive, like a 7zip sfx archive. But most likely better to download the zip and have the user manually extract it.

data/appcast.xml
denisfa: Well, once you get the right libs for compilation it is a piece of cake to code it.

denisfa:
> The only thing I can think of is just a self extracting archive, like a 7zip sfx archive. But most likely better to download the zip and have the user manually extract it.

I saw a tutorial for that, but it feels so weird. We could create a `msi` from a XML using `msitools`. I just don't think it is needed at all for us. We don't use the windows registry or anything advanced.

ZachBacon:
if we were to have an installer, I'd rather nsis. And even then just a simple 7z self extracting archive.

rkitover:
@denisfa good shit, some notes:

- put the `.lib` files you need in the `dependencies` submodule, it's a repo under our org, we use it for win32 stuff

- do not use `FINAL_BUILD` for anything, we need to remove it, we will always have a version, if not from git then from the changelog

- don't use `version.h` use `version_cpp.h` and `version.cpp` which I just added yesterday, these are for speeding up rebuilds

- later on I will make a powershell script that we'll compile in that will do the update in-place, but don't worry about this for now

denisfa:
@rkitover I believe I know what should be done (as in to get it to work), but I am too busy at the moment to actually test it. I will need some days to finish this. I will probably nuke the current work for a new one.

> * put the `.lib` files you need in the `dependencies` submodule, it's a repo under our org, we use it for win32 stuff

Should we compile it ourselves? I was wondering how we link the SFML and FFMPEG libs for windows... Is this something like adding the `.lib` to the final exe?

> * do not use `FINAL_BUILD` for anything, we need to remove it, we will always have a version, if not from git then from the changelog

What is the proper way to get the version? Just the macro?

> * don't use `version.h` use `version_cpp.h` and `version.cpp` which I just added yesterday, these are for speeding up rebuilds

Ok.

> * later on I will make a powershell script that we'll compile in that will do the update in-place, but don't worry about this for now

Is this to replace the current method? Or something like a `setup.exe`?

rkitover:
> I believe I know what should be done (as in to get it to work), but I am too busy at the moment to actually test it. I will need some days to finish this. I will probably nuke the current work for a new one.

Don't worry about it, this can definitely wait.

> Should we compile it ourselves? I was wondering how we link the SFML and FFMPEG libs for windows... Is this something like adding the .lib to the final exe?

We can. We compile sfml and ffmpeg ourselves using my build script. I will look into adding it into my scripts.

> Is this to replace the current method? Or something like a setup.exe?

I mean once you're done with your work, I will have it run a script to update the `.exe` in-place using the `.zip` file using something like a compiled-in powershell script. We are not going to change the release process for this.

ZachBacon:
@denisfa Take your time, first and foremost, Life always comes first, in the end you got to take care of yourself before you finish anything here, this is all volunteer work, and we appreciate what you do.

rkitover:
I think @ZachBacon said there was a mingw package in msys2, I'll add the relevant cmake glue to make it work first.

denisfa:
@ZachBacon Thank you! I really dislike to leave unfinished work on the queue. Some of this happens because I add too much stuff in my To-Check/Study list, sometimes the majority of something I want to code.

My issue with this is the following: `winsparkle` is meant to be a shared lib. It is not supported to be compiled as static. I try to do it myself (the only supported platform is visualstudio), and got a `.lib` sized 1GB... So, something is not right.

I want to get an `exe` which does not need additional `dll` to work, same directory or installed alongside.

Adding to the `cmake` files is easy, I just need to format it properly and get it adjusted to our current configuration. @rkitover I have something like a skeleton of what I want to do. Since this is wip, I will push these commits here so you can take a look. Then I change it accordingly to what we discuss.

denisfa:
@ZachBacon @rkitover If we go forward with this, we will very likely need to also update the `dependencies` repo. At least for the current solution I needed my own branch.

src/wx/CMakeLists.txt
```diff
@@ -763,6 +763,19 @@ target_link_libraries(
...
+        add_library(WinSparkle STATIC IMPORTED)
+        if(AMD64)
+            set_property(TARGET WinSparkle PROPERTY IMPORTED_LOCATION "${CMAKE_SOURCE_DIR}/dependencies/WinSparkle/x64/Release/WinSparkle.lib")
+        else()
+            set_property(TARGET WinSparkle PROPERTY IMPORTED_LOCATION "${CMAKE_SOURCE_DIR}/dependencies/WinSparkle/Release/WinSparkle.lib")
+        endif()
+        target_link_libraries(visualboyadvance-m WinSparkle)
+    endif()
+endif()
+
```
denisfa: @rkitover I believe in something like this, but probably not here. What would be the best location for this scope to be?

ZachBacon:
https://github.com/msys2/MINGW-packages/tree/master/mingw-w64-winsparkle here's the relevant stuff that msys2 does to make their static copy of winsparkle

denisfa:
@ZachBacon Thank you for the link! But I don't think that is an actual static lib. Most source has stuff like `__declspec(dllexport)` or `__declspec(dllimport)`. Even with this compilation, the final result means a DLL. If you check the symbols from the `.a`, you have stuff like `__imp_*`. So the linking should fail.

I have very little experience with windows code, so I may be misunderstanding this.

ZachBacon:
@denisfa So I did an objdump -t of libwinsparkle.a from msys2. https://hasteb.in/ehiziram.md The only __imp stuff I'm seeing is for native windows libs such as __imp_VerifyVersionInfoW for kernel32,dll

however libwinsparkle.dll.a https://hasteb.in/lasowake.md different story. So yes libwinsparkle is a static library.

denisfa:
@ZachBacon Thank you for all the info, but I hit a wall yet again: on file `ui.cpp`, we have this on `void UI::Run()`:
```
if ( !ms_hInstance )
    {
        // If DllMain() was not called, assume we're statically linked
        // and use the hInstance of the containing program.
        ms_hInstance = GetModuleHandle(NULL);
    }
```
It is `dllmain.c` that sets this on `dllmain.cpp` (func `DllMain`). I believe this is related to the next issue.

When I try to run vbam with the `libwinsparkle.a`, it links (I had to compile again, though), but nothing works. The vbam window opens and if I go to `about` > `check for updates`, it opens another vbam window instead of winsparkle ui. The initial startup check also fails silently.

Do you have any idea what may be the cause of this?

ZachBacon:
@denisfa this maybe an issue how it's calling things, let's leave winsparkle as a shared build then. One DLL shouldn't be an issue to be honest.

denisfa:
@ZachBacon for confirmation: we ship with the dll on the zip? I am testing a solution to see if we can load the dll at runtime so we can check if it is there. If not, disable the update checking. Do you think it is worth? Or is it better to save for later?

ZachBacon:
Yes we ship the DLL in the zip. And I don't know if it's worth it, but if you think you can, can't hurt to try.

denisfa:
> No, we are not going to ship any dlls. I will fix this.

@rkitover What solution do you have in mind here?

ZachBacon:
My guess, in a static link, we would have to find a way to call the winsparkle UI internally instead of a callback, which would have to done via winsparkle itself I guess. But that's what I can think of on the top of my head, and even then that may not exactly be the correct approach.

denisfa:
@ZachBacon @rkitover I was thinking about this and I believe we should take a step back and scrap it for now. Unless you have some wip, I would like to close this pr.

I want to try the sparkle for macos first before this. Check how it is going to work and how to do it on C++. I will continue to work on this branch, so we don't need to delete it. If the work goes further, I will open another pr.

ZachBacon:
@denisfa I think @rkitover still has some ideas he wants to tinker around with first.

rkitover:
@denisfa so I took care of the DLL packaging issue and your code works fine with it. Did you want to do anything else with it?

denisfa:
@rkitover I wanted to create a namespace for the updater. Since last time, I was having a go on using the sparkle for a macOS updater too, but did not have much time to work on it. This is mostly a code refactoring, so if you want, close this pr and I will try to submit a new one (macOS sparkle auto-updater and isolate all this code on a proper namespace).

rkitover:
@denisfa well there is a lot of useful code in this PR, both yours and mine.

Any objects if I just disable the cmake option for the time being, clean up the history and merge all of this. You could then pick up where you left off later.

denisfa:
> Any objects if I just disable the cmake option for the time being, clean up the history and merge all of this. You could then pick up where you left off later.

Alright, I will create a new branch for all the work I want to do soon.

rkitover:
Merged to master with `ENABLE_ONLINEUPDATES` disabled.

PR #521, opened 2019-10-05
Fix key accel overriding menus and dialogs.

### LINKED ISSUE #516: VBA-M doesn't let me type any number in any preference window
I was trying to configure the cable link in server-client mode when I noticed I couldn't fill forms like timeout or server IP.

The emulator captured automatically the key 1-6 and resize the window despite not having focus and being paused, so I couldn't write those number.

A possible workaround would be disabling those shortcuts, but it's not a normal behavior and should be fixed.
Option->Link->Configure (type 123456 numbers)

This happens in the recent VBA-M 2.1.4 released yesterday, with any rom I tested.

denisfa:
```
All keys were being captured before being processed by any window of the
app to check for accelerators. This meant being impossible to use any
keys for input if they were saved for an accel.

- Fix #516.
```

Sorry about this. This slip through because my custom desktop settings.

### DIFF
--- a/src/wx/wxvbam.cpp
+++ b/src/wx/wxvbam.cpp
@@ -821,7 +821,7 @@ void MainFrame::OnSize(wxSizeEvent& event)
 
 int MainFrame::FilterEvent(wxEvent& event)
 {
-    if (event.GetEventType() == wxEVT_KEY_DOWN)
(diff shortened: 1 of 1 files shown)


rkitover:
Awesome, thank you very much for the quick fix, I will make a new release.

PR #532, opened 2019-10-08
cmake: Don't install visualboyadvance-m twice on WIN32.

orbea:
On the `mingw` travis builds it will install `visualboyadvance-m` twice.
```
-- Installing: /tmp/VBAM/home/travis/build/visualboyadvance-m/visualboyadvance-m/build/visualboyadvance-m.exe
-- Installing: /tmp/VBAM/usr/lib/mxe/usr/x86_64-w64-mingw32.static/bin/visualboyadvance-m.exe
```
This will make it so that only the second instance is installed.
```
-- Installing: /tmp/VBAM/usr/lib/mxe/usr/x86_64-w64-mingw32.static/bin/visualboyadvance-m.exe
```
I am not sure there is a good reason why its being installed twice or not and am opening this PR for review.

I also removed the unused `WX_EXE_NAME` variable.

### DIFF
--- a/src/wx/CMakeLists.txt
+++ b/src/wx/CMakeLists.txt
@@ -751,7 +751,6 @@ add_executable(
     ${CM_STUFF}
 )
 
-
 target_link_libraries(
     visualboyadvance-m
     ${VBAMCORE_LIBS}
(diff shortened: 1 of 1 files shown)


ZachBacon:
@orbea I know why, mxe wraps cmake with a wrapper that points to a cmake script.

orbea:
I'm not sure I understand and am probably missing something? Its being installed directly to the build directory which to me seems strange, maybe this was needed before `visualboyadvance-m` was being installed with an install target?

ZachBacon:
@orbea it's because the cmake conf they use is probably setting their own cmake install directory

PR #534, opened 2019-10-08
cmake: Silence _FORTIFY_SOURCE warnings with debug builds.

orbea:
Silences many noisy warnings with debug builds.
```
[79/196] Building CXX object fex/CMakeFiles/fex.dir/fex/Zlib_Inflater.cpp.o
In file included from ../fex/fex/Zlib_Inflater.cpp:3:
In file included from ../fex/fex/Zlib_Inflater.h:7:
In file included from ../fex/fex/Data_Reader.h:7:
In file included from ../fex/fex/blargg_common.h:8:
In file included from /usr/include/assert.h:35:
/usr/include/features.h:382:4: warning: _FORTIFY_SOURCE requires compiling with optimization (-O) [-W#warnings]
   ^
1 warning generated.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -505,7 +505,11 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID STREQUAL Clang)
     endif()
 
     # common flags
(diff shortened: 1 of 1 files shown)


orbea:
There are probably better ways, but they are much more involved and I think this works for now so that the build output is readable. Redoing the compiler flags can be done later as the cmake project gets cleaned up.

https://cmake.org/cmake/help/latest/variable/CMAKE_BUILD_TYPE.html

PR #553, opened 2019-10-22
Update en.po

LordMagnum:
line corrections


ZachBacon:
Translation updates should be sent to https://www.transifex.com/bgk/vba-m/

rkitover:
It seems `en.po` was not managed by transifex, I am deleting it because there is already an `en_US.po`, `en_GB.po` and `en_BR.po`.

LordMagnum:
Sorry foe this Question, but wat is en_BR <-- ? Me one will German de_DE Translate ^^

rkitover:
I have no idea, I'm guessing it's some kind of brazillian dialect of english.

PR #568, opened 2019-11-21
Joystick handling refactor.

rkitover:
Use the SDL GameController API with SDL events instead of polling
manually. Run SDL_PollEvent() in the panel OnIdle().

The API to the GUI remains the same, the sending of wxSDLJoyEvent
events. Except there are no more hat events, we just use the
GameController buttons. Also the GUI now has to make arrangements for
wxSDLJoy::Poll() to be called periodically, for the config dialog this
is done with a timer.

All Xbox 360 controller buttons and axes are now in the defaults, and
the SDL GameController API will map them to the appropriate keys on
other controllers.

As a consequence of using SDL events, controller attach/detach from the
system is now also handled correctly. It is no longer necessary to have
the controller attached and turned on when the program launches.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,9 +4,7 @@ src/wx/cmd-evtable.h
 src/wx/cmdhandlers.h
 src/wx/cmdtab.cpp
 src/wx/wxvbam.xrs
-build/*
-build32/*
-vsbuild/*
+*build*/*
 cmake-build*
 dependencies/*
 vcpkg/*
--- a/po/wxvbam/wxvbam.pot
+++ b/po/wxvbam/wxvbam.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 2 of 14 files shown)


ZachBacon:
@rkitover looking pretty good so far, felt like there was some minor input lag, but I think it was just in my head as other games I tested worked just fine. Next round of tests will be the actual device polling. Just waiting for my gamepad to naturally turn off before I hit it up with bluetooth.

rkitover:
@denisfa thank you for taking a look!

I hopefully fixed the valgrind errors and I added an osd message for controller attach and detach.

rkitover:
@denisfa I think this is pretty much done and can be merged.

There is an issue with the binary built by my script on msys2 and probably cross-built on linux with mingw. It freezes for a second randomly and the joystick events do not work. I have no idea why that happens, it may be due to static linkage or the way I build some of the libraries, or it may be something else.

This is a separate issue that I'll work on. I think first I'll try making a static visual studio built binary and see if that exhibits the same behavior or not. I was going to add support for this anyway.

PR #595, opened 2020-01-22
Some leftovers.

denisfa:
@rkitover I changed the default audio recording to `.wav`. Is there a reason why when you select the file type on the wxWidgets window, it does not change the filename to its current extension?

If I select `mp3`, I want to not only show the `*.mp3` files on the current dir, but also toggle the `game.wav` to `game.mp3` on the filename.

I want to include a possible fix for this here, so if there is none, I will merge this myself.

### DIFF
--- a/src/common/Patch.cpp
+++ b/src/common/Patch.cpp
@@ -116,7 +116,7 @@ static int64_t readSignVarPtr(FILE* f)
 {
     int64_t offset = readVarPtr(f);
     bool sign =  offset & 1;
-    
+
     offset = offset >> 1;
     if (sign) {
         offset = -offset;
@@ -325,7 +325,7 @@ static bool patchApplyBPS(const char* patchname, uint8_t** rom, int* size)
 
     fseeko64(f, 0, SEEK_SET);
     uint32_t crc = computePatchCRC(f, patchSize - 4);
-    
+
     if (crc != patchCRC) {
         fclose(f);
         return false;
@@ -357,13 +357,13 @@ static bool patchApplyBPS(const char* patchname, uint8_t** rom, int* size)
         fclose(f);
         return false;
     }
-    
+
     uint8_t* new_rom = (uint8_t*)calloc(1, dataSize);
-    
+
     int64_t length = 0;
     uint8_t action = 0;
     uint32_t outputOffset = 0, sourceRelativeOffset = 0, targetRelativeOffset = 0;
-    
+
     while (ftello64(f) < patchSize - 12) {
         length = readVarPtr(f);
         action = length & 3 ;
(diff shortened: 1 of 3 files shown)


rkitover:
Looks fine to me, as far as the code goes.

I don't like the string munging code in cmdevents for this menu item to be honest, although this is not relevant to this PR. I'd do something more elegant with modern C++.

I'll look at this file dialog and see if I can affect this behavior in any way. What I read on forums so far is not promising.

denisfa:
> I'll look at this file dialog and see if I can affect this behavior in any way. What I read on forums so far is not promising.

There seems to be something available starting from `wxWidgets 3.0` from what I read... Do you have a distro in mind when keeping support for `2.8`? For Windows and Mac, we can use static libs starting from `3.0`.

rkitover:
My feeling on 2.8 compat, is that it's fine if we drop it. So far I have been trying to maintain it for people with ancient systems or outdated packages. However, it's probably not worth the effort anymore and we can officially drop support for it, especially if this will make your life easier.

In some cases, a feature can just be compiled out for 2.8 with an `#ifdef`, in other cases code becomes more complex to maintain compatibility,  e.g. using `Connect()` instead of the much much more powerful `Bind()`.

So I'll leave this decision to you and @ZachBacon .

If you do want to maintain 2.8 compat, I can add a travis job to check it and do any necessary fixes.

If you don't, I can make some quick changes to make the code nicer using `3.x` conveniences.

ZachBacon:
At this point, most Linux distro's already are at 3.0 and higher, I think it's safe to drop it now.

rkitover:
@denisfa so I just tested the sound recording save dialog, took me a while because I had to fix the ffmpeg issue on windows.

When I select `.aac` it does change the filename extension for the file being saved to `.aac` and then back to `.mp3` if I select `.mp3` again.

This is testing the visual studio build on windows, I will test on linux as well.

denisfa:
@rkitover For version `3.1`, right? Nevertheless, I will check something else too for Linux.

rkitover:
@denisfa vcpkg uses git wx, which is 3.1, it's a pretty terrible port and I do a lot of stuff in cmake to make it work.

Another thing to keep in mind here, is that wx is going to use the native file picker dialogs for the most part, it doesn't implement its own, and the control you will have over them is going to be limited.

denisfa:
OK, so I did some testing and there is something we can do. I will do on another PR, though, so I can get the WAV recording going to master.

rkitover:
@denisfa so you didn't say what your vote is on 2.8 compat.

denisfa:
@rkitover Sorry, I vote to drop it. I was worried about not being able to build with fedora, since it was installing `2.8` version, but I was able to install the `3.0` one.

rkitover:
@denisfa oh good, I'll update the `installdeps` script for fedora cross builds.

rkitover:
@denisfa I'll also do some cleanup related to dropping 2.8 officially, as promised.

PR #601, opened 2020-02-01
GBA experiments.

denisfa:
@rkitover I need either you or @ZachBacon to do some extra testing here. I have an idea about how to tackle the ROM hacks overrides, but it may not be a good one. My testing was some Pokemon Fire Red hacks (ultra violet, liquid crystal), but there are probably more games that have popular hacks.

### DIFF
--- a/src/wx/panel.cpp
+++ b/src/wx/panel.cpp
@@ -49,11 +49,11 @@ GameArea::GameArea()
 }
 
 
-static void checkOverrides(uint rom_crc32, wxString id)
+static void checkOverrides(unsigned int romCrc32, wxString id)
 {
     wxString crc32;
     wxString identifier;
-    crc32.Printf(wxT("%08X"), rom_crc32);
+    crc32.Printf(wxT("%08X"), romCrc32);
 
     //const char *aux = crc32.mb_str(wxConvUTF8);
     //const char *aux2 = id.mb_str(wxConvUTF8);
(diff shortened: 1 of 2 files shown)


src/wx/panel.cpp
```diff
@@ -48,6 +48,32 @@ GameArea::GameArea()
...
+{
+    wxString crc32;
+    wxString identifier;
+    crc32.Printf(wxT("%08X"), rom_crc32);
+
+    //const char *aux = crc32.mb_str(wxConvUTF8);
+    //const char *aux2 = id.mb_str(wxConvUTF8);
+    //fprintf(stderr, "\n\n%s -- %s\n\n", aux, aux2);
+
+    if (id == wxString("AZLE") && wxString("8E91CD13") != crc32) { // Legend of Zelda, A Link to The Past ROM hacks
```
denisfa: @rkitover This was the issue 484. The index for our `vba-over.ini` are these initials only. If the user set an override, then the original game gives the save "corrupted" messages. If he does not set it, then the hack does not save. The value `8E91CD13` is from the original ROM.

Since the overrides are only for the WX port, I decided to create this function here.

denisfa:
Just realized that I pushed that crap where I tagged the wrong issue. Sorry about this...

denisfa:
As per discussion, I need to figure a better way to do this, although I should keep in mind the crc32 stuff.

I will open a new pr when I have something.

PR #608, opened 2020-02-14
Fix: a issue with loading e-Reader dot code

s520:
This PR fix a issue with loading e-Reader dot code.

Specifically, it is as follows.
- Even if Dot code file is not specified, it tries to read as eternal and does not accept button input.
- In Japanese etc., the path of Dot code file is garbled internally.

### DIFF
--- a/src/gba/GBA.cpp
+++ b/src/gba/GBA.cpp
@@ -1712,9 +1712,19 @@ const char* GetSaveDotCodeFile()
     return saveDotCodeFile;
 }
 
+void ResetLoadDotCodeFile()
+{
+    loadDotCodeFile = strdup("");
+}
+
 void SetLoadDotCodeFile(const char* szFile)
 {
     loadDotCodeFile = strdup(szFile);
+}
+
+void ResetSaveDotCodeFile()
+{
+    saveDotCodeFile = strdup("");
 }
 
 void SetSaveDotCodeFile(const char* szFile)
--- a/src/gba/GBA.cpp
+++ b/src/gba/GBA.cpp
@@ -1714,6 +1714,11 @@ const char* GetSaveDotCodeFile()
 
 void ResetLoadDotCodeFile()
 {
+	if(loadDotCodeFile)
+	{
+        free((char*)loadDotCodeFile);
+    }
+
     loadDotCodeFile = strdup("");
 }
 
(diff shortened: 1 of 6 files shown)


denisfa:
@s520 Thank you very much for your contribution! I have some questions, though. I will put them on the code.

src/gba/GBA.cpp
```diff
@@ -1712,11 +1712,21 @@ const char* GetSaveDotCodeFile()
     return saveDotCodeFile;
 }
 
+void ResetLoadDotCodeFile()
+{
+    loadDotCodeFile = strdup("");
```
denisfa: @s520 Is this leaking memory? I believe you should `if (loadDotCodeFile) free(loadDotCodeFile)` before the `strdup`.

src/gba/GBA.cpp
```diff
@@ -1712,11 +1712,21 @@ const char* GetSaveDotCodeFile()
...
+}
+
 void SetLoadDotCodeFile(const char* szFile)
 {
     loadDotCodeFile = strdup(szFile);
 }
 
+void ResetSaveDotCodeFile()
+{
+    saveDotCodeFile = strdup("");
```
denisfa: Likewise.

src/wx/cmdevents.cpp
```diff
@@ -869,7 +878,7 @@ EVT_HANDLER_MASK(SaveDotCodeFile, "Save e-Reader Dot Code...", CMDEN_GBA)
         return;
 
     savedotcodefile_path = dlg.GetPath();
-    SetSaveDotCodeFile(savedotcodefile_path.mb_str(wxConvUTF8));
+    SetSaveDotCodeFile(savedotcodefile_path.mb_str());
```
denisfa: @s520 Just for clarification, were the issues regarding the path being garbled on Windows?

s520: Yes. I confirmed it in Japanese environment of Windows 10 64bit.
I don't know wxWidgets in detail, but if you specify wxConvUTF8 as an argument, this part will be garbled if the path contains Japanese (multibyte characters?).

src/gba/GBA.cpp
s520: Oops, that's right.
I will fix it.

s520: Please check that I have fixed it.

src/gba/GBA.cpp
s520: I have also fixed this, so please check it.

src/wx/cmdevents.cpp
denisfa: Thank you! We have some bugs about converting between UTF16 and UTF8 for windows, so I just wanted to confirm this was part of it. Nevertheless, this should be enough for now.

PR #614, opened 2020-02-18
Various Fixes

denisfa:
@rkitover I think I found a way to solve the modal issue for Windows. Check it out.

I am going to do a double check on all of this, but I also ask for your help to test it and review the overall design of the changes (especially the modal solution).

I have a slight suspicion about some of our debugging functions, but I have to go on a little more before making it work. I should be able to give some feedback this weekend.

While working on this, I had a light about our current throttle issues. Try using the `2.1.1` version and see if you can actually have a speedup.

There is an LCD mode filter for GBA that I made work like a toggle. So you have to enable or disable it. I really like to give the user this type of power, although someone will probably complain...

I will update this comment with extra information that I will get from more testing. It will be commit by commit.

### DIFF
--- a/src/common/ConfigManager.h
+++ b/src/common/ConfigManager.h
@@ -139,7 +139,6 @@ extern int windowMaximized;
 extern int windowPositionX, bkpPosX;
 extern int windowPositionY, bkpPosY;
 extern uint32_t windowWidth;
-extern int winFlashSize;
 extern int winGbBorderOn;
 extern int winGbPrinterEnabled;
 extern int winPauseNextFrame;
--- a/src/common/ConfigManager.h
+++ b/src/common/ConfigManager.h
@@ -25,6 +25,8 @@ extern bool mirroringEnable;
 extern bool parseDebug;
 extern bool speedHack;
 extern bool speedup;
+extern bool gbaLcdFilter;
+extern bool gbLcdFilter;
 extern char *rewindMemory;
 extern const char *aviRecordDir;
 extern const char *biosFileNameGB;
--- a/src/wx/gfxviewers.cpp
+++ b/src/wx/gfxviewers.cpp
@@ -1582,6 +1582,44 @@ public:
         }
     }
 
+    void SaveGBATile(wxCommandEvent& ev)
+    {
+        (void)ev; // unused params
+        GameArea* panel = wxGetApp().frame->GetPanel();
+        wxString bmp_save_dir = wxGetApp().frame->GetGamePath(gopts.scrshot_dir);
+        // no attempt is made here to translate the dialog type name
+        // it's just a suggested name, anyway
+        wxString def_name = panel->game_name() + wxT('-') + dname;
+        def_name.resize(def_name.size() - 6); // strlen("Viewer")
+
+        if (captureFormat)
+            def_name += wxT(".bmp");
+        else
+            def_name += wxT(".png");
+
+        wxFileDialog dlg(GetGrandParent(), _("Select output file"), bmp_save_dir, def_name,
+            _("PNG images|*.png|BMP images|*.bmp"), wxFD_SAVE | wxFD_OVERWRITE_PROMPT);
+        dlg.SetFilterIndex(captureFormat);
+        int ret = dlg.ShowModal();
+        bmp_save_dir = dlg.GetDirectory();
+
+        if (ret != wxID_OK)
+            return;
+
+        wxBitmap obmp = gv->bm->GetSubBitmap(wxRect(0, 0, gv->bmw, gv->bmh));
+        wxString fn = dlg.GetPath();
+        wxBitmapType fmt = dlg.GetFilterIndex() ? wxBITMAP_TYPE_BMP : wxBITMAP_TYPE_PNG;
+
+        if (fn.size() > 4) {
+            if (wxString(fn.substr(fn.size() - 4)).IsSameAs(wxT(".bmp"), false))
+                fmt = wxBITMAP_TYPE_BMP;
+            else if (wxString(fn.substr(fn.size() - 4)).IsSameAs(wxT(".png"), false))
+                fmt = wxBITMAP_TYPE_PNG;
+        }
+
+        obmp.SaveFile(fn, fmt);
+    }
+
 protected:
     int charbase, is256, palette;
     wxControl *tileno, *addr;
@@ -1591,6 +1629,7 @@ protected:
 };
 
 BEGIN_EVENT_TABLE(TileViewer, GfxViewer)
+EVT_BUTTON(XRCID("SaveGBATile"), TileViewer::SaveGBATile)
 EVT_GFX_CLICK(wxID_ANY, TileViewer::UpdateMouseInfoEv)
 END_EVENT_TABLE()
 
@@ -1685,6 +1724,44 @@ public:
         }
     }
 
+    void SaveGBTile(wxCommandEvent& ev)
+    {
+        (void)ev; // unused params
+        GameArea* panel = wxGetApp().frame->GetPanel();
+        wxString bmp_save_dir = wxGetApp().frame->GetGamePath(gopts.scrshot_dir);
+        // no attempt is made here to translate the dialog type name
+        // it's just a suggested name, anyway
+        wxString def_name = panel->game_name() + wxT('-') + dname;
+        def_name.resize(def_name.size() - 6); // strlen("Viewer")
+
+        if (captureFormat)
+            def_name += wxT(".bmp");
+        else
+            def_name += wxT(".png");
+
+        wxFileDialog dlg(GetGrandParent(), _("Select output file"), bmp_save_dir, def_name,
+            _("PNG images|*.png|BMP images|*.bmp"), wxFD_SAVE | wxFD_OVERWRITE_PROMPT);
+        dlg.SetFilterIndex(captureFormat);
+        int ret = dlg.ShowModal();
+        bmp_save_dir = dlg.GetDirectory();
+
+        if (ret != wxID_OK)
+            return;
+
+        wxBitmap obmp = gv->bm->GetSubBitmap(wxRect(0, 0, gv->bmw, gv->bmh));
+        wxString fn = dlg.GetPath();
+        wxBitmapType fmt = dlg.GetFilterIndex() ? wxBITMAP_TYPE_BMP : wxBITMAP_TYPE_PNG;
+
+        if (fn.size() > 4) {
+            if (wxString(fn.substr(fn.size() - 4)).IsSameAs(wxT(".bmp"), false))
+                fmt = wxBITMAP_TYPE_BMP;
+            else if (wxString(fn.substr(fn.size() - 4)).IsSameAs(wxT(".png"), false))
+                fmt = wxBITMAP_TYPE_PNG;
+        }
+
+        obmp.SaveFile(fn, fmt);
+    }
+
 protected:
     int bank, charbase, palette;
     wxControl *addr, *tileno;
@@ -1694,6 +1771,7 @@ protected:
 };
 
 BEGIN_EVENT_TABLE(GBTileViewer, GfxViewer)
+EVT_BUTTON(XRCID("SaveGBTile"), GBTileViewer::SaveGBTile)
 EVT_GFX_CLICK(wxID_ANY, GBTileViewer::UpdateMouseInfoEv)
 END_EVENT_TABLE()
 }
--- a/src/wx/opts.cpp
+++ b/src/wx/opts.cpp
@@ -208,6 +208,8 @@ opt_desc opts[] = {
 
     /// GB
     STROPT("GB/BiosFile", "", wxTRANSLATE("BIOS file to use for GB, if enabled"), gopts.gb_bios),
+    INTOPT("GB/ColorOption", "GBColorOption", wxTRANSLATE("GB color enhancement, if enabled"), gbColorOption, 0, 1),
+    BOOLOPT("GB/LCDFilter", "GBLcdFilter", wxTRANSLATE("Apply LCD filter, if enabled"), gbLcdFilter),
     STROPT("GB/GBCBiosFile", "", wxTRANSLATE("BIOS file to use for GBC, if enabled"), gopts.gbc_bios),
     NOOPT(wxT("GB/Palette0"), "", wxTRANSLATE("The default palette, as 8 comma-separated 4-digit hex integers (rgb555).")),
     NOOPT(wxT("GB/Palette1"), "", wxTRANSLATE("The first user palette, as 8 comma-separated 4-digit hex integers (rgb555).")),
@@ -219,6 +221,7 @@ opt_desc opts[] = {
 
     /// GBA
     STROPT("GBA/BiosFile", "", wxTRANSLATE("BIOS file to use, if enabled"), gopts.gba_bios),
+    BOOLOPT("GBA/LCDFilter", "GBALcdFilter", wxTRANSLATE("Apply LCD filter, if enabled"), gbaLcdFilter),
 #ifndef NO_LINK
     BOOLOPT("GBA/LinkAuto", "LinkAuto", wxTRANSLATE("Enable link at boot"), gopts.link_auto),
     INTOPT("GBA/LinkFast", "SpeedOn", wxTRANSLATE("Enable faster network protocol by default"), linkHacks, 0, 1),
--- a/src/wx/opts.cpp
+++ b/src/wx/opts.cpp
@@ -569,7 +569,7 @@ void load_opts()
             cfg->Read(opt.opt, &opt.curint, *opt.intopt);
 
             if (opt.curint < opt.min || opt.curint > opt.max) {
-                wxLogWarning(_("Invalid value %d for option %s; valid values are %d - %d"), opt.curint, opt.opt.c_str(), opt.min, opt.max);
+                wxLogWarning(_("Invalid value %d for option %s; valid values are %d - %d"), opt.curint, opt.opt.c_str(), int(opt.min), int(opt.max));
             } else
                 *opt.intopt = opt.curint;
         } else if (opt.doubleopt) {
--- a/src/wx/panel.cpp
+++ b/src/wx/panel.cpp
@@ -1078,12 +1078,14 @@ void GameArea::OnIdle(wxIdleEvent& event)
 
         // set focus to panel
         w->SetFocus();
+
+        // generate system color maps (after output module init)
+        if (loaded == IMAGE_GBA) utilUpdateSystemColorMaps(gbaLcdFilter);
+        else if (loaded == IMAGE_GB) utilUpdateSystemColorMaps(gbLcdFilter);
+        else utilUpdateSystemColorMaps(false);
     }
 
     mf->PollJoysticks();
-
-    if (loaded == IMAGE_GBA) utilUpdateSystemColorMaps(gbaLcdFilter);
-    else if (loaded == IMAGE_GB) utilUpdateSystemColorMaps(gbLcdFilter);
 
     if (!paused) {
         HidePointer();
@@ -1471,9 +1473,6 @@ DrawingPanelBase::DrawingPanelBase(int _width, int _height)
         systemBlueShift = 0;
         RGB_LOW_BITS_MASK = 0x0421;
     }
-
-    // needs to be run once before the first draw
-    utilUpdateSystemColorMaps(false);
 }
 
 DrawingPanel::DrawingPanel(wxWindow* parent, int _width, int _height)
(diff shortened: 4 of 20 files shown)


src/wx/gfxviewers.cpp
```diff
@@ -1685,6 +1724,44 @@ class GBTileViewer : public GfxViewer {
...
+        wxFileDialog dlg(GetGrandParent(), _("Select output file"), bmp_save_dir, def_name,
+            _("PNG images|*.png|BMP images|*.bmp"), wxFD_SAVE | wxFD_OVERWRITE_PROMPT);
+        dlg.SetFilterIndex(captureFormat);
+        int ret = dlg.ShowModal();
+        bmp_save_dir = dlg.GetDirectory();
+
+        if (ret != wxID_OK)
+            return;
+
+        wxBitmap obmp = gv->bm->GetSubBitmap(wxRect(0, 0, gv->bmh, gv->bmh));
```
denisfa: It should be `gv->bmw, gv->bmh` instead of `gv->bmh, gv->bmh`.

src/wx/gfxviewers.cpp
```diff
@@ -1582,6 +1582,44 @@ class TileViewer : public GfxViewer {
...
```
denisfa: It should be `gv->bmw, gv->bmh` instead of `gv->bmh, gv->bmh`.

rkitover:
@denisfa this is an amazing amount of work. I will read the diff in a minute, do you need any help with sound drivers?

src/wx/panel.cpp
```diff
@@ -1080,7 +1084,9 @@ void GameArea::OnIdle(wxIdleEvent& event)
 
     if (!paused) {
         HidePointer();
+#ifndef __WXMSW__
         event.RequestMore();
+#endif // __WXMSW__
```
rkitover: Actually use a dummy `OnIdle` event value, and this will just do nothing.

src/wx/wxvbam.cpp
```diff
@@ -1292,3 +1296,21 @@ int wxvbamApp::FilterEvent(wxEvent& event)
...
+}
+
+void RenderTimer::Notify()
+{
+    panel->OnIdle();
+}
+
+void RenderTimer::start()
+{
+    wxTimer::Start(5);
```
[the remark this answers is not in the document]
rkitover: We can even adjust this value based on some kind of Windows performance index API. To support people's potato laptops, We get **ALOT** of those.

src/common/ConfigManager.h
```diff
@@ -25,6 +25,7 @@ extern bool mirroringEnable;
 extern bool parseDebug;
 extern bool speedHack;
 extern bool speedup;
+extern bool gbaLcdFilter;
```
rkitover: Yes, looking after the good old `ConfigManager`, I do that constantly. We need to rip it out and unify the wx config system.

I am **NOT** asking you to do that here. I'm just saying eventually we will need to do this.

src/wx/gfxviewers.cpp
rkitover: When you do a `rebase -i` you can stop to amend a commit.

src/wx/gfxviewers.cpp
```diff
@@ -1582,6 +1582,44 @@ class TileViewer : public GfxViewer {
...
+        wxString bmp_save_dir = wxGetApp().frame->GetGamePath(gopts.scrshot_dir);
+        // no attempt is made here to translate the dialog type name
+        // it's just a suggested name, anyway
+        wxString def_name = panel->game_name() + wxT('-') + dname;
+        def_name.resize(def_name.size() - 6); // strlen("Viewer")
+
+        if (captureFormat)
+            def_name += wxT(".bmp");
+        else
+            def_name += wxT(".png");
```
rkitover: Only tangentially related to this, because wx has its own libpng dependency, but we need to replace whatever we use libpng for with stb_image. At some point. To remove one direct dependency of our own.

src/wx/opts.cpp
```diff
@@ -564,7 +566,7 @@ void load_opts()
             cfg->Read(opt.opt, &opt.curint, *opt.intopt);
 
             if (opt.curint < opt.min || opt.curint > opt.max) {
-                wxLogWarning(_("Invalid value %d for option %s; valid values are %d - %d"), opt.curint, opt.opt.c_str(), opt.min, opt.max);
+                wxLogWarning(_("Invalid value %d for option %s; valid values are %d - %d"), opt.curint, opt.opt.c_str(), int(opt.min), int(opt.max));
```
rkitover: Using modern C++ casts, +1

src/wx/wxvbam.cpp
```diff
@@ -1110,10 +1114,10 @@ void MainFrame::MenuPopped(wxMenuEvent& evt)
     if (popped)
         panel->ShowPointer();
 
-    if (menus_opened)
-        panel->Pause();
-    else if (!IsPaused())
-        panel->Resume();
+    //if (menus_opened)
```
rkitover: Just delete this. No one ever liked it.

src/wx/wxvbam.cpp
```diff
@@ -1155,7 +1159,7 @@ void MainFrame::StartModal()
     // pointer when dialog popped up
     // it will auto-hide again once game resumes
     panel->ShowPointer();
-    panel->Pause();
+    //panel->Pause();
```
rkitover: Likewise.

src/wx/xrc/MainMenu.xrc
```diff
@@ -508,13 +508,21 @@
...
+        </object>
       </object>
       <object class="wxMenu">
         <label>_Game Boy</label>
         <object class="wxMenuItem" name="GameBoyConfigure">
           <label>Configure ...</label>
         </object>
         <object class="separator"/>
+        <object class="wxMenuItem" name="GBColorOption">
+          <label>_GB color option</label>
```
rkitover: You have the accelerator `_G` already being used above, choose a different letter.

rkitover: See this:

https://docs.wxwidgets.org/3.0/overview_xrcformat.html#overview_xrcformat_type_text

src/wx/xrc/OAMViewer.xrc
```diff
@@ -141,7 +141,7 @@
               </object>
               <object class="sizeritem">
                 <object class="unknown" name="Color"/>
-                <flag>wxALL|wxALIGN_BOTTOM</flag>
```
rkitover: Thank you so much for fixing this shit. I forgot where I saw it last.

rkitover:
I rebased on master and added a commit to update the gettext pot, a feature I just added like a day or two ago.

We will do the transifex push once it's merged.

rkitover:
To update when multiple people are working on a git branch, keep a couple of things in mind:

- Always `push -f` unless you're adding a commit on top. And it's almost always better to edit the history than to add more commits. Never add commits fixing previous commits, only improving or adding to them. In this case, I added the gettext commit on top because that is a reasonable thing to do in this case. I could have of course amended each option commit, but that would make the history worse not better.

- To update when someone else (very rudely you might say) did a `push -f`, `pull --rebase` will **USUALLY** work. Verify the log, and if necessary do this instead:

```bash
git status # should be clean, with your work having been already pushed
git fetch --all --prune
git reset --hard origin/<branch-name>
```

I will add this to the manual.

src/wx/wxvbam.cpp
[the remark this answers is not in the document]
rkitover: What I meant to say the frequency should be with the emulator active, should depend on the desired framerate.

That is, it should be the time between frames minus `emu->main` and the rendering.

The desired framerate would be calculated by the normal hardware framerate (GBA or GB) multiplied by the throttle.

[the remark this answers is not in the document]
rkitover: Also just realized, in the `throttle=0` (full speed) or `speedup` case, we would need to just loop inside `OnIdle` and not leave the timer until the setting is changed.

rkitover:
The windows dialog fix for link is going to be one of the best features of the next release.

denisfa:
@rkitover Thank you for the initial feedback! I really appreciate it! I don't want to go MIA for no reason, so I will need some more days before going further into this. I am currently checking some other solutions. I don't really like the timer solution, because it will screw our already bad throttle issue.

> Always push -f unless you're adding a commit on top. And it's almost always better to edit the history than to add more commits.

Should I do a `git rebase -i origin/master` when ammending my commits?

src/wx/xrc/MainMenu.xrc
denisfa: So this is the reason of these `_` before some letters... I will check it out.

src/wx/opts.cpp
denisfa: I know this is stupid, but this was giving a warning->segfault message on wxWidgets `3.0`->`3.1` on Windows 10. That is the reason here =P

src/wx/gfxviewers.cpp
denisfa: Where are we using it?

src/wx/panel.cpp
denisfa: How can I create a dummy one? I mean preferably without compiler warning messages.

src/wx/wxvbam.cpp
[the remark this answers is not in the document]
denisfa: Actually, this has a major drawback. This breaks my current throttle setup. As I cannot get it to work, even using the `Simple` output module. I am still investigating this.

Keep in mind that the slowest action here is the drawing. Not much we can do without checking vsync and other stuff. I know nothing about Windows. Have you ever seen an API for this before?

denisfa:
> this is an amazing amount of work. I will read the diff in a minute, do you need any help with sound drivers?

I don't think I need anything from the sound drivers yet, thank you.

We really need to solve the throttle issue here. It is driving me nuts. Were you able to get either turbo or throttle working with a previous version of vbam with your windows/linux? Using the vsync disable trick or others?

rkitover:
> Should I do a `git rebase -i origin/master` when amending my commits?

Yeah, pretty much:

```bash
git fetch --all --prune
git rebase -i origin/master
```

I will add a note to the manual.

BTW @ZachBacon  got the main site back on the VPS I made for him on my dedi, and he is doing a phpbb conversion of the board which will hopefully have markdown support, and that will run there as well.

rkitover:
> We really need to solve the throttle issue here. It is driving me nuts. Were you able to get either turbo or throttle working with a previous version of vbam with your windows/linux? Using the vsync disable trick or others?

On those installations it is not worth the trouble to try to fix. Considering I have an installation where it works. I will look into the vsync APIs too and redo that dialog as promised.

rkitover:
> We really need to solve the throttle issue here. It is driving me nuts.

Oh, another reason I mentioned the sound drivers.

To some degree, the speed of the game is controlled by the current sound driver.

The sound speed is the game speed and the sound driver forces this synchronization.

That's part of how throttle works.

I think I told you this before, but in any case just throwing it out there in case it helps you with your testing.

denisfa:
> The sound speed is the game speed and the sound driver forces this synchronization.

@rkitover Yes, you have told me this  before.

What I am thinking is to do some workaround if everything else fails.

Let's say I have a 60FPS with 100% throttle (openGL). Due to the vsync issue, I cannot put 200% throttle with 120FPS. What I could do in this case is to setup a 1 frame skip for 200%; it yields to 60FPS (skipping 1 frame after every draw) and 200% throttle. Same thing for any throttle percentage (300% -> 2 frame skip, 400% -> 3 frame skip etc). Same method for turbo/speedup.

This could be a good idea, but I am getting the behaviour that this is a regression. When I use the simple module, I get 120FPS and 200% throttle. I am not fixing anything here.

rkitover:
Sure! And that's why I made that dialog in the first place, so you could use the method that works for you. It just needs better instructions and a reordering.

The behavior of course is not exactly the same, with throttle you get speeded up sound instead of dropping sound, the video is smoother, etc..

denisfa:
> The behavior of course is not exactly the same, with throttle you get speeded up sound instead of dropping sound, the video is smoother, etc..

What do you mean here? The frame skip, at least from what I did see on the code, is only for video frames. Or it is also for audio ones?

On this topic, I finally discovered how to do the video recording when frame skipping. I will include in this pr.

rkitover:
> What do you mean here? The frame skip, at least from what I did see on the code, is only for video frames. Or it is also for audio ones?

Well, with throttle at least the actual audio frequency changes, and it outputs the sound at a faster frequency. In other modes the extra audio data is just dropped. At least from what I remember when working on the sound drivers last.

rkitover:
Also just noticed that in the video config, you can set a frame skip that always applies, this actually helps a lot on broken configurations like my windows VM one. We should document this.

rkitover:
By the way, on windows are you getting an error about deleting a temp file on app close with master or this branch?

rkitover:
Also I think we should make an alternative to winsparkle that is cross platform and properly links as a static library. Perhaps by extending the mac sparkle.

src/wx/opts.cpp
```diff
@@ -219,6 +219,7 @@ opt_desc opts[] = {
 
     /// GBA
     STROPT("GBA/BiosFile", "", wxTRANSLATE("BIOS file to use, if enabled"), gopts.gba_bios),
+    BOOLOPT("GBA/LCDFilter", "GBALcdFilter", wxTRANSLATE("Apply LCD filter, if enabled"), gbaLcdFilter),
```
negativeExponent: might as well make this a global option instead of just exclusive on GBA since utilUpdateSystemColorMaps() affects gba and gb/gbc colormaps or just give each system its own menu item so one can enable on GBA and disable on GB got e.g. for users who want some effect combinations.

denisfa:
> By the way, on windows are you getting an error about deleting a temp file on app close with master or this branch?

I can confirm this error. But it is random for me, sometimes I get it and other times I do not.

src/wx/opts.cpp
denisfa: It already is global, I just added it exclusively to the GBA menu part. If you toggle it while playing a GB, it will work on the GB game. Would you still want it in both menus partitions?

negativeExponent: it should be on a global menu if it was intended to be global, not on the GBA menu and apply it globally just do no make sense.

also whatever value is set should apply directly at startup 
https://github.com/visualboyadvance-m/visualboyadvance-m/blob/master/src/wx/panel.cpp#L1470

rkitover:
The tempfile error has been fixed in 0e8e46f0, this branch has been rebased on top of that.

rkitover:
Another issue with winsparkle is that it considers git builds newer than a tag to be older than that tag, we need to fix that.

rkitover:
Verified that the winsparkle temp file fix works on xp.

src/wx/opts.cpp
ZachBacon: I'm going to agree with @negativeExponent on this one, makes more logical sense to do that instead of the gba menu. But that's my 2 cents on the matter.

rkitover:
@ZachBacon @denisfa I merged the dev manual. There are some brief sections on our process, please let me know if you disagree with anything or anything needs to be clarified.

denisfa:
Ok, I am going to wrap this up. I will do some more changes based on the reviews and clean for the merge.

@rkitover That does not seem very complicated. I will test some ideas to confirm if it is indeed easy to fix.

rkitover:
Are you still going with the timer solution? If you don't mind I'll do some testing of things like throttle when you're ready to see if they are affected.

denisfa:
@rkitover For now, yes. I have a threading solution on the sideline, but it is really unstable. If you can, please wait a little bit for me to do some last coding regarding this. Feel free to add feedback about throttle issues, this way we can check what works and whatnot.

rkitover:
@denisfa I was going to ask, can you make the timer based event loop a cmake option, and just have it default to `ON` on `WIN32`. Because I need to test this on linux, and it could also be useful.

rkitover:
I could of course do that myself when you're done with the changes too, it's pretty simple, just setting a different `add_definition()` to use as a preprocessor symbol to test.

denisfa:
@rkitover Absolutely, I will add a "ENABLE_TIME_LOOP" variable. Although I did not plan to replace the current method for Linux/MacOS. We will still have to talk about this, since I also have some results to report about using the timer.

Please do not add the `turbo-enhancement` changes here, at least for now. I am doing some commit rewriting and this can make things very complicated on my end.

rkitover:
Oh I would never do that, that is not done yet and will go through a PR.

I actually wrote about this in the dev manual and of course I will do my
best to follow it myself.

The stuff I rebased your branch on was just minor follow ups to process
related things I was working on the last few days, and all of that is now
taken care of.

>

src/wx/gfxviewers.cpp
rkitover: Apparently `utilWritePNGFile` in `src/Util.cpp`, a few things use that.

src/wx/panel.cpp
rkitover: Just something like this:

```cpp
wxIdleEvent dummy;
OnIdle(dummy);
```

will work fine.

src/wx/opts.cpp
negativeExponent: not everyone played on gbasp, which is where this lcd saturation was based upon, causes too much saturation for some titles

denisfa:
@rkitover Check lat commits. I enabled the timer as a cmake option, but it is very buggy on Linux. If you want to test something, then this is the time for it.

Unless something new appears, I will create the pr to be merged. I believe some of the issues described here will need another PR to properly tackle them.

rkitover:
Nice, I will play with this on my working linux throttle config.

Just a note that libretro is not actually useless, we are the upstream for the vba-m libretro core and we pull changes from them.

The SDL vbam binary is more useless, but still occasionally useful to some people so we do minimal maintenance on it. It does need more love.

negativeExponent:
nice PR label.

rkitover:
@denisfa BTW please see my responses to some of the code comment threads.

denisfa:
What are you talking about? It is literally an empty function definition for both. It is useless. 

Or are you referring to poor choice of words in the commit message? Like libretro/vbam are useless? These commits will be squashed with 86948eb as soon as I rebase some more stuff I am doing, so no need to overreact.

src/wx/gfxviewers.cpp
denisfa: I will move this topic to #622. I will close this here.

rkitover:
Yeah I'm sorry, I misunderstood, I thought you were saying that libretro is useless, and I was not taking offense at that, I just misunderstood what you meant.

negativeExponent:
https://github.com/visualboyadvance-m/visualboyadvance-m/pull/614/files#diff-15937defa9f38aa6788aa03d6c127f7eR1081-R1082

this is totally unnecessary and performace degrading to keep regeneration systemColorMapxx on the main loop. its totally fine to ignore my suggestion to this matter but just do not make any sense

denisfa:
There are more stuff to fix. This is also intermediary work, since there are some changes that I want to do. Just taking some time researching.

rkitover:
I will do some testing of the timer loop on my laptop linux configuration where throttle works.

denisfa:
@rkitover There has been some time here, I would like to finally merge this. I want some of these to be ready for next nightly. I still need to work more on the timer solution (especially for linux), but I believe this can be done later.

src/wx/wxvbam.cpp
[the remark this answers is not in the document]
denisfa: I can start solving this by chaining the timer events at the end of the handler. This way we don't accumulate wxEvents on our queue.

> Also just realized, in the throttle=0 (full speed) or speedup case, we would need to just loop inside OnIdle and not leave the timer until the setting is changed.

We can loop inside the function for a given time. What is the best library for this? `chrono`?

rkitover:
I have no objections to merging this now, as long as we disable the TIMER_LOOP default on Windows until we do some more testing. I know I promised to do more testing myself, and I will try to get to doing some of that today.

The other option I can offer is tweaking my nightly script to make a timer-loop binary for people to test, this way we have both versions until we figure out all the issues.

Let me know what you would like to do.

denisfa:
> I have no objections to merging this now, as long as we disable the TIMER_LOOP default on Windows until we do some more testing. I know I promised to do more testing myself, and I will try to get to doing some of that today.

I partially agree. I have this setup because it is more stable than my other solution (threading for the main loop). I would really like to some users to test this for Windows.

> The other option I can offer is tweaking my nightly script to make a timer-loop binary for people to test, this way we have both versions until we figure out all the issues.

> Let me know what you would like to do.

We could get everything from here except the TIMER commits and move to another PR. Just like we did some time ago.

rkitover:
This is done. Separate pr has been created for the timer-loop branch.

Your chain calling change addresses some of my initial concerns.

I want to check CPU usage however and make sure we are not just busy looping always.

And I will also make sure turbo related stuff works, such as it is.

The nightly runs at 23:00 UTC, I just missed it by two hours, I'll just run it manually.

Stay safe.

PR #619, opened 2020-02-27
Force Adwaita:dark on GTK, disable vsync.

rkitover:
Add shell script wrapper on unix to set environment variables for GTK 2
and 3+ theme to Adwaita:dark. This fixes various Gnome etc. theme
incompatibilities that can make the app unusable.

Also set environment variables to turn off vsync on both mesa and Nvidia
drivers.

@ZachBacon your thoughts please, based on your packaging expertise.

@denisfa this is at least a minor improvement for the vsync issue, right?

### DIFF
--- a/src/wx/linux-shell-wrapper.sh
+++ b/src/wx/linux-shell-wrapper.sh
@@ -1,0 +1,24 @@
+#!/bin/sh
+
+# Force Adwaita-dark theme for GTK2 and 3+, other themes can cause serious
+# issues:
+
+# For GTK2:
+if [ -e /usr/share/themes/Adwaita-dark/gtk-2.0/gtkrc ]; then
+	GTK2_RC_FILES=/usr/share/themes/Adwaita-dark/gtk-2.0/gtkrc
+fi
+
+# For GTK3+:
+GTK_THEME=Adwaita:dark
+
+# Turn off vsync:
+
+# for mesa drivers.
+vblank_mode=0
+
+# for Nvidia drivers.
+__GL_SYNC_TO_VBLANK=0
+
+export GTK2_RC_FILES GTK_THEME vblank_mode __GL_SYNC_TO_VBLANK
+
+${0%/*}/visualboyadvance-m.real "$@"
--- a/src/wx/CMakeLists.txt
+++ b/src/wx/CMakeLists.txt
@@ -1115,6 +1115,19 @@ if(APPLE)
     endif()
 endif()
 
+if(NOT WIN32 AND NOT APPLE)
+        add_custom_command(
+            TARGET visualboyadvance-m
+            POST_BUILD
+            COMMAND ${CMAKE_COMMAND} -E copy   ${CMAKE_BINARY_DIR}/visualboyadvance-m ${CMAKE_BINARY_DIR}/visualboyadvance-m.real
+            COMMAND ${CMAKE_COMMAND} -E remove ${CMAKE_BINARY_DIR}/visualboyadvance-m
+            COMMAND ${CMAKE_COMMAND} -E copy   ${CMAKE_CURRENT_SOURCE_DIR}/linux-shell-wrapper.sh ${CMAKE_BINARY_DIR}/visualboyadvance-m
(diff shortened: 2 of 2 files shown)


ZachBacon:
I for one can't approve of forcing a theme on a user. But that's my 2 cents. And as for disabling vsync, we should do that in code not as a shell script.

src/wx/linux-shell-wrapper.sh
```diff
@@ -0,0 +1,24 @@
...
+	GTK2_RC_FILES=/usr/share/themes/Adwaita-dark/gtk-2.0/gtkrc
+fi
+
+# For GTK3+:
+GTK_THEME=Adwaita:dark
+
+# Turn off vsync:
+
+# for mesa drivers.
+vblank_mode=0
```
denisfa: This is ignoring our vsync menu option. Can you check if `vblank_mode=1` works on your config (since I have nvidia proprietary drivers).

It goes like this, if i remember correctly: `vblank_mode=`
```
0 - no vsync.

1 - default no vsync, but obey application's choice.

2 - vsync on, but obey application's choice. <== Here is our issue

3 - vsync forced on, but app can choose when to swap buffers(?).
```

denisfa:
@ZachBacon No clear idea about Windows yet, but on Linux the driver may lock this type of config away from the app. Hence you need to call with one of these variables set. You don't need to aggro the system, since you can just

`$ __GL_SYNC_TO_VBLANK=0 visualboyadvance-m`

like any variable to be set to that specific command. All this GTK stuff can be set like this too, no?

ZachBacon:
@denisfa the vsync option I know can be done the way you describe, but I still rather opt for a coded option and as for themes, I don't agree with taking away the option for a user to have a theme of their choice.

denisfa:
@ZachBacon For vsync, there are some workarounds... I discussed somewhere but it goes like this:

Let's say I have `60FPS` (locked by drivers/config) max for OpenGL with `100% throttle`.

If I want to set `200% throttle`, then I include `1` frame skip. So we get `120FPS`, but skip showing half of the frames. Therefore we will still be `60 FPS` for OpenGL to draw. 

If it is `300% throttle`, then I use `2` frames skip and follow the same idea as above. And so on for bigger throttle.

It will limit the options for throttle (multiples of 100), but then we have a stable way of getting it. If the user ticks our `vsync` menu option, then we set our frame skip to `0`. So it will be up to his drivers and configuration to figure it out how many frames it can show per second. This should be enough for a sane solution. This is obviously for Linux, since I have to test a lot of stuff for Windows and MacOS.

rkitover:
@ZachBacon The reason I did the theme stuff as a shell script wrapper was because on my opensuse laptop installation, Gnome completely fucked up the app on a GTK2 wx3. Like the none of the menu items are visible without being selected.

Now, the binary is still installed, and the user can just run `visualboyadvance-m.real` to get just the binary without any theme setting variables.

I am open to other alternatives, but I still need to fix the issue with the app being completely unusable in my Gnome for whatever reason. We have to consider that our target audience is often not savvy enough to do things with GTK themes, even though in theory forcing a theme is not the nicest solution. But for users who want to use their own theme, they can just use non-wrapped binary.

@denisfa @ZachBacon As for the vsync issue, I would rather we just set the variables and remove the vsync option from the menu. If anyone complains, we could come up with something, but it seems to me that very few if any people would actually want vsync for an app like this. Also disabling it with these variables is far more reliable than having the app request it.

@denisfa About your throttle implementation idea, my Speedup/Turbo dialog already has mutually exclusive frame skip and throttle settings, and the user may want and select actual throttle, because the behavior is not exactly the same. However, if we determine that the behavior is sufficiently similar, we can remove the frame skip setting and just set it based on the throttle percentage. I'd be fine with that. For the time being I'm just fixing the behavior to be more correct, reordering the options, and adding a warning about vsync.

Keep in mind also that we still need to keep the normal throttle behavior to **SLOW DOWN** the emulator, for e.g. `25%`, `50%`, etc..

rkitover:
@ZachBacon anyway, I'm going to do more research on these issues to see if there are other solutions and close this for now. If you have any suggestions for my theme issue please let me know.

Issue #613, opened 2020-02-18
[Fuzzing] Hang in utilLoad / malloc


ps1337:
Heyo, this input file causes a hang upon loading it into VBA-M since the malloc call in utilLoad uses a size that's directly being read from the input file. I've attached a proof of concept
[hang1.gb.gz](https://github.com/visualboyadvance-m/visualboyadvance-m/files/4219151/hang1.gb.gz)

negativeExponent:
i dont get it. what "size" should be used to allocate? in anycase GB should realloc these when the size will not match what the header would indicate (assuming this is a valid rom, can't check since i cannot extract it)

ps1337:
You can load the `gb.gz` file directly into VBA-M. I was referring to a check whether the size that's being passed to `malloc` actually makes sense for a GB ROM. Using this sanitation the hang can maybe be fixed - only a suggestion though from my side

negativeExponent:
thanks for the reply, although i still dont understand the way you phrase it..

in any case the problem is that (size) actually goes negative during utilGetSize()
https://github.com/visualboyadvance-m/visualboyadvance-m/blob/master/src/Util.cpp#L642

solution should be utilGetSize should use unsigned types, much preferable uint64_t

![Screenshot_2020-02-19_17-16-36](https://user-images.githubusercontent.com/54053706/74820636-573dcd80-533d-11ea-8a97-604becbdfb7b.png)

Issue #612, opened 2020-02-18
[Fuzzing] Out of bounds writes


ps1337:
heyo, I've performed some fuzzing in VBA-M and found two OOB writes that are being triggered using these input files and an ASAN build of vbam

[write1.gb.gz](https://github.com/visualboyadvance-m/visualboyadvance-m/files/4219089/write1.gb.gz)
[write2.gb.gz](https://github.com/visualboyadvance-m/visualboyadvance-m/files/4219090/write2.gb.gz)

**I've got more samples for all issues I've created, if you need them please let me know :)**

Issue #611, opened 2020-02-18
[Fuzzing] Out of bounds reads


ps1337:
heyo, I've performed some fuzzing in VBA-M and found some OOB reads that are being triggered using these input files and an ASAN build of vbam

[read1.gb.gz](https://github.com/visualboyadvance-m/visualboyadvance-m/files/4219083/read1.gb.gz)
[read2.gb.gz](https://github.com/visualboyadvance-m/visualboyadvance-m/files/4219084/read2.gb.gz)

Issue #610, opened 2020-02-18
[Fuzzing] Null Derefs


ps1337:
heyo, I've performed some fuzzing in VBA-M and found some null derefs.

[null_deref.gb.gz](https://github.com/visualboyadvance-m/visualboyadvance-m/files/4219071/null_deref.gb.gz)
[null_deref2.gb.gz](https://github.com/visualboyadvance-m/visualboyadvance-m/files/4219072/null_deref2.gb.gz)
[null_deref3.gb.gz](https://github.com/visualboyadvance-m/visualboyadvance-m/files/4219073/null_deref3.gb.gz)

rkitover:
These are GB ROM files that cause the crash right?

ps1337:
yes indeed, it's loaded directly as `gb.gz` into the emulator

PR #620, opened 2020-02-28
limit max filesize and header check for valid rom files

negativeExponent:
https://github.com/visualboyadvance-m/visualboyadvance-m/issues/612
https://github.com/visualboyadvance-m/visualboyadvance-m/issues/611
https://github.com/visualboyadvance-m/visualboyadvance-m/issues/610
https://github.com/visualboyadvance-m/visualboyadvance-m/issues/609
These roms are invalid in a number of ways. An additional header check
should prevent loading invalid roms.

https://github.com/visualboyadvance-m/visualboyadvance-m/issues/613
File is over 1GB which is far to large for know max cart size. GBC is 8MB,
GBA max is 32MB. Limiting the file size that can be opened to 32MB prevents
issue.

strange that asan on libretro port does not cause asan from aborting. is vba-m
doing any extra checks that the one used as parameter to -fsanitize? In any case
these are just quick fixes for these common cases, and common i meant is that rom is expected to
be a known one.

### DIFF
--- a/src/gb/GB.cpp
+++ b/src/gb/GB.cpp
@@ -778,6 +778,34 @@ static const uint16_t gbColorizationPaletteData[32][3][4] = {
 #define GBSAVE_GAME_VERSION_12 12
 #define GBSAVE_GAME_VERSION GBSAVE_GAME_VERSION_12
 
+
+static bool gbCheckRomHeader(void)
+{
+    uint8_t nlogo[16] = {
+        0xCE, 0xED, 0x66, 0x66, 0xCC, 0x0D, 0x00, 0x0B,
+        0x03, 0x73, 0x00, 0x83, 0x00, 0x0C, 0x00, 0x0D
+    };
+
+    // Game Genie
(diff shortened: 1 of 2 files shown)


src/gb/GB.cpp
```diff
@@ -778,6 +778,34 @@ static const uint16_t gbColorizationPaletteData[32][3][4] = {
 #define GBSAVE_GAME_VERSION_12 12
 #define GBSAVE_GAME_VERSION GBSAVE_GAME_VERSION_12
 
+
+static bool gbCheckRomHeader(void)
+{
+    uint8_t nlogo[16] = {
```
rkitover: Make this const so the compiler does not allocate local storage.

rkitover:
Thank you for this, I left one comment.

rkitover:
This is merged, thank you very much!

Artoria2e5:
I agree that the size-check will fix the malloc issue (as suggested by the bug reporter), but I can't see how checking for a logo in the header actually avoids the crash in other issues. Won't it be trivial to just change the bytes in the crashing roms to have the logo instead? It's [not even executable code](https://stackoverflow.com/a/24630503), just some pointless data.

PR #621, opened 2020-02-29
Speedup/Turbo/Throttle fixes.

rkitover:
In the Speedup/Turbo config panel, make frameskip the first choice,
followed by throttle, and add a warning that throttle may not work due
to vsync.

Fix throttle=0 (unlimited speed) settings for both the Speedup/Turbo
config panel and the general throttle setting and make 100 the default.

### DIFF
--- a/src/wx/xrc/SpeedupConfig.xrc
+++ b/src/wx/xrc/SpeedupConfig.xrc
@@ -42,7 +42,7 @@
               <object class="sizeritem">
                 <object class="wxChoice" name="SpeedupThrottleSel">
                   <content>
-                    <item>None</item>
+                    <item>Full Speed/No Limit</item>
                     <item>25%</item>
                     <item>50%</item>
                     <item>75%</item>
@@ -80,84 +80,10 @@
         <border>5</border>
       </object>
       <object class="sizeritem">
-        <object class="wxStaticText">
-          <label>Speedup Frame Skip</label>
-          <font>
-            <weight>bold</weight>
-          </font>
+        <object class="wxCheckBox" name="SpeedupThrottleFrameSkip">
+          <label>Frame skip</label>
         </object>
         <flag>wxALL</flag>
-        <border>5</border>
-      </object>
-      <object class="sizeritem">
-        <object class="wxBoxSizer">
-          <orient>wxVERTICAL</orient>
-          <object class="sizeritem">
-            <flag>wxALL|wxEXPAND</flag>
-            <border>5</border>
-            <object class="wxBoxSizer">
-              <orient>wxHORIZONTAL</orient>
-              <object class="sizeritem">
-                <object class="wxStaticText">
-                  <label># of frames:</label>
-                  <size>60,-1d</size>
-                </object>
-                <flag>wxALL|wxALIGN_CENTRE_VERTICAL</flag>
-                <border>5</border>
-              </object>
-              <object class="sizeritem">
-                <object class="wxSpinCtrl" name="SpeedupFrameSkip">
-                  <value>9</value>
-                  <min>0</min>
-                  <max>30</max>
-                </object>
-                <flag>wxALL|wxALIGN_CENTRE_VERTICAL</flag>
-                <border>5</border>
-              </object>
-              <object class="sizeritem">
-                <object class="wxChoice" name="SpeedupFrameSkipSel">
-                  <content>
-                    <item>None</item>
-                    <item>1</item>
-                    <item>2</item>
-                    <item>3</item>
-                    <item>4</item>
-                    <item>5</item>
-                    <item>6</item>
-                    <item>7</item>
-                    <item>8</item>
-                    <item>9</item>
-                    <item>10</item>
-                    <item>11</item>
-                    <item>12</item>
-                    <item>13</item>
-                    <item>14</item>
-                    <item>15</item>
-                    <item>16</item>
-                    <item>17</item>
-                    <item>18</item>
-                    <item>19</item>
-                    <item>20</item>
-                    <item>21</item>
-                    <item>22</item>
-                    <item>23</item>
-                    <item>24</item>
-                    <item>25</item>
-                    <item>26</item>
-                    <item>27</item>
-                    <item>28</item>
-                    <item>29</item>
-                    <item>30</item>
-                  </content>
-                  <size>40,-1d</size>
-                </object>
-                <flag>wxALL|wxALIGN_CENTRE_VERTICAL</flag>
-                <border>5</border>
-              </object>
-            </object>
-          </object>
-        </object>
-        <flag>wxALL|wxEXPAND</flag>
         <border>5</border>
       </object>
       <object class="sizeritem">
--- a/po/wxvbam/wxvbam.pot
+++ b/po/wxvbam/wxvbam.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-03-13 14:40+0000\n"
+"POT-Creation-Date: 2020-03-27 09:03+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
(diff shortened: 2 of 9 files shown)


rkitover:
@denisfa we have more work to do here to improve the behavior and make it work more consistently, I am starting to think we have to combine the throttle and frame skip behaviors to get the desired effect more consistently regardless of configuration.

This however, just makes the existing system work correctly.

negativeExponent:
from the description, speed throttle at "No Limit" unless im mistaken, is suppose to run at whatever speed you system can. but on linux this does not do anything, with or without vsync, with or without compositor, fullscreen on windowed.

rkitover:
@negativeExponent Thank you for testing, this is actually a long-standing issue over which I had an argument with a former project member, this being one of the reasons why he left.

I can tell you that I have one linux configuration where this does in fact work.

We have determined that one of the causes can be vsync, if you haven't yet, can you please try the environment variable method of disabling vsync described here:

https://stackoverflow.com/questions/17196117/disable-vertical-sync-for-glxgears

the variables are different for mesa drivers such as amdgpu and nvidia. The environment variables are more likely to work than other methods of disabling vsync.

Please try one other thing, just to make sure, go into the General config and try setting the throttle there, and see if that also does not work.

There may be other reasons for throttle to not work that we have not determined yet

denisfa:
@rkitover There are a couple of bugs related to the text boxes of these options. You are probably aware of this, but if not, I can upload some screenshots.

We really need to change the name of our menu option currently `vsync` to something else. It is only for our OpenGL output module and it is most of times override by other config.

How are we going to go here? Can I commit too or do you prefer feedback only?

src/wx/xrc/SpeedupConfig.xrc
```diff
@@ -24,49 +24,55 @@
...
-                    <item>600%</item>
+                    <item>1</item>
+                    <item>2</item>
+                    <item>3</item>
+                    <item>4</item>
+                    <item>5</item>
+                    <item>6</item>
+                    <item>7</item>
+                    <item>8</item>
+                    <item>9</item>
```
denisfa: 9 frame skip means 1000% speedup... I don't think it is even playable beyond this.

src/wx/xrc/SpeedupConfig.xrc
```diff
@@ -81,7 +87,7 @@
       </object>
       <object class="sizeritem">
         <object class="wxStaticText">
-          <label>Speedup Frame Skip</label>
+          <label>Speedup Throttle (WARNING: May not work if VSYNC is enabled.)</label>
```
denisfa: It is not showing the entire text. It cuts on VSYNC.

denisfa:
The `No Limit` only works for me when using `__GL_SYNC_TO_VBLANK=0` (disabled vsync). Otherwise it does nothing as well (both audio and video remain the same).

About throttle, it goes like I expected: if I set `200%` without the above variable, it does not work (video stays the same, audio bugged). But if i go to `Video` >  `Configure...` > `Speed` and set `Number of frames to skip` to `1`, then it starts working. This is the way to go about this.

I will check on Windows and report back.

rkitover:
@denisfa Yeah, that's what I was trying to get at. That is exactly what we need to do. I will do it in this PR.

src/wx/xrc/SpeedupConfig.xrc
rkitover: The way this works currently, is if frameskip is selected that is the number of frames to skip when speedup is in effect. When speedup is in effect, the audio driver never waits and just drops excess data. So skipping more frames just makes speedup go faster by reducing rendering time.

src/wx/xrc/SpeedupConfig.xrc
rkitover: I see it on wx3-gtk2, I will test on other configs. May need some formatting options for the label in the XML.

rkitover:
@denisfa Yeah feel free to commit to this branch if you want.

src/wx/xrc/SpeedupConfig.xrc
rkitover: Actually, I will just implement your solution, then we don't even need the warning.

denisfa:
@rkitover Let's use the video frame skip to be based on throttle.

If `throttle<=100%`, then `frameSkip=0`
else `frameSkip=int(throttle / 100) - 1`

What we have on menu called `vsync`, change it to `video sync`. When this is ticked/enabled, force `frameSkip=0` over the throttle frameSkip.

| throttle  | frameSkip  |
| ------------------- | ------------------- |
|  100 |  0 |
|  200 |  1 |
|  300 |  2 |
|  400 |  3 |

Now, besides the slower throttle {25,50,75}%, I also wish to keep some intermediary ones, like {125,150,175}%. This case is where our solution falls, because it would require some type of counting.

I am still wondering how we can tackle this. If the user can disable its own driver vsync config, then this is not necessary at all. I will do some more research and report again.

denisfa:
@rkitover Can you test this on your end, please? Linux and Windows. I did on mine and it worked, but I could have extra confirmation. Try to start from a clean config, if possible.

src/wx/panel.cpp
```diff
@@ -2135,6 +2145,42 @@ GLDrawingPanel::GLDrawingPanel(wxWindow* parent, int _width, int _height)
...
 
+void GLDrawingPanel::ChangeVsync(bool enable)
+{
+    int vsync = enable ? 1 : 0;
+// non-portable vsync code
+#if defined(__WXGTK__) && defined(GLX_SGI_swap_control)
+    static PFNGLXSWAPINTERVALSGIPROC si = NULL;
+
+    if (!si)
+        si = reinterpret_cast<PFNGLXSWAPINTERVALSGIPROC>(glXGetProcAddress(reinterpret_cast<const unsigned char*>("glXSwapIntervalSGI")));
```
denisfa: @rkitover This right here is a very big YIKES. This was legit broken previously... I have no idea how we did not get a segfault at some point.

src/wx/panel.cpp
```diff
@@ -2135,6 +2145,42 @@ GLDrawingPanel::GLDrawingPanel(wxWindow* parent, int _width, int _height)
...
+
+    if (!si or si(vsync) != 0)
+        systemScreenMessage(_("Failed to set glXSwapIntervalSGI"));
+
+#else
+#if defined(__WXMSW__) && defined(WGL_EXT_swap_control)
+    static PFNWGLSWAPINTERVALEXTPROC si = NULL;
+
+    if (!si)
+        si = reinterpret_cast<PFNWGLSWAPINTERVALEXTPROC>(glXGetProcAddress(reinterpret_cast<const unsigned char*>("wglSwapIntervalEXT")));
```
denisfa: Likewise.

src/wx/panel.cpp
rkitover: Don't worry, we got all kinds of segfaults at all points. Like 99% of the github issues is some kind of crash.

rkitover:
@denisfa I think slightly modifying your formula would work for the intermediate values as well:

```cpp
uint8_t frame_skip = std::ceil((double)throttle / 100) - 1
```

I will try this now.

In fact this broken VM configuration will serve my purposes for testing under vsync or similar conditions.

rkitover:
@denisfa I implemented your idea and removed `speedup_frame_skip` entirely. Using my modification of the formula. I tried testing `175%` and such, but I don't think I trust myself to perceive changes in speed so small correctly, but it seems to work in any case. The idea is that the sound driver still controls the speed based on the throttle, so at `175%` one frame would still be skipped, but the speed would be correct.

All of this works correctly on my desktop linux configuration where normal throttle **DOES NOT** work.

In terms of usability and making things less confusing for users, your idea is very good. The actual issues with vsync and speed we'll need to give some attention to as well, but that is much harder.

I am getting an error in my desktop linux configuration:

```
Failed to set glXSwapIntervalSGI
```

this is with amdgpu.

I will try my intel laptop as well.

@negativeExponent could you please test again and let us know if the new behavior meets your needs at least to some degree?

denisfa:
@rkitover I am bit worried because this should be very noticeable when it is working. It is like day and night. I did a very bad commit of the vsync stuff, there is something missing. I think you should revert my commit here. I am sorry about this. I will do a proper vsync tackle on a separate pr. I am missing some checks here and I believe your case should be one of them.

For now, use the shell variable to enable/disable vsync and do your checks.

rkitover:
@denisfa moved and reworded your commit to the vsync branch.

Going to test this again with just my changes.

rkitover:
@denisfa tested on my desktop linux configuration where throttle did **NOT** work, and I am getting the expected results at both 125%, 200% and 600%.

Would like confirmation from you. Especially since this is your idea.

denisfa:
@rkitover I think I can make a comparison of how noticeable is when you cut frames using the formula:
```
uint8_t frame_skip = std::ceil((double)throttle / 100) - 1
```

My issues are with intermediary values, like {`125`,`150`,`175`}%. So I did an experiment and recorded side-by-side one emulator running with some shell variable to remove vsync (in this case, using nouveau and `vblank_mode=0` for throttle). The left one is always using throttle and the right one is mode turbo (with vsync enabled).

[200-throttle-turbo.avi.zip](https://github.com/visualboyadvance-m/visualboyadvance-m/files/4308764/200-throttle-turbo.avi.zip)

When you set turbo to 200%, then you jump one frame after every draw. The result is very smooth comparing one to another.

[150-throttle-turbo.avi.zip](https://github.com/visualboyadvance-m/visualboyadvance-m/files/4308778/150-throttle-turbo.avi.zip)
 
If you stare at one and then look at the other, you should notice a little bit of choppiness on the one using turbo. Especially on the screen transition. It looks like the right has some "stops" before fully changing screens.

[125-throttle-turbo.avi.zip](https://github.com/visualboyadvance-m/visualboyadvance-m/files/4308795/125-throttle-turbo.avi.zip)

Now this should be very noticeable. The turbo feels to jump a lot of frames (because it is, jumping 1 frame after every draw). This is the result of our formula for this case.

With `175%`, I have to be honest, I could not catch the stops. It is closer to 200%, when the formula works perfectly. I didn't  think it was worth bringing to the table.

One very important detail I realized is that if vsync is disabled, we should not frame skip. There is a check somewhere on our frontend to be done about this. Like 

if `vsync==disabled`, then `throttle=turbo_throttle`;

and it should stop there.

I believe the previous state, when you could use frameSkip or throttle for turbo, was better for this. Although you did your own check and confirmed it was working, so it has this going, what is pretty good.

rkitover:
I've thought about this and examined your videos.

I don't think we should figure the vsync setting into this, because it's not reliable enough to reflect the actual user configuration.

I propose adding a checkbox to the speedup/turbo dialog captioned simply "frame skip", it would default to set.

This would set the bool `speedup_throttle_frame_skip`, when false no frames would be skipped and the old throttle method would be used.

I think this is the best that can be done right now.

With vsync or similar conditions, with throttle set to 125%, the emulator will still run at 100% without skipping the additional frame. Without delving deeper into the drawing systems, there is no other way to do this.

I need to do more research about how the drawing is actually done and how the frequencies figure into it etc., but that can be done later.

rkitover:
Oh, one thing I got wrong, I need to add the throttle frame skip to the system frame skip, in case the user has a frame skip setting under the video config.

rkitover:
That I fixed, will wait for what you have to say about my checkbox idea before I proceed further.

denisfa:
@rkitover Sorry for the delay. You can go on with your idea. I just don't want to lose some panel options if we are going to eventually add them back.

rkitover:
This dialog was not in the original codebase, I added it a few months ago, without understanding how all of this works very well. The speedup_frame_skip option does not actually make sense, because it doesn't really matter how many frames you skip when speedup is in effect, so it's not very useful.

rkitover:
I've implemented the checkbox.

The logic for all of this is a bit convoluted, but I think this will be a better user experience than my previous attempts.

@denisfa your thoughts?

rkitover:
@denisfa would like your final review of all this.

denisfa:
Ok, we will tackle this again at some point. For now it is enough.

rkitover:
Yeah, I agree that this isn't "good" by any stretch of the imagination, but at least it's "better" :grin:

PR #622, opened 2020-03-01
Replace libpng with stb_image.

denisfa:
@rkitover This one is the work following this [comment](https://github.com/visualboyadvance-m/visualboyadvance-m/pull/614#discussion_r381165263). There is probably more to do, to fully remove it, but I want to start here.

Tested with Pokemon Crystal prints, from the pokedex screen. It at least does what last version did (`2.1.4`).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -968,6 +968,12 @@ if(ENABLE_DEBUGGER)
     )
 endif()
 
+set(
+    HDR_STB_IMAGE
+    headers/stb/stb_image.h
+    headers/stb/stb_image_write.h
+)
+
 include_directories(
     ${ZLIB_INCLUDE_DIR}
     fex
@@ -997,6 +1003,7 @@ add_library(
     ${HDR_FILTERS}
     ${SRC_DEBUGGER}
     ${HDR_DEBUGGER}
+    ${HDR_STB_IMAGE}
 )
 add_dependencies(vbamcore generate)
 set_property(TARGET vbamcore PROPERTY CXX_STANDARD 11)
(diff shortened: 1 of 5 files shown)


rkitover:
@denisfa Thank you for this, we've been talking about this for years instead of **Just Doing It (tm)**.

We need to meet sometime for a nice beverage of choice, maybe there's some kind of conference for emulator devs.

@ZachBacon you too.

ZachBacon:
> @denisfa Thank you for this, we've been talking about this for years instead of **Just Doing It (tm)**.
> 
> We need to meet sometime for a nice beverage of choice, maybe there's some kind of conference for emulator devs.
> 
> @ZachBacon you too.

I'd be alright with that, another idea, group discord chat (at least for the short term).

And as for just doing it... I'm terrible for that, such as my faudio work. I really should rewrite that completely instead of basing it off the xaudio one.

rkitover:
I'm just finishing up the commit to remove the dependency.

I don't think denisfa likes these chat things, and I'm happy I left them.

Writing is a more civilized form of communication.

If you get the forum back that would be nice.

I managed to actually get my server offline trying to reconfigure the networking, while for some reason my kvm is down, amazingly your server still works because it's bridged to the network card.

Just waiting for the kvm to come back up in a few hours.

ZachBacon:
I still need to manually write my sql file. It's a bit taxing to be honest, but a lot of the translation I'm testing is all local.

rkitover:
Ok libpng is gone as of 05bd2a95 :tada:

If you need any SQL help, I've done lots of SQL.

ZachBacon:
Right now I'm exploring two avenues, one of them is a converting script for phpBB's installer (however the one I have is built to migrate myBB 1.8.11 to phpBB 3.1, so It would need to be updated to use the newer installer syntax for phpBB and to make use of the newer sql tables for each.) And the other option is by hand, which is semi easier due to the fact I don't really have much experience with php syntax. If you have experience with php, if I gave you the sql database and converter and the link to the current phpBB 3.3 database, you think you could get it going? Would be easier than doing it by hand, while easier for me, would take a while given the size of vba-m's database.

PR #624, opened 2020-03-06
Fix possible SEGFAULT in SoundReset

Mystro256:
A fedora user reported a segfault that traced to the "soundDriver->reset();" line in soundReset.
I assume this is because soundDriver is probably null and there's no sound driver to reset.
I noticed you usually check if it's null before doing something, so I have it just return if there's no sound driver.

Feel free to re-implement if my assumption is wrong here.
Thanks!

### DIFF
--- a/src/gba/Sound.cpp
+++ b/src/gba/Sound.cpp
@@ -518,6 +518,8 @@ int soundGetEnable()
 
 void soundReset()
 {
+    if (!soundDriver)
+        return;
     soundDriver->reset();
 
     remake_stereo_buffer();


ZachBacon:
hey! nice to see you again @Mystro256 thanks for the contributions, it looks good to me, but I'll allow the other members take a look and give their input.

negativeExponent:
the problem here is that the return value for soundInit() is never check if the soundDriver failed or not. It should not be possible to reach CPUReset() where SoundReset() is called without the soundDriver initialized.

rkitover:
@negativeExponent Very good point, and I will do this check.

I will merge this too, since it doesn't hurt anything. And it's never bad to check for null pointers.

rkitover:
Welcome back @Mystro256, this is merged thank you very much!

rkitover:
@negativeExponent the soundInit() check is done in f1099a21.

PR #626, opened 2020-03-08
Use proper functions and checks for OpenGL VSYNC.

denisfa:
@rkitover Would you mind testing this solution for vsync? It worked on my linux (nvidia proprietary) and `Windows 10` (optimus).

I have a notebook with intel HD graphics. As soon as I finish installing Linux, I will test this there too.

I am not interested yet on changing the structure of our output module classes. I will be reviewing that soon, but I would prefer to start from zero. If you agree, I will delete the `vsync` branch.

```
The vsync has to be {dis,en}abled before a ROM is loaded or we need to
restart the output module to apply it.

We also support MESA based installations (Intel HD Graphics, AMD/ATI and
Nouveau) on Linux.
```

### DIFF
--- a/src/wx/panel.cpp
+++ b/src/wx/panel.cpp
@@ -2220,23 +2220,32 @@ void GLDrawingPanel::DrawingPanelInit()
     glClearColor(0.0, 0.0, 0.0, 1.0);
 // non-portable vsync code
 #if defined(__WXGTK__)
+    static PFNGLXSWAPINTERVALEXTPROC glXSwapIntervalEXT = NULL;
     static PFNGLXSWAPINTERVALSGIPROC glXSwapIntervalSGI = NULL;
     static PFNGLXSWAPINTERVALMESAPROC glXSwapIntervalMESA = NULL;
 
     char* glxQuery = (char*)glXQueryExtensionsString(glXGetCurrentDisplay(), 0);
 
+    if (strstr(glxQuery, "GLX_EXT_swap_control") != NULL)
+    {
+        glXSwapIntervalEXT = reinterpret_cast<PFNGLXSWAPINTERVALEXTPROC>(glXGetProcAddress((const GLubyte*)"glXSwapIntervalEXT"));
+        if (glXSwapIntervalEXT)
+            glXSwapIntervalEXT(glXGetCurrentDisplay(), glXGetCurrentDrawable(), vsync);
+        else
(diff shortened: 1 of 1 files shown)


denisfa:
I clicked wrong. I meant to create a draft. I need to test this.

denisfa:
@rkitover On a sidenote, what is the proper way to have a menu being enabled only when no ROM is loaded (and disabled otherwise)? I don't want to apply this here necessarily, but might have an use for something else I am working on.

rkitover:
The rom loading happens in panel.cpp IIRC, and you should be able to just call `Enable(bool)` on the wxMenuItem.

Might be something like this will work:

```cpp
XRCITEM("VSync")->Enable(false);
```

There are various state variables for the emulator and rom status I think.

rkitover:
On my desktop amdgpu I get no errors about the vsync functions anymore. Whether this actually works or not I cannot test right now.

Please test my turbo refactor when you get a chance.

denisfa:
I confirm this is working on Nouveau. AMD and intel HD to go (forgot about finishing the installation).

rkitover:
Nice.

Should we merge this now and add all support code separately?

denisfa:
@rkitover No, I still need to fix one thing. Just tested this on nvidia laptop (intel + nouveau) and found out a bug. Wait a little.

denisfa:
@rkitover Ok, as soon as you can test the latest on your setup, we can merge. My tests were:

- linux: nouveau and nvidia desktop;
- linux: intel+nvidia notebook; set up sessions for each graphic card controlling the display once.
- windows: intel+nvidia same notebook of above.

I am only missing your AMD test. I have no AMD graphics card at the moment.

rkitover:
I can confirm that the vsync setting works on amdgpu. I get a slight speedup with throttle with vsync off.

rkitover:
I was actually going to ask why you used a `reinterpret_cast` in the first place. I think `static_cast` would work fine here.

denisfa:
> I can confirm that the vsync setting works on amdgpu. I get a slight speedup with throttle with vsync off.

Alright, that is very good!

> I was actually going to ask why you used a reinterpret_cast in the first place. I think static_cast would work fine here.

No special reason. As soon as I did this change (from `(PFNGLXSWAPINTERVALSGIPROC)` to current), it started working for my nvidia card. Literally same code and functions. So I just kept like that for all.

PR #633, opened 2020-03-15
Allow binding specific server IP.

### LINKED ISSUE #632: [enhancement] set server ip for linking
Expected behavior
in `options>link>start network link>server` field the ability to adjust the ip address for which i want to run the server
### Actual behavior
the ip field grays out the moment i switch from client to server mode
`options>link>start network link` en then server, and the IP setting grays out.

### Build details

Operating System (Windows, Mac, Linux (state distribution), etc.): WIN10 1909

### Extra Comment

also no option for it in vbam.ini

rkitover:
Allow binding specific server IP.

For the netlink dialog, leave the server IP text field enabled when the
server radio button is selected and bind it to the new option
gopts.server_ip which defaults to "*" for binding on all addresses,
which was the previous behavior.

Copy this value to the new GBALink global IP_LINK_BIND_ADDRESS and use
it when creating a listening socket.

SFML supports creating sf:IpAddress values from hostnames, consequently
no additional code is needed to support hostnames in the netlink dialog.

Change GetLinkServerHost to return the primary local address when the
value is "*", which was the previous behavior, and the set value
otherwise.

Remove the min/max macros from GBALink as they are unnecessary and
conflict with std headers.

Require at least SFML 2.4 instead of 2.x for the IP binding
functionality in cmake.

- Fix #632

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,7 +131,7 @@ if(WIN32 AND ((NOT (MINGW AND MSYS)) OR CMAKE_TOOLCHAIN_FILE MATCHES mxe) AND NO
     set(SFML_STATIC_LIBRARIES TRUE)
 endif()
 
-find_package(SFML 2 COMPONENTS network system)
+find_package(SFML 2.4 COMPONENTS network system)
 
 if(SFML_FOUND)
     set(ENABLE_LINK_DEFAULT ON)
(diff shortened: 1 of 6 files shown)


denisfa:
> Remove the min/max macros from GBALink as they are unnecessary and
conflict with std headers.

Were they being used somewhere? Or some old version?

rkitover:
They seem to override things like `std::min` and cause compile errors if I use any modern C++ headers.

Issue #269, opened 2018-06-27
AGBPrint Logging non-functional


felixjones:
### Expected behavior

Logging debug window displays AGBPrint output when "Debug print" is enabled.

### Actual behavior

Nothing printed out.

### Steps to reproduce the behavior

1. Enable Debug print via: _Options_ -> _Gameboy Advance_ -> _Debug print_
2. Open debug output log window via: _Tools_ -> _Logging..._
3. Enable _AGBPrint_
4. Launch ROM that uses AGBPrint or similar debug output logging

### ROM details

Any GBA ROM that uses AGBPrint will fail to output to the Logging window.

### Option details

N/A

### Build details

64bit VisualboyAdvance-M 2.0.2 92ba676 

Operating System: Windows 10 64bit

Version of code: master

felixjones:
The read/write issues were related to my ROM file. Regardless, AGBPrint is still non functional.

rkitover:
Once you figure this out, could you please help us make it work in the release version as well?

felixjones:
I've investigated, indeed there is an issue in the release version where AGBPrint is not logging. I can only get it working with debug builds.

Re-opening as this is a confirmed issue now. Steps to re-produce are in the first comment.

@rkitover I am unable to create a release build environment, I am on Windows and the most I can get working is an SDL debug version without any user-interface. If someone can guide me on how to create a buildable release version on Windows then I will be happy to fix this myself.

I'm actually willing to fix a number of issues that I've encountered, but unfortunately I am having incredible difficulty creating a release project to do so. It's taken me 4 attempts over the course of 3 hours to get a basic SDL debug environment working so I could figure out there the issue lies with AGBPrint.

rkitover:
The easiest way is to use the msys2 prompt as described in the README.md on the project page, we are working on proper Visual Studio support but it's some time away.

The release version is the default for the Wx frontend, to use debug mode pass in `-DCMAKE_BUILD_TYPE=Debug`.

SBird1337:
Is there any progress on where this issue lies? I tried a debug build as well as of 2fedaa63a6720247822327d41bb909513be1c2f1 (now) but this is still non-functional.

rkitover:
Can you give me a rom that uses print that I can test, or tell me how to get it without linking it if it's not legal.

rkitover:
By the way, we are in initial stages of Visual Studio 2017 support and you should be able to build the GUI.

[...]

PR #634, opened 2020-03-17
Make AGBPrint output to our Logging window.

denisfa:
@rkitover This is for #269. The functions seem to be correct. We were just missing the output to our log window.


rkitover:
Beautiful, I remember spending a couple of hours looking at this and I was completely in the wrong place.

PR #635, opened 2020-03-18
Timer loop

rkitover:
Moving timer loop development here.

This is to fix app menus freezing the emulator and consequently the link connection on windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -182,6 +182,18 @@ endif()
 
 option(ENABLE_ONLINEUPDATES "Enable online update checks" ${ONLINEUPDATES_DEFAULT})
 
+set(TIMER_LOOP_DEFAULT OFF)
+
+if(WIN32 AND (X64 OR X86))
+    set(TIMER_LOOP_DEFAULT ON)
+endif()
+
+option(ENABLE_TIMER_LOOP "Enable timer based main loop" ${TIMER_LOOP_DEFAULT})
+
+if(ENABLE_TIMER_LOOP)
+    add_definitions(-DTIMER_LOOP)
+endif()
+
 set(LTO_DEFAULT ON)
 
 # gcc lto produces buggy binaries for 64 bit mingw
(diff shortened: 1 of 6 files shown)


denisfa:
@rkitover Just to bring this forward, the problem with this timer solution is that timers have the lowest priority on Windows, therefore they may not be executed at the moment we demand, but rather a little late. [Here is a better description of the issue.](https://trac.wxwidgets.org/ticket/17173)

For Linux it works flawless, but we don't need to do this for Linux.

It will work for our 100% throttle (~60FPS), but it will limit how far we can go. Example: using my notebook Windows 10, it goes up to 175%; anything beyond this point stops at 175%.

One way to solve this would be setting the timer to something below `1ms` (limited by wxWidgets).
The other way is to loop inside `GameArea::OnIdle`:

```cpp
void GameArea::OnIdle(wxIdleEvent& event) {
  ...
  auto start = chrono::steady_clock::now();
  auto end = chrono::steady_clock::now();
  do {
    mf->PollJoysticks();
    ...
    end = chrono::steady_clock::now();
  } while ((chrono::duration_cast<chrono::milliseconds>(end - start).count() <= 16)); // 16ms
}
```

This is extremely ugly! Both code and idea wise. That's why I was working on a thread solution, running this function on a dedicated thread. I stopped because I entered this dimension of issues where I may need to queue events between threads, what is not exactly necessary for us.

If you have any ideas about how to tackle this, tell me and I will do a deeper research about it. I feel like I stumbled upon a library limitation and I am not sure how to go next.

rkitover:
I have an idea.

If you can get the throttle to 175% that's more than good enough for link.

Why don't we change this code to be dynamically switchable and activate it only when there's a link connection.

rkitover:
I mean, we need to rethink CPU usage for this app in general, but perhaps not here.

The menus being modal only really matters for link, and while linked no one will use speedup, well maybe they will but most of them won't.

denisfa:
@rkitover  I will close this, since I am almost finishing my threading solution. It should be way more reliable than what I wanted here. I have some testing done, and it does work on Windows while browsing menus. I will ask your review when I have something.

PR #639, opened 2020-03-26
Winsparkle changes.

denisfa:
@rkitover This whole quarantine situation is actually running over me. I have lots to do and very little time, nothing I can do at the moment.

This is my idea for how I think the autoupdater should be structured. I wanted to isolate these functions both code and header inside its own dir. The issue for me was that there should be only two cases: windows and mac.  I was not sure if I should do C-style functions or a proper C++ class for it.

My decision here was considering that if this was a class, it would have a single public method, the `checkUpdatesUi`. The init would be in the constructor and shutdown in its destructor. So we would have two classes code. For our source files, we would have to init the object somewhere, but I was not sure where and when (one more global variable, like a pointer to this class for us to `new Autoupdater()`?). On the other hand, calling functions at the given moments felt a natural catch.

Another idea was to extend this to our SDL port or make it available if we ever change frontend. But none of this pleases me.

Whenever you can, check how everything is being organized, both file and code wise. After we agree on a solution, I will move to add the sparkle lib for mac updates.

Take your time to think about this, since I will be unable to quickly respond.


rkitover:
Hey, good to see you're still with us even when the zombies are attacking.

I am extremely busy too with various things, including family issues.

I will review everything here thoroughly very soon.

Please log in to our new forum too, everything is more or less back up and running now, and with markdown support now thanks to zach.

src/wx/autoupdater/wxmsw/autoupdater.cpp
```diff
@@ -0,0 +1,27 @@
+#include "../autoupdater.h"
+#include "../../../common/version_cpp.h"
+#include "winsparkle-wrapper.h"
+
+
+void initAutoupdater()
+{
+    wxString version(vbam_version);
+    size_t index = version.Find("-");
```
rkitover: Maybe you can use my contains template function in src/common/contains.h.

I think we can drop support for non-stl wx builds.

rkitover: Also thank you for fixing this. This will really help with the nightlies a lot. I need to do some more work on the windows server too.

src/wx/autoupdater/wxmsw/winsparkle-wrapper.cpp
```diff
@@ -46,11 +48,14 @@ WinSparkleDllWrapper::WinSparkleDllWrapper()
...
 }
 
+
 WinSparkleDllWrapper::~WinSparkleDllWrapper()
 {
     delete winsparkle_dll;
 
     // Wait for the program to exit and release the DLL before deleting the temp file, otherwise access is denied.
+    // This is needed due to `win_sparkle_init` locking the file until app execution is over.
+    // `win_sparkle_cleanup` does not release the lock.
```
rkitover: Interesting, I didn't know any of this.

I actually submitted a question issue for winsparkle at the time and the maintainer was extremely rude to me, he even deleted the issue. But I fixed the problem anyway.

src/wx/CMakeLists.txt
```diff
@@ -784,12 +784,15 @@ set(
 
 set(ALL_HDR_WX ${HDR_WX})
 
-list(APPEND ALL_SRC_WX winsparkle-wrapper.cpp)
```
rkitover: The reason I place these outside the if check is so that the gettext pot generation includes them.

src/wx/autoupdater/wxmsw/wxvbam.rc
```diff
@@ -1,58 +1,58 @@
-#pragma code_page(65001)
```
rkitover: We cannot move this file, because it is needed for all windows builds, not just with winsparkle.

Just update the include paths.

If we add other implementations for windows, we can always do the necessary logic with cpp macros.

rkitover: This is the reason for the ci failure btw.

rkitover:
This is excellent, thank you. And I really wanted the nightlies fix.

src/wx/autoupdater/wxmsw/autoupdater.cpp
rkitover: Sorry, I misread the code, you are using the index to trim the value.

I'd write it like this:

```cpp
if ((size_t index = version.Find('-')) != wxString::npos)
    version = version.SubString(0, index - 1);
```

rkitover: Also I have a `str_split()` function in strutils.cpp, which you could use like this:

```cpp
wxString version = str_split(vbam_version, '-')[0];
```

untested, but should work.

rkitover:
@denisfa if you're busy with the current situation, I can clean this up and merge it, if you have no objections.

denisfa:
@rkitover Thank you for your offer! I will try tomorrow to do some changes, then you can do as you wish. Just want to include at least the minimum of sparkle for macos and address some of your revisions.

Also, I have a question. Is this behaviour of creating temp files not suspicious on Windows? I see that we are dealing with some false positives, this might make it worser.

ZachBacon:
> @rkitover Thank you for your offer! I will try tomorrow to do some changes, then you can do as you wish. Just want to include at least the minimum of sparkle for macos and address some of your revisions.
> 
> Also, I have a question. Is this behaviour of creating temp files not suspicious on Windows? I see that we are dealing with some false positives, this might make it worser.

Most updaters typically do that, it shouldn't cause much change and because we do have a digital signature it should be alright once we get a few false positives dealt with, most actually iirc have been dealt with already.

WildCard65:
I have a few nitpicks manuel optimizations in regards to the wrapper file based on observation.

WildCard65:
I have a few nitpick optimizations for "winsparkle-wrapper.h" based on how it's layed out.

"static WinSparkleDllWrapper *GetInstance();" can be replaced with a "static const WinSparkleDllWrapper instant;" member variable.

The friend methods can be removed and replaced with directly calling the function pointers (requires making them public), the const instance will make the pointers themselves const and not the declaration.
Example: func_win_sparkle_init const winsparkle_init;

src/wx/CMakeLists.txt
denisfa: I duplicated the appends, just like before.

denisfa:
@rkitover I am currently installing mac on a VM to debug my changes for macos. I left them out of here, at least for now. Check all new changes when you can, please.

> I have a few nitpick optimizations for "winsparkle-wrapper.h" based on how it's layed out.
> 
> "static WinSparkleDllWrapper *GetInstance();" can be replaced with a "static const WinSparkleDllWrapper instant;" member variable.
> 
> The friend methods can be removed and replaced with directly calling the function pointers (requires making them public), the const instance will make the pointers themselves const and not the declaration.
> Example: func_win_sparkle_init const winsparkle_init;

@WildCard65 Thank you for the tip. I am going to take a look at this as soon as I can.

WildCard65:
My nitpicks would mostly affect Debug builds by eliminating call overhead.

WildCard65:
Now that I think about it, why do we indirectly use WinSparkle's API? At the very least we can just straight link them into the class to gain further CXX optimizations.

rkitover:
@denisfa BTW, I mentioned this before but wanted to again, the best way to run a mac vm on linux is this guide:

https://passthroughpo.st/new-and-improved-mac-os-tutorial-part-1-the-basics/

Use a libvirt/virt-manager setup with this driver:

https://github.com/pmj/QemuUSBTablet-OSX

works amazing!

I'm going to set up a macos vm on the server at some point for nightlies and for us to use too.

rkitover:
> "static WinSparkleDllWrapper *GetInstance();" can be replaced with a "static const WinSparkleDllWrapper instant;" member variable.

The reason I made the `GetInstance()` method was for lifetime management.

The static object is not allocated until the first time `GetInstance()` is called by an API wrapper.

However, in this case it makes no difference, as long as we use a singleton object we can load it at program startup.

rkitover:
> I have a few nitpick optimizations for "winsparkle-wrapper.h" based on how it's layed out.

I should mention that all of these functions are called only once, and the winsparkle code runs in its own process in the dll. So optimizing any of this is rather pointless.

denisfa:
@rkitover I was having a look at this for Sparkle and I believe we could use a separate pr for it. So, if you agree, let's review this as it is and then just do the finishing touches. I will create another PR for sparkle and we could enable the updater for windows builds.

I have time this upcoming week, so this should be done quickly.

>  @denisfa BTW, I mentioned this before but wanted to again, the best way to run a mac vm on linux is this guide:

Yes, thank you! My current setup has no spare gpu, but I was able to do some tests for macos catalina anyway.

rkitover:
All of this looks great as-is, just clean up the history a bit. Or I could do that.

> Yes, thank you! My current setup has no spare gpu, but I was able to do some tests for macos catalina anyway.

Yeah, I am not using GPU passthrough with that guide either, I could not get that to work, even though I have it working for my windows VM.

denisfa:
@rkitover I need your assistance here, please. According to the [docs](https://sparkle-project.org/documentation/):

1. [ ] Serve the appcast over HTTPS.
2. [ ] Sign the application via Apple’s Developer ID program.
3. [ ] Sign the published update archive with Sparkle’s EdDSA (ed25519) signature (optional)

We do sign our app when deploying for macos, right? Is there a way to do it without being a release? What is the viability of others requirements?

cmake/FindSparkle.cmake
```diff
@@ -0,0 +1,33 @@
+# -*- cmake -*-
+
+if (APPLE)
+  include(FindPackageHandleStandardArgs)
+
+  find_library(SPARKLE_FRAMEWORK
+    NAMES Sparkle
+    HINTS /usr/local/Caskroom/sparkle/1.22.0
```
denisfa, line 8: @rkitover I used `brew cask install sparkle` and it installed at this location. What can you tell me about this? Is this standard or some type of hack?

cmake/FindSparkle.cmake
```diff
@@ -0,0 +1,33 @@
...
+
+  if (SPARKLE_INCLUDE_DIR)
+    include_directories(${SPARKLE_INCLUDE_DIR})
+  endif()
+
+  find_package_handle_standard_args(Sparkle DEFAULT_MSG SPARKLE_INCLUDE_DIR SPARKLE_FRAMEWORK)
+  mark_as_advanced(SPARKLE_INCLUDE_DIR SPARKLE_FRAMEWORK)
+
+  set(CCS1 mkdir -p ${CMAKE_BINARY_DIR}/visualboyadvance-m.app/Contents/Frameworks)
+  set(CCS2 cp -R ${SPARKLE_FRAMEWORK} ${CMAKE_BINARY_DIR}/visualboyadvance-m.app/Contents/Frameworks/Sparkle.framework)
```
denisfa, line 20: @rkitover I can copy the framework to the app, accordingly to some examples I saw. Is this ok for cmake or is there a better way?

cmake/FindSparkle.cmake
```diff
@@ -0,0 +1,33 @@
...
+    COMMAND ${CCS2}
+  )
+  message(WARNING ${CCS1})
+  message(WARNING ${CCS2})
+
+endif()
+
+# https://github.com/patacrep/patagui/blob/master/cmake/CMakeLists-Mac.cmake
+# http://mikedld.com/f/transmission/cmake/macosx/CMakeLists.txt
+# https://github.com/nextcloud/desktop/blob/master/cmake/modules/FindSparkle.cmake
```
denisfa, line 33: @rkitover  Examples here.

src/wx/autoupdater/macos/autoupdater.cpp
```diff
@@ -0,0 +1,25 @@
...
+#include "sparkle-wrapper.h"
+
+SparkleWrapper autoupdater;
+
+void initAutoupdater()
+{
+    // even if we are a nightly, only check latest stable version
+    wxString version = str_split(vbam_version, '-')[0];
+    //autoupdater.addAppcastURL("http://data.vba-m.com/appcast.xml");
+    autoupdater.addAppcastURL("https://raw.githubusercontent.com/visualboyadvance-m/visualboyadvance-m/master/data/appcast.xml");
```
denisfa, line 13: @rkitover Here I am using our old `appcast.xml` URL, it is a placeholder until we decide what to do next.

cmake/FindSparkle.cmake
ZachBacon: Homebrew does things in their own directory and then symlinks it to /usr/local

rkitover:
> We do sign our app when deploying for macos, right? Is there a way to do it without being a release? What is the viability of others requirements?

I will add https support for that domain a bit later, that is not difficult.

I have previously sent you my apple codesigning key, if you don't have it I will send you another encrypted copy.

Install it to your login keychain and change the permissions to allow access for all apps.

To codesign, set the env var `LOGIN_KEYCHAIN_PASSWORD` to the password for your login keychain, which is the same as your macOS account password unless you changed it.

And pass `-DUPSTREAM_RELEASE=TRUE` to cmake.

This should work in both release and debug modes.

rkitover:
Regarding the macOS sparkle code, all of that looks fine.

Just don't call the updater instance class `Private`, call it `SparkleInstance` or something.

As far as the cmake support goes, I will need to look at it and do some research. We can't depend on brew, I'd rather have cmake download or build the framework.

denisfa:
@rkitover I have been thinking about this. We could go the same route of WinSparkle and have a release version on our dependencies [1.23.0](https://github.com/sparkle-project/Sparkle/releases/download/1.23.0/Sparkle-1.23.0.tar.bz2) and then do the same for includes and linking the lib. I think I can put something like this very fast, so what do you think? 

Building the lib is a little tricky, since it needs some external repos and can only be done with xcode GUI. Their makefile is currently broken and I have no experience to tell if the fix is simple or not.

> > We do sign our app when deploying for macos, right? Is there a way to do it without being a release? What is the viability of others requirements?
> 
> I will add https support for that domain a bit later, that is not difficult.
> 
> I have previously sent you my apple codesigning key, if you don't have it I will send you another encrypted copy.
> 
> Install it to your login keychain and change the permissions to allow access for all apps.
> 
> To codesign, set the env var `LOGIN_KEYCHAIN_PASSWORD` to the password for your login keychain, which is the same as your macOS account password unless you changed it.
> 
> And pass `-DUPSTREAM_RELEASE=TRUE` to cmake.
> 
> This should work in both release and debug modes.

Is this the comodo cert? Otherwise I don't think I have this key.

rkitover:
@denisfa gpg is telling me the key I have for you expired, please send me your new public key so I can send you the apple key.

ZachBacon:
@rkitover that reminds me, I need to send you my new one.

rkitover:
@denisfa and I definitely sent it to you on august 19th, try searching for my email

rkitover:
I set up https://data.vba-m.com/appcast.xml, the plain http URL is also active for our XP users.

Push anything you want to https://github.com/visualboyadvance-m/visualboyadvance-m.github.io and it will sync automatically within a minute.

rkitover:
> @rkitover I have been thinking about this. We could go the same route of WinSparkle and have a release version on our dependencies [1.23.0](https://github.com/sparkle-project/Sparkle/releases/download/1.23.0/Sparkle-1.23.0.tar.bz2) and then do the same for includes and linking the lib. I think I can put something like this very fast, so what do you think?

I'm completely fine with this, let's just keep the dependencies submodule for windows stuff and make a new repo for the mac stuff.

cmake would just automatically check it out like it does for windows.

Also on mac it's much easier to stick a framework into the .app bundle, and none of these shenanigans I did with the DLL will be necessary.

You may need to do something to make sure the dynamic linkage is correct, I've done this stuff so if you do let me know.

denisfa:
> @denisfa and I definitely sent it to you on august 19th, try searching for my email

Sorry, I can't find it. No idea why I would delete that email, since I kept all the others.


> I set up https://data.vba-m.com/appcast.xml, the plain http URL is also active for our XP users.

Let's use `https` for everything except if a given flag is passed. This way we can keep this only for XP builds.

> Push anything you want to https://github.com/visualboyadvance-m/visualboyadvance-m.github.io and it will sync automatically within a minute.

Thank you! Is this only for the `master` branch? I may need to do some testing and probably ruin the commit logs.

> > @rkitover I have been thinking about this. We could go the same route of WinSparkle and have a release version on our dependencies [1.23.0](https://github.com/sparkle-project/Sparkle/releases/download/1.23.0/Sparkle-1.23.0.tar.bz2) and then do the same for includes and linking the lib. I think I can put something like this very fast, so what do you think?
> 
> I'm completely fine with this, let's just keep the dependencies submodule for windows stuff and make a new repo for the mac stuff.

Not sure if this is good or not, but I manage to get something like this working:
```cmake
if(APPLE AND ENABLE_ONLINEUPDATES)
    include(FetchContent)
    FetchContent_Declare(Sparkle
        URL "https://github.com/sparkle-project/Sparkle/releases/download/1.23.0/Sparkle-1.23.0.tar.bz2"
    )
    FetchContent_MakeAvailable(Sparkle)

    find_library(SPARKLE_FRAMEWORK
      NAMES Sparkle
      HINTS ${sparkle_SOURCE_DIR}
    )
...
```
What do you think about it? It feels ugly, but I learned something new about cmake.

> cmake would just automatically check it out like it does for windows.

Isn't our `installdeps` that does this? At least for mingw.

> Also on mac it's much easier to stick a framework into the .app bundle, and none of these shenanigans I did with the DLL will be necessary.

Yes, you can copy them inside the `.app`.

> You may need to do something to make sure the dynamic linkage is correct, I've done this stuff so if you do let me know.

I can get to work minus the code signing... Do you pay for this or is there a way to get an Apple  Developer ID for free? Holy shit if this is the only way to get this to work. I don't think it is worth at all.

If you do have a key, can you sign the zip file? You will need to send it back to me since I am hard-coding some stuff. I can finish my testing after this.

[visualboyadvance-m.zip](https://github.com/visualboyadvance-m/visualboyadvance-m/files/4526325/visualboyadvance-m.zip)

rkitover:
> Sorry, I can't find it. No idea why I would delete that email, since I kept all the others.

Just send me your new gpg public key.

> Let's use `https` for everything except if a given flag is passed. This way we can keep this only for XP builds.

Fine with me, we can add a `-DHTTPS=NO` to cmake for the XP builds.

I still need to set up nightly XP builds, the 32 bit builds I have there now are not for XP.

> 
> > Push anything you want to https://github.com/visualboyadvance-m/visualboyadvance-m.github.io and it will sync automatically within a minute.
> 
> Thank you! Is this only for the `master` branch? I may need to do some testing and probably ruin the commit logs.

Don't worry about the history of that repo, we are just using it to sync.

> Not sure if this is good or not, but I manage to get something like this working:
> 
> ```cmake
> if(APPLE AND ENABLE_ONLINEUPDATES)
>     include(FetchContent)
>     FetchContent_Declare(Sparkle
>         URL "https://github.com/sparkle-project/Sparkle/releases/download/1.23.0/Sparkle-1.23.0.tar.bz2"
>     )
>     FetchContent_MakeAvailable(Sparkle)
> 
>     find_library(SPARKLE_FRAMEWORK
>       NAMES Sparkle
>       HINTS ${sparkle_SOURCE_DIR}
>     )
> ...
> ```
> 
> What do you think about it? It feels ugly, but I learned something new about cmake.

As far as cmake goes, this is actually extremely nice.

We need a flag for distro packagers that will not try to download external resources, but I can do that after we are done with this.

> 
> > cmake would just automatically check it out like it does for windows.
> 
> Isn't our `installdeps` that does this? At least for mingw.

Line 195 in the root CMakeLists.txt.

But with your FetchContent module I guess this will not be necessary at all!

> I can get to work minus the code signing... Do you pay for this or is there a way to get an Apple Developer ID for free? Holy shit if this is the only way to get this to work. I don't think it is worth at all.
> 
> If you do have a key, could you sign the zip file, please? You will need to send it back to me since I am hard-coding some stuff. I can finish my testing after this.
> 
> [visualboyadvance-m.zip](https://github.com/visualboyadvance-m/visualboyadvance-m/files/4526325/visualboyadvance-m.zip)

Here you go:

[visualboyadvance-m.zip](https://github.com/visualboyadvance-m/visualboyadvance-m/files/4528869/visualboyadvance-m.zip)

As I said I am happy to send you my cert, I just need your gpg public key. You can sign with my cert anything you want.

denisfa:
@rkitover Thank you very much! I have an update on this, just finishing some code. I will need an extra round of discussion here, since there is something that I would like to add to our cmake build, but I am not sure how. I will continue working on this tomorrow.

cmake/FindSparkle.cmake
denisfa: @ZachBacon Thank you for the info. We have moved from this solution, but it is good to know for future reference.

denisfa:
@rkitover Not sure what is happening with the `codesign` tool, but it seems the `--deep` flag is not working properly. The Sparkle framework is not being signed at all. 

```
codesign -s "Developer ID Application" --deep ./visualboyadvance-m.app
codesign -s "Developer ID Application" ./visualboyadvance-m.app/Contents/Frameworks/Sparkle.framework/Versions/A
```

Could you please sign the framework using the above command? The reference is [here](https://developer.apple.com/library/archive/technotes/tn2206/_index.html#//apple_ref/doc/uid/DTS40007919-CH1-TNTAG13). This will be the last testing.

rkitover:
Unfortunately signing the framework separately is not sufficient.

Apparently the issue with your zip is that the framework was copied incorrectly, not preserving symlinks. Whatever you used to copy the framework into the .app has to preserve relative symlinks.

For example `visualboyadvance-m.app/Contents/Frameworks/Sparkle.framework/Versions/Current` is supposed to be a symlink pointing to `A`.

More on this here:

https://stackoverflow.com/questions/25969946/osx-10-9-5-code-signing-v2-signing-a-framework-with-bundle-format-is-ambiguou

rkitover:
So like if you used `cp -r` use `cp -a` instead.

denisfa:
> So like if you used `cp -r` use `cp -a` instead.

That was exactly my mistake... Thank you! Here is the new one.

[visualboyadvance-m.zip](https://github.com/visualboyadvance-m/visualboyadvance-m/files/4549626/visualboyadvance-m.zip)

rkitover:
This time I was able to sign correctly, but I still had to sign the framework separately, here are the commands I used:

```bash
rkitover@catalina  ➤  codesign -v -f -s 'Developer ID Application' ./visualboyadvance-m.app/Contents/Frameworks/Sparkle.framework/Versions/A
./visualboyadvance-m.app/Contents/Frameworks/Sparkle.framework/Versions/A: replacing existing signature
./visualboyadvance-m.app/Contents/Frameworks/Sparkle.framework/Versions/A: signed bundle with Mach-O thin (x86_64) [org.sparkle-project.Sparkle]
✔  ~
rkitover@catalina  ➤  codesign -v -s 'Developer ID Application' ./visualboyadvance-m.app
./visualboyadvance-m.app: signed app bundle with Mach-O thin (x86_64) []
✔  ~
rkitover@catalina  ➤  codesign -vvv ./visualboyadvance-m.app
--prepared:/Users/rkitover/visualboyadvance-m.app/Contents/Frameworks/Sparkle.framework/Versions/Current/.
--validated:/Users/rkitover/visualboyadvance-m.app/Contents/Frameworks/Sparkle.framework/Versions/Current/.
./visualboyadvance-m.app: valid on disk
./visualboyadvance-m.app: satisfies its Designated Requirement
✔  ~
rkitover@catalina  ➤  spctl --verbose=4 --assess ./visualboyadvance-m.app/
./visualboyadvance-m.app/: accepted
source=Developer ID
```

And here is the signed file:

[visualboyadvance-m.zip](https://github.com/visualboyadvance-m/visualboyadvance-m/files/4549745/visualboyadvance-m.zip)

denisfa:
@rkitover Thank you very much! I can confirm that it is working without needing to use Sparkle `sign_update`. I am going to clean my WIPs and finish this.

rkitover:
Nice, I'm going to work on the aspect ratio stuff a bit tonight.

denisfa:
> Fine with me, we can add a -DHTTPS=NO to cmake for the XP builds.

Done.

> We need a flag for distro packagers that will not try to download external resources, but I can do that after we are done with this.

@rkitover  What do you mean here? macOS distro packagers? Is this a thing?

rkitover:
> > Fine with me, we can add a -DHTTPS=NO to cmake for the XP builds.
> 
> Done.

Nice, thank you, I still need to do the XP nightly.

> 
> > We need a flag for distro packagers that will not try to download external resources, but I can do that after we are done with this.
> 
> @rkitover What do you mean here? macOS distro packagers? Is this a thing?

Yeah actually you're right, I wasn't thinking clearly. This is only a thing we need to worry about for Linux, and as far as that goes we're good for now.

denisfa:
@rkitover Just finished testing the `msbuild` build and it worked as expected. So I am just waiting for you to give a last OK for me to merge this. I will squash some commits during the merge.

rkitover:
@denisfa I have a couple of suggestions for the DLL change, and I will review all of the rest as well:

```cpp
while(FreeLibrary(hMod));
```

Better to qualify root functions on Win32 and to not busy loop, so:

```cpp
while (::FreeLibrary(hMod)) {
    wxMilliSleep(50);
}
```

rkitover:
BTW you don't need to use msbuild anymore, pass `-G Ninja` to cmake on windows and use ninja, it's much much better.

CMakeLists.txt
```diff
@@ -176,11 +176,12 @@ option(ENABLE_FFMPEG "Enable ffmpeg A/V recording" ${FFMPEG_DEFAULT})
...
 set(ONLINEUPDATES_DEFAULT OFF)
 
-if(WIN32 AND (X64 OR X86))
-    set(ONLINEUPDATES_DEFAULT ON) # use winsparkle
+if((WIN32 AND (X64 OR X86)) OR APPLE) # winsparkle/sparkle
+    set(ONLINEUPDATES_DEFAULT ON)
 endif()
 
 option(ENABLE_ONLINEUPDATES "Enable online update checks" ${ONLINEUPDATES_DEFAULT})
+option(HTTPS "Use https URL for winsparkle" ON)
```
rkitover, line 184: Here we need to add:

```cmake
if(NOT HTTPS)
    add_definitions(-DNO_HTTPS)
endif()
```

rkitover:
👍 All clear to merge!

rkitover:
Oh yeah, I forgot to mention, the reason that ping command didn't work in wine is that it uses the windows ping options not the linux ones.

Issue #625, opened 2020-03-06
Emulator can't load BIOS at all


Hulkninja:
### Expected behavior
Emulator loads the provided BIOS upon launch
### Actual behavior
Emulator gives the error message "Could not load BIOS D:\filepath\VisualBoyAdvance-M\BIOS.ROM" and skips the BIOS, booting straight into the game
### Steps to reproduce the behavior
Set up BIOS, set emulator to run BIOS, start game
### Crash bugs
No crash
If this is a crash bug, and you have the time and inclination, please provide a
symblic backtrace.

Instructions for doing so are
[here](https://github.com/visualboyadvance-m/visualboyadvance-m/blob/master/README.md#reporting-crash-bugs).

### ROM details

Which games are affected by the issue (please state if GBA or GB game):
All games I have tried
Which games, if any, are NOT affected by the issue (please state if GBA or GB game):
None that I know of
Include screenshots of ROM Information (File > ROM Information):
Can't get any info as the emulator refuses to read/open the BIOS
If you are using a BIOS file, please say which one.
Standard, unaltered BIOS. I've tried multiple sources and all of them result in the exact same message and failure. I can provide one of the specific BIOS I've tried, and/or a link to it.
### Option details
All except bios and link options are default
Video driver (simple, opengl):
Default, I assume simple
Sound driver (directsound, xaudio, openal, SDL):
Default settings, but the bug isn't sound related
### Build details

Operating System (Windows, Mac, Linux (state distribution), etc.):
Windows
If using Linux, specify if using xorg or Wayland:

Version of code (pre-built binary version, or commit ref, or just "master"):
2.1.4, pre-built I think
### Hardware details:
It's on a gaming laptop. The hardware isn't the issue here, I'm pretty sure of that much...
If your joystick or gamepad isn't working in Linux, please try to troubleshoot
using the information on this page:
https://wiki.archlinux.org/index.php/Gamepad **BEFORE** filing an issue. We use
SDL for joystick support, any problems will generally be an issue with SDL and
**NOT** the emulator.

If relevant, the joystick you are using and the drivers you are using for it:

ZachBacon:
It maybe your bios image, as it's working just fine for me in the release and nightlies.
![bios](https://user-images.githubusercontent.com/8948676/76130037-49dc4080-5fd7-11ea-8f2f-6eec21830e0e.png)

Hulkninja:
[...]

PR #644, opened 2020-04-10
[Windows] UTF8

denisfa:
@rkitover The issue #625 was more complicated than just the BIOS not loading. When I started playing with files on Windows, I was only being able to load the game (zip) and BIOS. Not battery, save states, or any other file associated with my games.

I tracked what was happening with the BIOS: 
```
./src/gba/GBA.cpp: void CPUInit(const char* biosFileName, bool useBiosFile)  ==> utilLoad(biosFileName, CPUIsGBABios, bios, size)

./src/Util.cpp: uint8_t *utilLoad(const char *file, bool (*accept)(const char *), uint8_t *data, int &size) ==> fex_t *fe = scan_arc(file_conv, accept, buffer);

./src/Util.cpp: static fex_t *scan_arc(const char *file, bool (*accept)(const char *), char (&buffer)[2048]) ==> fex_err_t err = fex_open(&fe, file);

./fex/fex/fex.h:80:fex_err_t fex_open(fex_t **out, const char path[]);
./fex/fex/fex.cpp:225:BLARGG_EXPORT fex_err_t fex_open( fex_t** fe_out, const char path [] )

./fex/fex/File_Extractor.h:237:#define FEX_FILE_READER Std_File_Reader ==> RETURN_ERR( blargg_fopen( &f, path ) );

./fex/fex/Data_Reader.cpp:510:static FILE* blargg_fopen( const char path [], const char mode [] )
```

On Windows, we have `UTF16-LE` being used everywhere; the function `blargg_fopen` gets a `const char *` and converts it to `const wchar_t *` (multibyte to widechar) to use with `_wfopen`, a wide-char friendly `fopen`. So I decided to take a look around and verify why other functions were not getting the same calls or conversion somewhere.

The issues is like this: if you are using a codepage on Windows that does not support a given char, some conversion functions replace it with a broken char (eg `?????`), so it can no longer be used as a path to a file (at least as a `const char *`).

I had a test game with japanese chars (my codepage was `CP_ACP` for english-USA install with the `pt_BR` one available as well):

- `四月是你的謊言.zip`
- `四月是你的謊言-01.sgm`
- `四月是你的謊言.sav`

If the char belongs to your codepage, you can use `.mb_str()` to retrieve the `const char *` for `fopen` without issues. But if not, like using the example game battery, then our path becomes garbled and it cannot be read with `fopen`.

I did this same testing using other frontends, to see if this wasn't going to be an overkill, but all I tried ended up being a crash or the game not being recognized as a game... This of course is no issue at all on Linux, since everything "behaves" as far as codification goes.

The solution here is to always use UTF8 internally for our functions and, on Windows, to convert the UTF8 to UTF16 very alike how `blargg_fopen` does.  I also needed a wide-char func from `zlib` (bumps version to `1.27`).

wxWidgets for some reason cannot print `wchar_t *` properly. I believe it is because some unicode change they did between `2.9` and `3.0`. But at this point this is a detail, probably the easiest part here.

- [X] print proper verbose messages for loading/saving files.

Feel free to comment everywhere.

@ZachBacon Are you interested in this as well? For the SDL port? Does it make sense for it?

### DIFF
--- a/src/Util.cpp
+++ b/src/Util.cpp
@@ -72,6 +72,45 @@ bool FileExists(const char *filename)
 #endif
 }
 
+#ifdef _WIN32
+#include <windows.h>
+
+wchar_t* utf8ToUtf16(const char *utf8)
+{
+    wchar_t *utf16 = nullptr;
+    size_t size = MultiByteToWideChar(CP_UTF8, 0, utf8, -1, NULL , 0);
+    if (size == 0) return nullptr; // error
+    utf16 = new wchar_t[size];
+    size = MultiByteToWideChar(CP_UTF8, 0, utf8 , -1, utf16, size);
+    if (size == 0) {
+        delete[] utf16;
+        return nullptr; // error
+    }
+    return utf16;
+}
+#endif // _WIN32
+
+FILE* utilOpenFile(const char *filename, const char *mode)
+{
+    FILE *f = NULL;
+#ifdef _WIN32
+    wchar_t *wfilename = utf8ToUtf16(filename);
+    if (!wfilename) return nullptr;
+    wchar_t *wmode = utf8ToUtf16(mode);
+    if (!wmode) {
+        delete[] wfilename;
+        return nullptr;
+    }
+
+    f = _wfopen(wfilename, wmode);
+    delete[] wfilename;
+    delete[] wmode;
+#else
+    f = fopen(filename, mode);
+#endif // _WIN32
+    return f;
+}
+
 // Get user-specific config dir manually.
 // apple:   ~/Library/Application Support/
 // windows: %APPDATA%/
@@ -504,10 +543,6 @@ static bool utilIsImage(const char *file)
         return utilIsGBAImage(file) || utilIsGBImage(file);
 }
 
-#ifdef WIN32
-#include <windows.h>
-#endif
-
 IMAGE_TYPE utilFindType(const char *file, char (&buffer)[2048]);
 
 IMAGE_TYPE utilFindType(const char *file)
@@ -590,7 +625,7 @@ uint8_t *utilLoad(const char *file, bool (*accept)(const char *), uint8_t *data,
 
         if (size > MAX_CART_SIZE)
                 return NULL;
-        
+
         uint8_t *image = data;
 
         if (image == NULL) {
@@ -722,6 +757,19 @@ void utilWriteData(gzFile gzFile, variable_desc *data)
         }
 }
 
+gzFile utilAutoGzOpen(const char *file, const char *mode)
+{
+#ifdef _WIN32
+        wchar_t *wfile = utf8ToUtf16(file);
+        if (!wfile) return nullptr;
+        gzFile handler = gzopen_w(wfile, mode);
+        delete[] wfile;
+        return handler;
+#else
+        return gzopen(file, mode);
+#endif
+}
+
 gzFile utilGzOpen(const char *file, const char *mode)
 {
         utilGzWriteFunc = (int(ZEXPORT *)(gzFile, void *const, unsigned int))gzwrite;
@@ -729,7 +777,7 @@ gzFile utilGzOpen(const char *file, const char *mode)
         utilGzCloseFunc = gzclose;
         utilGzSeekFunc = gzseek;
 
-        return gzopen(file, mode);
+        return utilAutoGzOpen(file, mode);
 }
 
 gzFile utilMemGzOpen(char *memory, int available, const char *mode)
--- a/src/wx/panel.cpp
+++ b/src/wx/panel.cpp
@@ -81,7 +81,7 @@ void GameArea::LoadGame(const wxString& name)
 
     // auto-conversion of wxCharBuffer to const char * seems broken
     // so save underlying wxCharBuffer (or create one of none is used)
-    wxCharBuffer fnb(fnfn.GetFullPath().mb_str(wxConvUTF8));
+    wxCharBuffer fnb(UTF8(fnfn.GetFullPath()));
     const char* fn = fnb.data();
     IMAGE_TYPE t = badfile ? IMAGE_UNKNOWN : utilFindType(fn);
 
@@ -158,7 +158,7 @@ void GameArea::LoadGame(const wxString& name)
 
         if (loadpatch) {
             int size = rom_size;
-            applyPatch(pfn.GetFullPath().mb_str(), &gbRom, &size);
+            applyPatch(UTF8(pfn.GetFullPath()), &gbRom, &size);
 
             if (size != (int)rom_size)
                 gbUpdateSizes();
@@ -183,7 +183,9 @@ void GameArea::LoadGame(const wxString& name)
         gbGetHardwareType();
 
         bool use_bios  =  gbCgbMode ? useBiosFileGBC : useBiosFileGB;
-        const char* fn = (gbCgbMode ? gopts.gbc_bios : gopts.gb_bios).mb_str();
+
+        wxCharBuffer fnb(UTF8((gbCgbMode ? gopts.gbc_bios : gopts.gb_bios)));
+        const char* fn = fnb.data();
 
         gbCPUInit(fn, use_bios);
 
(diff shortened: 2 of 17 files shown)


denisfa:
```
..\src\common\SoundSDL.cpp(73): error C2589: '(': illegal token on right side of '::'
..\src\common\SoundSDL.cpp(73): error C2062: type 'unknown-type' unexpected
..\src\common\SoundSDL.cpp(73): error C2059: syntax error: ')'
..\src\common\SoundSDL.cpp(96): warning C4267: '-=': conversion from 'size_t' to 'unsigned int', possible loss of data
```

LOL. I will have a better look at this tomorrow.

src/Util.cpp
```diff
@@ -72,6 +72,43 @@ bool FileExists(const char *filename)
 #endif
 }
 
+#ifdef _WIN32
+wchar_t* utf8ToUtf16(const char *utf8)
+{
+    wchar_t *utf16 = NULL;
```
rkitover, line 78: You can use the nicer C++11 version of `NULL` called `nullptr`.

src/wx/wxhead.h
```diff
@@ -113,8 +113,7 @@ static inline void DoSetAccel(wxMenuItem* mi, wxAcceleratorEntry* acc)
 #define XRCCTRL(win, id, type) XRCCTRL_I(win, XRCID(id), type)
 #define XRCCTRL_D(win, id, type) XRCCTRL_I(win, XRCID_D(id), type)
 
-// wxWidgets provides fn_str(), but no mb_fn_str() or equiv.
-#define mb_fn_str() mb_str(wxConvFile)
+#define UTF8(wxstring) ((wxstring).mb_str(wxConvUTF8))
```
rkitover, line 116: In C++ it's better to use a function instead of a cpp macro, the compiler takes care of all the inlining etc..

src/wx/panel.cpp
```diff
@@ -578,14 +580,14 @@ bool GameArea::LoadState()
 bool GameArea::LoadState(int slot)
 {
     wxString fname;
-    fname.Printf(SAVESLOT_FMT, game_name().mb_str(), slot);
+    fname.Printf(SAVESLOT_FMT, game_name().wc_str(), slot);
```
rkitover, line 583: What is your thinking about when we should use this `UTF8()` function as opposed to `.wc_str()` or `.c_str()`.

Do we want to aim for consistency in all the wx code with respect to this?

denisfa: We should always use `UTF8()` when we are going to call a core function for path, but it has to be as late as possible.
`.c_str()` is very destructive for our issue because it uses current locale encoding implicitly. We lose chars not recognized by the current codepage.
`.wc_str()` is always safe for us, although printing may be a problem (as in to status bar or drawing on screen).

We just must do this when calling our core functions. I don't think there are any major issues elsewhere.

denisfa:
> Looks great.
> 
> We should think about making all the codepage conversions consistent, if not here than at some point in the future.

There is nothing much we can do if the libs we use are not unicode friendly on Windows. We are always going to need either a wrapper or be a little bit more intrusive to solve it.

> Too many commits for the number of logical changes, squash some together into more general commits.

Absolutely. I have one question, though. If you go through the commits, you will see a weird spacing being fixed. How do we signal this type of fix on a PR? Sometimes I am not changing files due to how out of place this is.

src/Util.cpp
denisfa: Done.

src/wx/wxhead.h
denisfa: Done.

denisfa:
@rkitover I will squash commits after the review. Found out how to print correctly, at least on the wx frontend. The terminal part now prints something too, although it falls on the codepage issue.

> We should think about making all the codepage conversions consistent, if not here than at some point in the future.

What are you thinking here? At least for documenting what we should do.

ZachBacon:
@denisfa sorry for taking some time to get to you. But yeah probably would be a good idea to apply it for the SDL Binary

rkitover:
@denisfa sorry, I was confused about all of this, but I did a bit of research.

This page has some good details about wide characters, the `wchar_t` type etc.:

https://stackoverflow.com/questions/4588302/why-isnt-wchar-t-widely-used-in-code-for-linux-related-platforms

Here's what I learned:

- Windows unicode representation is UTF-16 using a 2 byte `wchar_t` type. This means that a codepoint can span multiple wide characters. So that we know. Also Windows often has separate APIs ending in either `A` for ASCII or `W` for widechar/unicode. The ASCII APIs do not support UTF-8, so you have to use widechars and UTF-16 for unicode on Windows.

- On Linux and such, UTF-8 is more prevalent and `wchar_t` is not used as much.

- On Linux and such, `wchar_t` is 4 bytes and can represent any unicode codepoint as a single value.

What exactly all of this means for us, I'm not completely sure yet, I'll have to think about it some more and look at our usage of character set conversions.

rkitover:
Regarding the appveyor fail, I will take a look.

rkitover:
Read through the changes again, looks good to me.

I understand why you are using `.wc_str()` now, given the above, and given that wx will do whatever necessary conversions as non-destructively as possible.

rkitover:
@denisfa I thought about this some more.

You have code like this:

```cpp
wxString str;
str.Printf("foo %s", str2.wc_str());
```

In this case, `str2.wc_str()` returns an object that is convertible to `wchar_t*` but is actually a type of `wxString`. The `wchar_t*` conversion is not actually used. Can you try this instead:

```cpp
wxString str;
str.Printf("foo %s", str2);
```

denisfa:
> @denisfa sorry, I was confused about all of this, but I did a bit of research.

Yeah, you nailed it. This is only useful for Windows, and under these circumstances.



> @denisfa I thought about this some more.
> 
> You have code like this:
> 
> ```c++
> wxString str;
> str.Printf("foo %s", str2.wc_str());
> ```
> 
> In this case, `str2.wc_str()` returns an object that is convertible to `wchar_t*` but is actually a type of `wxString`. The `wchar_t*` conversion is not actually used. Can you try this instead:
> 
> ```c++
> wxString str;
> str.Printf("foo %s", str2);
> ```

These were one of my first tries tackling this. I agree with you that it should be enough, but unfortunately did not work. While querying the docs, I found about this mysteriously [pointer to the internal string contents in wide character mode (Windows)](https://docs.wxwidgets.org/3.0/classwx_string.html#a6cd4782263a3ed4064eca915eb6e27e6).

It works as expected on Windows and does not change anything for Linux. What is weird here is how this works for the GUI, but not for command line. If you try `wxPuts` or similar, it does not display correctly (Windows).

> Regarding the appveyor fail, I will take a look.

I just want to be sure this is not a side effect of something I am doing here.

> @denisfa sorry for taking some time to get to you. But yeah probably would be a good idea to apply it for the SDL Binary

Can I add as another PR? Or do you prefer this to be done here ASAP?

rkitover:
> These were one of my first tries tackling this. I agree with you that it should be enough, but unfortunately did not work. While querying the docs, I found about this mysteriously [pointer to the internal string contents in wide character mode (Windows)](https://docs.wxwidgets.org/3.0/classwx_string.html#a6cd4782263a3ed4064eca915eb6e27e6).
> 
> It works as expected on Windows and does not change anything for Linux. What is weird here is how this works for the GUI, but not for command line. If you try `wxPuts` or similar, it does not display correctly (Windows).

I will write some additional docs for our manual with all of this and ask you to review it. Thank you for clarifying!

> > @denisfa sorry for taking some time to get to you. But yeah probably would be a good idea to apply it for the SDL Binary
> 
> Can I add as another PR? Or do you prefer this to be done here ASAP?

Definitely not a high priority, IMO, whichever is easier for you.

rkitover:
@denisfa I just tested a windows build of the branch, and the appveyor fail is legitimate.

I suspect your addition of:

```cpp
#include <windows.h>
```

is causing some kind of macro interactions.

I can take a look at this if you like.

rkitover:
Regarding documentation changes related to this, please review:

https://github.com/visualboyadvance-m/visualboyadvance-m/pull/649

denisfa:
> @denisfa I just tested a windows build of the branch, and the appveyor fail is legitimate.
> 
> I suspect your addition of:
> 
> ```c++
> #include <windows.h>
> ```
> 
> is causing some kind of macro interactions.
> 
> I can take a look at this if you like.

Let me try a couple of things first. I will squash commits before merge, so don't worry about meaningless commits for the moment.

rkitover:
Looks like you got it!

denisfa:
> Looks like you got it!

Yes, I learned that on Windows we should include its headers on cpp files, accordingly to what we need to where and when we need to use. Just checking the SDL stuff and we should merge this.

denisfa:
@rkitover I will do a separate PR for the SDL port. There are a couple of things I want to check first. Check if there is something missing. If there is not, I will merge this later.

rkitover:
@denisfa the appveyor fail is a spurious vcpkg rebuild that ran out of time.

You have a go from me, merge whenever you're ready.

PR #649, opened 2020-04-12
Add to dev manual on translations and char sets.

rkitover:
Document how we interact with transifex and how to pass strings to
non-wxWidgets code.


denisfa:
Looks great!

PR #655, opened 2020-04-26
Hide menubar + minor UI refactor.

rkitover:
Add ui/hideMenuBar boolean option, defaulting to on, which hides the
main menubar when the mouse is idle or outside the frame.

This is disabled on mac, because on macs the main menubar is not part of
the application window.

Fix pointer hiding/unhiding by connecting panel events to the gamearea
mouse event handler.

Clean up the pausing when menus are opened code, make it actually work
on Windows.

### DIFF
--- a/po/wxvbam/wxvbam.pot
+++ b/po/wxvbam/wxvbam.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-04-12 00:19+0000\n"
+"POT-Creation-Date: 2020-04-27 20:23+0000\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
@@ -155,7 +155,7 @@ msgstr ""
 msgid "VBA cheat lists (*.clt)|*.clt|CHT cheat lists (*.cht)|*.cht"
 msgstr ""
 
-#: guiinit.cpp:256 panel.cpp:417
+#: guiinit.cpp:256 panel.cpp:424
 msgid "Loaded cheats"
 msgstr ""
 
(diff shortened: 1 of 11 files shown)


denisfa:
Looks good. Just one observation. While I was testing, sometimes the resize would be perfect and others would leave a gap on top and bottom (3x example for fail).

3x

No idea if there is something that we can do about this. Just found curious that it seems random.
Anyway, this isn't a major issue (probably just need to trigger a resize event somewhere).

rkitover:
@denisfa that isn't just it preserving the aspect ratio?

rkitover:
Well I tried with aspect ratio off and it also does not work, I think I need to call `Layout()` somewhere.

rkitover:
With aspect ratio on, and horizontal slack, it maximizes vertically correctly on menu bar hide.

With aspect ratio on, and vertical slack, it stays centered on menu bar hide/show.

Also noticed that turning aspect ratio off does not work without restarting, we should really fix that.

I added a `SendSizeEvent()`, no idea if this fixed anything.

Now with aspect ratio off, after initial startup, on menu hide there is vertical slack like in your screenshot.

Now if after this, you resize the window, then the menu bar hiding and showing works perfectly and the panel is maximized correctly.

At other times, the stretching just randomly stops working entirely, independently of the menu bar.

This is testing wx 3 on gtk2. I will test wx master on gtk3, and windows.

rkitover:
With wx master and gtk3 turning off preserve aspect ratio works even worse than on gtk2.

I've come to the conclusion that turning off preserve aspect ratio is the thing we need to fix, and the menu bar will then work correctly in that aspect.

Since most people will not turn that off, and the behavior is correct in that case, we can do this separately.

denisfa:
@rkitover I see. So let's fix the aspect ratio before moving on here. I will be able to help this week.

> Also noticed that turning aspect ratio off does not work without restarting, we should really fix that.

Do you have a list anywhere about all these issues/wishlist? You should probably put it on the repo, even if only as a `txt` file.

rkitover:
We could use issues with markdown checkboxes to track these things I guess.

rkitover:
I've opened #657 and #658 to track minor and major UI issues, respectively.

I will start a branch to fix aspect ratio for a separate PR.

rkitover:
When I rebase on top of the aspect ratio PR, there are no issues in either mode.

PR #659, opened 2020-04-29
Removing meaningless conditional

### LINKED ISSUE #578: Compiler warning: tautological comparison
During make -j
```
~/projects/visualboyadvance-m/src/gba/ereader.cpp:459:23: warning: comparison of unsigned expression >= 0 is always true [-Wtautological-compare]
        if ((reg[0].I >= 0) && (reg[0].I <= 0x10)) {
             ~~~~~~~~ ^  ~
1 warning generated.
```
This might be causing another issue.

operant:
reg[0].I is unsigned so checking if it is greater than or equal to zero is always true by definition.

Fixes #578


rkitover:
Thank you!

PR #660, opened 2020-05-01
Fix changing aspect ratio option.

rkitover:
When toggling "retain aspect ratio", destroy the drawing panel so it is
recreated with the desired aspect ratio settings.

Only add the vertical spacers when aspect ratio is on, they are not
needed when it's off.

Set the sizer priority of the DrawingPanel to 1 if aspect ratio is off
so that wxEXPAND always works.

Clear the GameArea sizer when attaching a new DrawingPanel, this is so
that the vertical spacers used when aspect ratio is on are not included
when aspect ratio is off.


rkitover:
@denisfa tested on linux and windows, I believe this will fix the aspect ratio issues we found in the hide menu bar pr.

PR #662, opened 2020-05-04
Added a boatload of range checks to ELF loader

kode54:
This should hopefully fix most possible attempts to crash the loader, I hope. Kindly shoot me for making this change from the Github editor.


kode54:
Disregard this, it breaks ELF loading even more. And it's not like it was working top notch to begin with. Load addresses need to be mapped into the GBA space, not just assumed to be in a tiny specific range.

ZachBacon:
@kode54 you still want to continue with this, or should I close the merge request?

PR #663, opened 2020-05-09
Turbo config refactor followup.

rkitover:
Turbo config refactor followup.

In the Turbo config dialog, remove the selection list for entering the
value and leave only the spin control. Allow values above 600%, up to
4000%.

Reintroduce the speedup_frame_skip config variable, defaulting to 9, use
it for turbo selection values > 600%, with speedup_throttle == 0.

The rationale for this is that on average modern hardware, throttle
values above 500% or 600% will not be effective.

The default is now shown as 1000%, which is:

frame_skip == 9 && speedup == 1,

where:

speedup == 1 is equivalent to throttle == 0,

as was the case before the turbo config changes.

Values above 600% are automatically rounded up or down to the nearest
100%, on entry and on click of the up/down arrows of the spin control.

The frame skip checkbox is cleared and disabled for the "Unlimited"
setting (throttle == 0), and set and disabled for values > 600%, to
reflect the mechanism to the user.

When the value again enters the modifiable range in the spin control,
the previous value of the checkbox is restored.

Misc:

- Turn off translation of percentage values in the xrc.

- Remove the size element for the throttle selection list in the general
  config dialog xrc, it breaks the layout on GTK3.

- Add a note about passing wxWidgets_CONFIG_EXECUTABLE to cmake to
  select wxWidgets version to README.md.

Hopefully this will reduce confusion and present a nicer UI.


rkitover:
@denisfa can you take a look please?

denisfa:
@rkitover Absolutely. I will as soon as possible.

denisfa:
@rkitover Sorry for the delay. I will be adding comments on the code now.

src/wx/xrc/GeneralConfig.xrc
```diff
@@ -111,31 +111,31 @@
...
+                    <item translate="0">375%</item>
+                    <item translate="0">400%</item>
+                    <item translate="0">425%</item>
+                    <item translate="0">450%</item>
+                    <item translate="0">475%</item>
+                    <item translate="0">500%</item>
+                    <item translate="0">525%</item>
+                    <item translate="0">550%</item>
+                    <item translate="0">575%</item>
+                    <item translate="0">600%</item>
```
denisfa, line 138: Instead of selecting values for throttle, could we just display a text box allowing the user to set any value? We could have a button to set to default value (`100%`). This is only an idea.

src/wx/xrc/GeneralConfig.xrc
```diff
@@ -111,31 +111,31 @@
...
-                    <item>400%</item>
-                    <item>425%</item>
-                    <item>450%</item>
-                    <item>475%</item>
-                    <item>500%</item>
-                    <item>525%</item>
-                    <item>550%</item>
-                    <item>575%</item>
-                    <item>600%</item>
+                    <item>Unlimited</item>
```
denisfa, line 114: I am having some issues with how this is being displayed on `GTK3, wxWidgets 3.0.1`.
Screenshot_2020-05-12_12-27-50. Removing this line fixes this.

denisfa:
> The default is now shown as 900%, as was the case before the turbo
config changes.

If the default was to skip 9 frames, then it was a `1000%` speedup. 9 frames skipped for each one shown.

rkitover:
> > The default is now shown as 900%, as was the case before the turbo
> > config changes.
> 
> If the default was to skip 9 frames, then it was a `1000%` speedup. 9 frames skipped for each one shown.

I will fix that.

> 
> > The frame skip checkbox is cleared and disabled for the "Unlimited"
> > setting, and set and disabled for values > 600%, to reflect the
> > mechanism to the user.
> 
> So there is an automatically change when throttle > 600%. We always use frame skip for these cases.
> Is this correct? Did I get it right?

Yeah, with > 600% throttle is set to zero, and frame skip is set appropriately.

src/wx/xrc/GeneralConfig.xrc
rkitover: Yeah I will replace it with a textbox with some sort of on-change handler to round up large values.

src/wx/xrc/GeneralConfig.xrc
rkitover: Will take a look.

rkitover:
@denisfa all fixed!

Your idea about not using a selection list was good, I have replaced it with a properly behaving spin control.

The default now shows 1000%.

Also fixed the GTK3 issue in the General config.

denisfa:
@rkitover Nice job! Looks great!

denisfa:
My comments on the code disappeared. I assume it is because of the force-push event. Do you know how can I approve without having to self-request another review?

rkitover:
I marked them as resolved, they should show up collapsed as "show resolved", they do for me.

rkitover:
Thank you very much for looking at this, given everything else that is happening.

My next project will be fixing the gamecube controller in windows, it works fine in linux. On windows it seems to trigger some kind of SDL bug. Before I thought it was not working because it was detected as a joystick and not as a game controller, but that is not the case, it is in fact a game controller.

Also I'm getting a switch pro controller in the next couple of days, and I will look into why that isn't working either.

PR #686, opened 2020-06-19
Follow up on #671.

denisfa:
@rkitover The tests should be:

- [ ] unicode key as accel/input (`Ç`)
- [ ] normal/ASCII key as accel/input (`L`)
- [ ] special chars as accel/input (`TAB`, `CTRL`, `BACKSPACE`, `ENTER`)
- [ ] keys combinations as accel (`CTRL+Ç`, `ALT+A`) (No reason for this as input, unless this is a regression for something I changed)
- [ ] set a default keybinding to another input/accel (`KP_ENTER`,  `KP_ADD`)
- [ ] set a key to both accel/input: the accel should have priority over the input

I decided to write this here so I can remember/document what I tested. Please edit/add if you see something missing.

### DIFF
--- a/src/gb/GB.cpp
+++ b/src/gb/GB.cpp
@@ -1085,14 +1085,6 @@ void gbWriteMemory(uint16_t address, uint8_t value)
                 gbSerialOn = 0;
                 gbMemory[0xff0f] = register_IF |= 8;
             }
-#ifdef OLD_GB_LINK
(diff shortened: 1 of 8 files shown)


denisfa:
@rkitover Sorry, this was a experiment branch that I pushed. I only want the last 2 commits from here.

PR #687, opened 2020-06-19
Input followup

denisfa:
@rkitover Now it is correct.

The tests should be:

- [x] unicode key as accel/input (`Ç`)
- [x] normal/ASCII key as accel/input (`L`)
- [x] special chars as accel/input (`TAB`, `CTRL`, `BACKSPACE`, `ENTER`)
- [x] keys combinations as accel (`CTRL+Ç`, `ALT+A`) (No reason for this as input, unless this is a regression for something I changed)
- [x] set a default keybinding to another input/accel (`KP_ENTER`,  `KP_ADD`)
- [x] set a key to both accel/input: the accel should have priority over the input
- [x] `SPACE` and `DELETE` also work as both game input and accels

I decided to write this here so I can remember/document what I tested. Please edit/add if you see something missing.

### DIFF
--- a/src/wx/guiinit.cpp
+++ b/src/wx/guiinit.cpp
@@ -2124,7 +2124,7 @@ public:
             if ((sys_accels[i].GetFlags() == selmod && sys_accels[i].GetKeyCode() == selkey)
                 || (seljoy != 0 && sys_accels[i].GetUkey() == selstr)) // joystick system bindings?
             {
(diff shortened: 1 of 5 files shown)


denisfa:
Ok, this should be it. Only had one issue: 

> set a default keybinding to another input/accel (KP_ENTER, KP_ADD)

On Windows, the label on the menu item would remain active even after I removed the default key from the shortcuts. I wasn't able to use it as input until I restarted the app, when it would work (and the label would not be there anymore).

One way or another, I still need to keep an eye on this for any Windows issues.

PR #692, opened 2020-06-23
Fix for non-stl wx builds.

rkitover:
Followup on baa0341b.

Use ToStdWstring() instead of wc_str() to look up strings in the map,
the key for the map is std::wstring.

I need to add this to the CI.

### DIFF
--- a/src/wx/wxutil.cpp
+++ b/src/wx/wxutil.cpp
@@ -77,16 +77,16 @@ KeyboardInputMap::KeyboardInputMap(){}
 void KeyboardInputMap::AddMap(wxString keyStr, int key, int mod)
 {
     KeyboardInputMap* singleton = getInstance();
-    singleton->keysMap[keyStr.wc_str()] = singleton->newPair(key, mod);
+    singleton->keysMap[keyStr.ToStdWstring()] = singleton->newPair(key, mod);
 }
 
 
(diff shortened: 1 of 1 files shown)


denisfa:
@rkitover Sorry for not seeing this earlier. It has been a busy week. What is a non-stl build? `wxWidgets 2.8`?

rkitover:
@denisfa 

It's wxWdigets built **without** `--enable-stl`, which changes the API and ABI.

Most distributions use `--enable-stl`, but some still don't, I noticed this on FreeBSD, with 3.1-gtk3.

I was thinking of some ways to test these in the CI, without rebuilding wxWidgets, which is very slow on 2 cores, I think I have some ideas now, I will test later today.

rkitover:
@denisfa 

As it turns out, I was wrong about this.

The option on FreeBSD which caused the failure is `--enable-utf8`, Linux distribution don't seem to use this generally, probably for compatibility with Windows builds which generally don't, because it makes no sense on Windows where the Unicode format is usually UTF-16.

As of 36e61d76 both the `--enable-utf8` and `--disable-stl` variants are being tested in the CI on Travis.

I tested for the fail by reverting this commit.

The other thing I promised to do was add a unit testing framework like cppunit, I'm going to do some research on this now.

denisfa:
@rkitover Ok, I will keep an eye on this as well.

Issue #689, opened 2020-06-22
Misbehaving arrow keys


EpicWink:
About a year ago (or maybe a little more), there was a bug (in Linux I think) where the emulator would intermittently not register my arrow key release. This was fixed in an input refactor back then (I think near or at the same time the input latency on Linux was significantly improved).

It now seems to have cropped up since the recent input refactor. In addition, some of my inputs seem to be dropped. My most recent successful commit tested is 0c5ee99 (although there could have been newer that I haven't tested).

### ROM details
* Title: Tetris DX (GBC)
* Maker: 01 Nintendo
* Unit: 03
* Cartridge: 03
* Version 00

### Option details
Video: OpenGL, xBRZ 6x filter, no blending
Sound: SDL, 2-frame buffer

### Build details
OS: Ubuntu 19.10
Graphics: Xorg, Nvidia 435.21, GLX 1.4, OpenGL 4.6.0, ES 3.2
Build: 3a2b97f (current master)

<details>
<summary>cmake -L (click to expand)</summary>

```
-- Found SFML 2.5.1 in /usr/include
-- Could NOT find FFmpeg (missing: AVCODEC_LIBRARIES AVCODEC_INCLUDE_DIRS AVFORMAT_LIBRARIES AVFORMAT_INCLUDE_DIRS SWSCALE_LIBRARIES SWSCALE_INCLUDE_DIRS AVUTIL_LIBRARIES AVUTIL_INCLUDE_DIRS SWRESAMPLE_LIBRARIES SWRESAMPLE_INCLUDE_DIRS) 
-- Found gcc-ar: /usr/bin/gcc-ar
-- Found gcc-nm: /usr/bin/gcc-nm
-- Found gcc-ranlib: /usr/bin/gcc-ranlib
-- Configuring done
-- Generating done
-- Build files have been written to: ~/src/visualboyadvance-m/build
-- Cache values
CCACHE_FOUND:FILEPATH=/usr/bin/ccache
CMAKE_BUILD_TYPE:STRING=Release
CMAKE_INSTALL_PREFIX:PATH=/usr/local
ENABLE_ASAN:BOOL=OFF
ENABLE_ASM:BOOL=OFF
ENABLE_ASM_CORE:BOOL=OFF
ENABLE_ASM_SCALERS:BOOL=OFF
ENABLE_DEBUGGER:BOOL=OFF
ENABLE_FAUDIO:BOOL=OFF
[...]

PR #694, opened 2020-06-28
Check game kbd input with wxGetKeyState() again.

rkitover:
Followup on baa0341b.

Reintroduce checking if a key is actually pressed in the panel OnKeyDown
event with wxGetKeyState(key_code).

Introduced in b0ec846 and removed in baa0341b.

wxGetKeyState() does not work on Wayland and does not work for Unicode
keys, support for which was introduced in baa0341b, which is why the
call was removed.

Add two static functions is_key_pressed(ev) and is_key_released(ev) that
return true under Wayland or if the key is Unicode, and call
wxGetKeyState() otherwise.

The reason this call was introduced in b0ec846 was to work around a bug
in some Linux distributions that caused spurious keyboard events to be
generated. The bug apparently still persists, see #689.

Because the two additional checks were necessarily added, the bug may
persist on Wayland, if it exists there, and for Unicode keys, but will
be fixed in the more common case of Latin keys under Xorg.

### DIFF
--- a/src/wx/panel.cpp
+++ b/src/wx/panel.cpp
@@ -17,6 +17,7 @@
 #include "filters.h"
 #include "wxvbam.h"
 #include "wxutil.h"
+#include "wayland.h"
 
 #ifdef __WXMSW__
 #include <windows.h>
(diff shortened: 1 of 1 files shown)


rkitover:
Tested on Linux with both Latin and Unicode game keys.

denisfa:
Since the user already tested this and everything works, it should be enough for now.

PR #697, opened 2020-07-08
Add Catch2 unit testing framework + some tests.

rkitover:
Add the Catch2 headers to third_party/include/catch2.

Add Catch.cmake and CatchAddTests.cmake to cmake/.

Add unit tests src/wx/tests/strutils.cpp using Catch2 to test
src/wx/strutils.cpp.

Make some code changes to strutils.cpp to make the tests pass.

See src/wx/tests/CMakeLists.txt for how to set up unit test files; they
plug into the normal CTest mechanism in cmake.

Start running ctest on travis.

Disable tests for Win32 because there seem to be compilation issues with
Visual Studio that need to be looked at later.

Move stb to third_party/include where catch2 is now as well.


denisfa:
@rkitover This looks like a really good start for testing code. The only one I know for C++ is `gtest` (since it is the only one that I actively interact, although I use java and python on my current job). Is this plug and play already, eg adding a tests folder and we just create tests there?

Could you wait a few days here? I should be able to review this soon. If this is locking something for you, then I approve and ask questions later.

rkitover:
@denisfa Of course, take your time.

Yeah the point of this was to add some stuff to start being able to add tests, you can look in `src/wx/tests` in the `CMakeLists.txt` for how to add new test binaries with the objects they depend on, and `src/wx/tests/strutils.cpp` for the specific tests I wrote here as a start.

rkitover:
I made some of improvements

- Tests can be disabled with `-DBUILD_TESTING=OFF`.
- Test binaries are written to `build/tests`.
- The Visual Studio compilation issue is fixed, was a minor problem.
- Tests run on appveyor too now.

denisfa:
I feel like it is a good opportunity to learn something new here. It looks good.

rkitover:
I went back to your idea of using a parameter for `str_split()` to return the separator, just made it have a default of false and made `str_split_with_sep()` a wrapper around it.

rkitover:
The nice thing about this stuff, is I can change a function like this, and not have to do a lot of laborious GUI functionality testing, because I have a test that proves that it works!

That's the main idea.

We need to take it further, maybe steal the mGBA emulator tests, etc..

PR #700, opened 2020-07-15
Separate thread for main loop.

denisfa:
Here it is. I will still need to do some cleanup, but at least you can see and check/test how this is going. I will need all commits, despite my description, but I will squash them later.

The good thing here is to no longer pause while browsing menus on Windows, others may include actually being able to maneuver the framerate.

### DIFF
--- a/src/wx/panel.cpp
+++ b/src/wx/panel.cpp
@@ -28,41 +28,6 @@ static void clear_input_press();
 
 int emulating;
 
-IMPLEMENT_DYNAMIC_CLASS(GameArea, wxPanel)
-
-GameArea::GameArea()
-    : wxPanel()
-    , panel(NULL)
-    , emusys(NULL)
-    , was_paused(false)
-    , rewind_time(0)
-    , do_rewind(false)
-    , rewind_mem(0)
-    , num_rewind_states(0)
-    , loaded(IMAGE_UNKNOWN)
-    , basic_width(GBAWidth)
-    , basic_height(GBAHeight)
-    , fullscreen(false)
-    , paused(false)
-    , pointer_blanked(false)
-    , mouse_active_time(0)
-{
-    SetSizer(new wxBoxSizer(wxVERTICAL));
-    // all renderers prefer 32-bit
-    // well, "simple" prefers 24-bit, but that's not available for filters
-    systemColorDepth = 32;
-    hq2x_init(32);
-    Init_2xSaI(32);
-
-#ifndef NO_THREAD_MAINLOOP
-    Bind(WX_THREAD_REQUEST_UPDATEDRAWPANEL, &GameArea::RequestUpdateDrawPanel, this);
-    Bind(WX_THREAD_REQUEST_DRAWFRAME, &GameArea::RequestDrawFrame, this);
-    Bind(WX_THREAD_REQUEST_UPDATESTATUSBAR, &GameArea::RequestUpdateStatusBar, this);
-    Bind(WX_THREAD_REQUEST_GBPRINTER, &GameArea::ShowPrinter, this);
-    Bind(WX_THREAD_REQUEST_UPDATELOG, &GameArea::RequestUpdateLog, this);
-#endif
-}
-
 void GameArea::LoadGame(const wxString& name)
 {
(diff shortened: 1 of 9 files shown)


src/wx/panel.cpp
```diff
@@ -973,6 +961,307 @@ void GameArea::Resume()
...
+                wxGetApp().frame->PollJoysticks();
+                if (!panel) {
+                    wxQueueEvent(this, new wxThreadEvent(WX_THREAD_REQUEST_UPDATEDRAWPANEL));
+                }
+                if (!paused && panel) {
+                    emusys->emuMain(emusys->emuCount);
+                }
+	    }
+        }
+        wxMilliSleep(1);
```
denisfa, line 1032: @rkitover Here we can play a little bit about how to do proper timesteps and other controls. We can try to control when and for how long we should sleep, like an actual main loop for a game.

src/wx/panel.cpp
```diff
@@ -973,6 +961,307 @@ void GameArea::Resume()
...
+    wxQueueEvent(this, event);
+}
+
+struct PrinterDataDialog {
+    uint16_t* to_print;
+    uint16_t** accum_prdata;
+    int lines, feed;
+    int *accum_prdata_len, *accum_prdata_size;
+};
+PrinterDataDialog printerDataDialog;
```
denisfa, line 1169: @rkitover YIKES alert here. If you know something that could help me here with wxWidgets, I would appreciate it very much.

src/wx/wxvbam.cpp
```diff
@@ -772,6 +775,10 @@ void MainFrame::OnDropFile(wxDropFilesEvent& event)
     wxString* f = event.GetFiles();
     // ignore all but last
     wxGetApp().pending_load = f[event.GetNumberOfFiles() - 1];
+
+#ifndef NO_THREAD_MAINLOOP
+    GetPanel()->LoadGame(wxGetApp().pending_load);
+#endif
```
denisfa, line 781: @rkitover If I understood correctly, the IDLE event main loop keeps checking if this is set to load a game and start emulation. I would like to keep this idea away from the thread solution, if possible.

rkitover:
Will start going through this later today, sorry been a rough few days for me.

src/wx/wxvbam.h
```diff
@@ -500,11 +504,42 @@ class DrawingPanelBase;
...
+                 , public wxThreadHelper
+#endif
+{
 public:
     GameArea();
     virtual ~GameArea();
 
+#ifndef NO_THREAD_MAINLOOP
+    virtual wxThread::ExitCode Entry();
+    void StartEmulationThread();
```
rkitover, line 528: Where are these methods defined? Did you forget to include a file in your commit?

rkitover: Sorry I didn't see that some sections of the github diff view need to be expanded.

rkitover: No they are actually not there, I misunderstood the point of the expand buttons.

denisfa:
@rkitover Glad to see you are back. This is the skeleton of the changes, I have some more to add here. I will focus on just isolating the separate thread emulation. I should put something this weekend (I have been busy with work).

src/wx/panel.cpp
```diff
@@ -2290,12 +2290,12 @@ void GLDrawingPanel::DrawingPanelInit()
     }
 #elif defined(__WXMSW__)
     typedef char* (*wglext)();
-    wglext wglGetExtensionsString = (wglext)wglGetProcAddress("wglGetExtensionsString");
-    if (wglGetExtensionsString == NULL || strstr(wglGetExtensionsString(), "WGL_EXT_swap_control") == 0) {
-        if (wglGetExtensionsString == NULL)
-            systemScreenMessage(_("No support for wglGetExtensionsString"));
-        else
-            systemScreenMessage(_("No support for WGL_EXT_swap_control"));
+    wglext wglGetExtensionsStringEXT = (wglext)wglGetProcAddress("wglGetExtensionsStringEXT");
```
rkitover, line 2584: I cherry-picked this to master, if you don't mind.

PR #705, opened 2020-07-25
Allow background input without window focus, closes #207

### LINKED ISSUE #207: Can't background input in VBA-M 2.0.2
I'm using VBA-M 2.0.2 to play TLoZ: Four Swords Adventures with my friend but VBA-M doesn't accept the background input although I unchecked **Pause when inactive**

v1993:
Fixed this bug since it was affecting me and preventing two-player game in Dolphin.

### DIFF
--- a/src/wx/widgets/sdljoy.cpp
+++ b/src/wx/widgets/sdljoy.cpp
@@ -104,9 +104,11 @@ void wxSDLJoy::Poll()
                         ev.prev_val      = prev_val;
 
                         handler->ProcessEvent(ev);
+                    }
 
-                        joystate[joy].axis[axis] = val;
+                    joystate[joy].axis[axis] = val;
 
+                    if (val != prev_val) {
                         wxLogDebug("GOT SDL_CONTROLLERAXISMOTION: joy:%d axis:%d val:%d prev_val:%d", joy, axis, val, prev_val);
                     }
                 }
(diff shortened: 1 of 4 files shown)


v1993:
Oops, for now this break joystick input outside of input setup window and accelerators at all. I'm kinda confused right now about how input is handled but will look  for solution.

denisfa:
@v1993 Thank you for this. Unfortunately, I did not see this earlier and I was working on a solution. Would you test using the branch `background-input`, please? I believe it should work as you want.

src/wx/wxvbam.cpp
```diff
@@ -722,6 +722,7 @@ MainFrame::MainFrame()
 {
     jpoll = new JoystickPoller();
     this->Connect(wxID_ANY, wxEVT_SHOW, wxShowEventHandler(JoystickPoller::ShowDialog), jpoll, jpoll);
+    AttachJoystickHandler(this);
```
denisfa, line 725: @v1993 You had the right idea here, It is confusing that it is not the frame that get the input, but the panel (GameArea and DrawingPanelBase classes). After a game is loaded, we Bind from the DrawingPanel to the GameArea functions and on some of them we call `Skip`, which then go to `filterEvents` on our `MainFrame`.

src/wx/widgets/sdljoy.cpp
```diff
@@ -104,9 +104,11 @@ void wxSDLJoy::Poll()
                         ev.prev_val      = prev_val;
 
                         handler->ProcessEvent(ev);
+                    }
```
rkitover, line 107: Does this change anything in terms of functionality? If not, please remove it.

src/wx/wxvbam.h
```diff
@@ -397,6 +399,7 @@ class JoystickPoller : public wxTimer {
         void Notify() {
             wxGetApp().frame->PollJoysticks();
         }
+        // This isn't called for dialogs but actually only for main window. Is this intended?
```
rkitover, line 402: The previous implementation only used this for dialogs, since the OnIdle loop in the panel does the polling itself, feel free to change the method name to `ShowWindow` or some such. That would be clearer.

rkitover:
@denisfa sorry just saw that you were already fixing this, well no matter, we will merge whichever branch you prefer for this.

rkitover:
Superseded by #706.

PR #706, opened 2020-07-27
Input while on background.

denisfa:
@rkitover 

For joystick we always send the event to our app. This is actually a regression, since an older version allowed it.

For keyboard, I made an option to toggle it. It uses a separate thread and a couple of functions not directly exposed by wxWidgets to convert native keycodes to WX ones.

- For Linux, Wayland is not supported, as stated on commit message due to security concerns. Using XLib, we have essentially a skeleton of a keylogger, which is not exactly a good thing but it is the only way to do it.
- For Windows, it depends on some system configuration. There is a way to block `GetAsyncKeyState` that is by default disabled on desktops (but enabled on Servers, it seems).
- For MacOS I have no way to test any work, so I left it out on purpose.

I am a little bit interested on alternative solutions for the joystick, so I will see what that user can offer for joystick background input, if it is better than what we have here, I will take care of conflicts.

Please take a look and see if there is something that should be fixed or can be improved.

EDIT1: I will add the `- Fix #ISSUE.` on the merge commit.


src/wx/background-input.cpp
```diff
@@ -0,0 +1,2331 @@
...
+    if (x11display) {
+        XCloseDisplay(x11display);
+        x11display = nullptr;
+    }
+#endif
+}
+
+wxThread::ExitCode BackgroundInput::Entry()
+{
+    if (Setup())
```
rkitover, line 2300: Minor nitpick, feel free to ignore, I would write this as:

```cpp
if (Setup() != 0)
```

otherwise it looks like you are testing a boolean, which makes it more confusing.

src/wx/background-input.cpp
```diff
@@ -0,0 +1,2331 @@
...
+        return (wxThread::ExitCode)1;
+
+    while (!TestDestroy()) {
+        if (CheckKeyboard()) break;
```
rkitover, line 2304: Here likewise.

rkitover:
Ignore the CI fails, the travis ones were because of a stale ccache, I restarted those, the appveyor ones were something to do with deps in vcpkg, I restarted those as well.

denisfa:
Thank you! This was indeed very troublesome, but I learned a lot doing this. I will put the finishing touches soon, I am just short of reinstalling everything on my side (desktop went BOOM).

src/wx/background-input.cpp
```diff
@@ -0,0 +1,2331 @@
...
+#include "wayland.h"
+
+/* The following functions are copied from wxWidgets repo file
+ * `src/x11/utils.cpp` that convert a XLib keycode into a
+ * WX keycode.
+ */
+int wxCharCodeXToWX(int keySym);
+int wxUnicodeCharXToWX(int keySym);
+
+static const struct {
```
rkitover, line 268: Maybe just declare this as `std::unordered_map<unsigned, int>`, then access would be direct and the initialization syntax can remain the same.

rkitover: Look at my joystick code for how I use that.

rkitover: Also, we should probably put the map definitions into a header file, just the definitions, then we can later generate the file if needed.

denisfa:
An update on this: I was having some issues grabbing specific X11 keys, such as `KP_SEPARATOR`, `XK_Super_L` and some other cases. I will update this solution to try something. We should expect users reporting specific keys not working on background input, at least on linux.

src/wx/background-input.cpp
```diff
@@ -0,0 +1,2341 @@
...
+#include <X11/keysym.h>
+#include <X11/XKBlib.h>
+#include "wayland.h"
+
+/* The following functions are copied from wxWidgets repo file
+ * `src/x11/utils.cpp` that convert a XLib keycode into a
+ * WX keycode.
+ */
+std::unordered_map<unsigned, int> x11KeySym = {
+    {0x0020, 0x0020},    // space
```
rkitover, line 260: I think we should put this map into a header file, so it could be generated later.

Declare it like this:

```cpp
std::unordered_map<unsigned, int> x11KeySym = {
#include "xkeymap.h"
}
```

Then in `xkeymap.h` you would have just the definitions by themselves:

```cpp
    {0x0020, 0x0020},    // space
    {0x0021, 0x0021},    // exclam
...
```

This will also make the file easier to edit, as a bonus.

The windows keymap is relatively small and can be left as-is.

denisfa:
@rkitover Considering the variable `allowBackgroundInput`, what do you think about also using it for the joystick background input? The previous behaviour (the regression on this commit) was to always send to our app.

I believe I have checked all your comments this time, although I am still unable to test it on MacOS.

rkitover:
> @rkitover Considering the variable `allowBackgroundInput`, what do you think about also using it for the joystick background input? The previous behaviour (the regression on this commit) was to always send to our app.

That is a great idea, I suggest encapsulating this logic in an app method, for greater separation of concerns:

```cpp
    auto handler = wxGetApp().GetJoystickHandler();
    if (!handler) return;
```

This could do the checking for whether we are in focus or the option is enabled etc..

We can also remove the `evthandler` member of `wxSDLJoy`.

> 
> I believe I have checked all your comments this time, although I am still unable to test it on MacOS.

Everything looks awesome, if I can get my mac VM working again, I will do a separate PR to extend your system for it.

rkitover:
Actually now that I think about this some more, we can also deactivate the joystick polling when we don't need it.

The panel `OnIdle` would not call the poll method if not emulating.

The app would stop the timer when not in focus (unless the option is enabled) and restart it when regaining focus.

denisfa:
> Actually now that I think about this some more, we can also deactivate the joystick polling when we don't need it.
> The app would stop the timer when not in focus (unless the option is enabled) and restart it when regaining focus.

This is a little bit tricky due to how the timer is needed for possible accelerators. I don't think we really should do this now. I will add this to my list on the separate thread PR. I can see how to handle this there.

> The panel `OnIdle` would not call the poll method if not emulating.

It would also not call if is paused, since we need the  `event.RequestMore` for it to run another loop and try `PollJoystick` again.

rkitover:
Please ignore appveyor failures, I am having problems with that right now that I already spent hours trying to fix, but I will hopefully fix it soon. Looking at your changes now.

rkitover:
All clear from me, unless there was anything else you wanted to do. Though I still think an `app.GetJoystickHandler()` method would be good.

rkitover:
Hi @denisfa,

I hope you are doing well!

I'm sorry it took me so long to merge this, I wanted to do a minor refactor on top related to the joystick event handler, but various more pressing issues kept popping up, with the windows build and CI, etc. etc..

Anyway, I added my minor refactor in 2a5acc32, and this is finally merged.

Take care.

PR #712, opened 2020-08-10
Addressing audio plugin crash

briansrls:
Stereo buffer is not remade in the case of audio plugin
switch (only affects GBA, not GB).  This results in a crash.
We need to remake the stereo buffer and reinit the pointers
in Gba_Pcm.output or they will be dangling, and lead to a
crash.  Also, cleaning up the Multi_Buffer class a bit.

### DIFF
--- a/src/apu/Multi_Buffer.h
+++ b/src/apu/Multi_Buffer.h
@@ -35,11 +35,12 @@ class Multi_Buffer
         };
         virtual channel_t channel(int index) BLARGG_PURE(;)
 
-            // See Blip_Buffer.h
-            virtual blargg_err_t set_sample_rate(long rate, int msec = blip_default_length)
-                BLARGG_PURE(;) virtual void clock_rate(long)
-                    BLARGG_PURE({}) virtual void bass_freq(int) BLARGG_PURE({}) virtual void clear()
-                        BLARGG_PURE({}) long sample_rate() const;
+        // See Blip_Buffer.h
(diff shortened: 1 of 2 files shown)


negativeExponent:
soundReset() is suppose to be called after soundInit(). dunno how the standalone plugins are loaded, but if its calling soundInit, then its missing soundReset then.

this should work as well though, other than duplicated func calls.

briansrls:
> soundReset() is suppose to be called after soundInit(). dunno how the standalone plugins are loaded, but if its calling soundInit, then its missing soundReset then.
> 
> this should work as well though, other than duplicated func calls.

The relevant code is at cmdevents.cpp:2689, it's calling soundInit without calling soundReset.  Is it safer to put a call to soundReset within soundInit, assuming there is no reason to call soundInit without soundReset?

rkitover:
@briansrls 

I think this is fine, especially since you only call the function if the buffer does not exist.

`soundReset()` does other things which we may not always want.

And not having memory errors regardless of the API call sequence is a positive IMHO.

Are you ready for this to be merged?

Please ignore the appveyor failure, there is an issue with it right now.

briansrls:
@rkitover yes, please merge it when you can.  Thanks

negativeExponent:
a full init without the reset? ok fine. no further comment. more band-aid fixes

briansrls:
#713

rkitover:
@briansrls if you want the automatic commit -> issue closing behavior, you have to put a line like this in your commit message next time:

```
- Fix #713.
```

briansrls:
Thanks, I'll try it next time

Issue #711, opened 2020-08-09
[Input] Cannot remap inputs with gamepad


negativeExponent:
### Expected behavior
Should be able to remap gamepad buttons
### Actual behavior
No effect
### Steps to reproduce the behavior
- goto options > input > configure
- clear and try to set new binds
- nothing happens with gamepad, keyboard binds works fine. the default gamepad presets also works but remapping any does not. 

Conditions:
- The issue happens when remapping while a game is already loaded.
- **Emulation** > **Pause when inactive** should be **checked**

As seen in screenshot, inputs are detected from the log window.
![1596962525_2020_08_09_16:42:05](https://user-images.githubusercontent.com/54053706/89728465-1eff8800-da60-11ea-9ef7-878caf768347.png)

**First bad commit**: https://github.com/visualboyadvance-m/visualboyadvance-m/commit/baa0341bd52aa96db9eb9dde37b20c50ef3f48b6

### ROM details

Which games are affected by the issue (please state if GBA or GB game): (not game related)

Which games, if any, are NOT affected by the issue (please state if GBA or GB game): (not game related)

Include screenshots of ROM Information (File > ROM Information): (not game related)

If you are using a BIOS file, please say which one: (irrelevant in this case)

### Option details

Video driver (simple, opengl): opengl

Sound driver (directsound, xaudio, openal, SDL): sdl

### Build details

Operating System (Windows, Mac, Linux (state distribution), etc.): Artix (Arch-based_

If using Linux, specify if using xorg or Wayland: xorg (running rootless if that matters)

Version of code (pre-built binary version, or commit ref, or just "master"): master

PR #716, opened 2020-08-12
Add back joystick polling while processing GUI event.

briansrls:
Also, add define for num joypads instead of magic number

### DIFF
--- a/src/wx/guiinit.cpp
+++ b/src/wx/guiinit.cpp
@@ -1693,7 +1693,7 @@ public:
             }
         }
     }
-} JoyPadConfigHandler[4];
+} JoyPadConfigHandler[NUM_JOYPADS];
 
 // manage fullscreen mode widget
 // technically, it's more than a validator: it modifies the widget as well
@@ -3791,7 +3791,7 @@ bool MainFrame::BindControls()
         wxDialog* joyDialog = LoadXRCropertySheetDialog("JoypadConfig");
         wxFarRadio* r = 0;
 
-        for (int i = 0; i < 4; i++) {
+        for (int i = 0; i < NUM_JOYPADS; i++) {
             wxString pn;
             // NOTE: wx2.9.1 behaves differently for referenced nodes
             // than 2.8!  Unless there is an actual child node, the ID field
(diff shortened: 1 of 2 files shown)


briansrls:
Fixes #711

rkitover:
@denisfa we may have broken this in the input refactor, can you work with him on this.

@briansrls also you have to delete objects created with new, otherwise you have a memory leak. But more importantly, we already have a `JoystickPoller` running for the main app, because since the input refactor denisfa did we support joystick events for accelerators.

rkitover:
@denisfa the issue here may be that with a game loaded, the poller is off, but with the config dialog open, the OnIdle calls to poll aren't running either.

briansrls:
> @denisfa we may have broken this in the input refactor, can you work with him on this.
> 
> @briansrls also you have to delete objects created with new, otherwise you have a memory leak. But more importantly, we already have a `JoystickPoller` running for the main app, because since the input refactor denisfa did we support joystick events for accelerators.

@rkitover you're right, that was a silly mistake.  I can fix the memory leak, but from reading your comment I understand that there may be a more appropriate solution.  I don't understand the infrastructure of this program well enough to know, but I hope this PR highlights something important that has changed.

briansrls:
Hi @rkitover, I looked into what you suggested, here is the behavior I'm seeing:

OnIdle is called while no game is running, both without the config dialogue open and with it open.
OnIdle is called while the game is running, unless the config dialogue is open.  For some reason the calls stop when config dialogue is opened.

I'm trying to figure out why now.

rkitover:
Well the dialogs are modal, which means they are supposed to stop processing in other windows of the app, the first scenario you describe is strange.

briansrls:
> Well the dialogs are modal, which means they are supposed to stop processing in other windows of the app, the first scenario you describe is strange.

I was able to confirm this behavior via gdb and also logging.  I thought the first scenario was correct, because don't we have OnIdle calls to a poller?  Right now config via joystick is working when a game isn't running. and I think it's working because of the OnIdle polling calls.  If we need to disallow OnIdle calls while the dialogue is open, it sounds like we need a new place for the joystick polling for input config.  Does that sound right?

rkitover:
The timer runs when a game isn't loaded, I'm going to fix this now anyway.

PR #736, opened 2020-09-02
Remove bilinear filter

briansrls:
This change removes bilinear filter from the Video menu, It's already
available in the OpenGL plugin menu in the filter dropdown.  The
option present in the normal video dropdown doesn't work well,
so we should remove it.

### DIFF
--- a/po/wxvbam/wxvbam.pot
+++ b/po/wxvbam/wxvbam.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2020-08-29 20:11+0000\n"
+"POT-Creation-Date: 2020-09-01 18:22-0700\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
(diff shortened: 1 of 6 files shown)


briansrls:
A effect of this change is that bilinear filter will no longer be enabled by default, but I welcome this change because I never liked the filter anyway

negativeExponent:
removing a feature because "you dont like it" doesn't sound like reason enough. In any case, both filter does not do the same thing. one applies soft-filters to gba's frame buffer while one applies to video rendering which can be used ontop of any filters you enable.

briansrls:
Hi @negativeExponent, my intention wasn't to remove the feature because I didn't like it, I agree that is not a good reason.  I was only saying that I didn't like that it was enabled by default.

We have a bug open on this feature because it isn't working as intended, when I looked into it it seemed redundant with the Bilinear filter feature in the list of textures.  However, after your explanation it sounds like it's not actually redundant.

This is a little confusing, because we intended to move this filter to the OpenGL plugin menu, but it will be very confusing to have a checkbox for Bilinear filter, and also have it present in the dropdown menu.

Here is the bug for reference:  #736

negativeExponent:
then you are missing a re-init. you need to call GLDrawingPanel::DrawingPanelInit() when Video>Bilinear filter is changed.

briansrls:
@negativeExponent That would fix the issue, but the option doesn't belong here.  It's only implemented for OpenGL, so it needs to be moved there.  Currently the position it's in implies it works for all plugins.

briansrls:
Looks like this fix isn't right, thanks to @negativeExponent for clarifying.

PR #740, opened 2020-09-05
Force a panel update when configuring bilinear.

### LINKED ISSUE #727: Billinear filter does not toggle until you open/close the video plugin menu
### Expected behavior
Bilinear filter should immediately take effect when toggled through the menu.

### Actual behavior
Bilinear filter does not seem to take effect until the video plugin menu opens.  Also, bilinear filter doesn't seem to work with the "Simple" video driver?

Toggle bilinear filter.

Video driver (simple, opengl):
opengl

### Build details

Windows

master

briansrls:
- Fix #727

### DIFF
--- a/src/wx/cmdevents.cpp
+++ b/src/wx/cmdevents.cpp
@@ -2872,6 +2872,11 @@ EVT_HANDLER(wxID_ABOUT, "About...")
 EVT_HANDLER(Bilinear, "Use bilinear filter with 3d renderer")
 {
     GetMenuOptionBool("Bilinear", gopts.bilinear);
+    // Force new panel with new bilinear option
+    if (panel->panel) {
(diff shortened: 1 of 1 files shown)


src/wx/cmdevents.cpp
```diff
@@ -2872,6 +2872,11 @@ EVT_HANDLER(wxID_ABOUT, "About...")
 EVT_HANDLER(Bilinear, "Use bilinear filter with 3d renderer")
 {
     GetMenuOptionBool("Bilinear", gopts.bilinear);
+    // Force new panel with new bilinear option
+    if (panel->panel) {
```
briansrls, line 2876: It's not necessary to check if the option has changed, because this code is handling a change event.

PR #741, opened 2020-09-05
Initial support for GC and Switch controllers.

rkitover:
Add the capability to treat some GameController devices as joysticks,
because they only send joystick events.

Make a set of names, as returned by SDL, with the names for the Switch
Pro controller and the Mayflash GameCube adapter and use them as
joysticks, filtering out GameController events for them.

Do some minor refactoring and improvements as well, like using the
`is_gc` bool member of state instead of calling
`SDL_IsGameController()`.

### DIFF
--- a/src/wx/widgets/sdljoy.cpp
+++ b/src/wx/widgets/sdljoy.cpp
@@ -1,10 +1,17 @@
 #include <cstddef>
+#include <unordered_set>
 #include "wxvbam.h"
 #include "wx/sdljoy.h"
 #include "SDL.h"
 #include <SDL_events.h>
 #include "../common/range.hpp"
 #include "../common/contains.h"
+
+// Use these as a joystick and not as a GameController.
+static const std::unordered_set<std::string> JOYSTICK_OVERRIDES{
+    "Nintendo Switch Pro Controller",
+    "Mayflash GameCube Controller Adapter"
+};
 
 using namespace Range;
 
@@ -70,6 +77,10 @@ void wxSDLJoy::Poll()
 
                 if (contains(instance_map, joy)) {
                     auto& state   = *(instance_map[joy]);
+
+                    if (!state.is_gc)
+                        break;
+
                     auto but      = e.cbutton.button;
                     auto val      = e.cbutton.state;
                     auto prev_val = state.button[but];
@@ -93,6 +104,10 @@ void wxSDLJoy::Poll()
 
                 if (contains(instance_map, joy)) {
                     auto& state   = *(instance_map[joy]);
+
+                    if (!state.is_gc)
+                        break;
+
                     auto axis     = e.caxis.axis;
                     auto val      = axisval(e.caxis.value);
                     auto prev_val = state.axis[axis];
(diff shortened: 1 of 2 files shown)


rkitover:
@briansrls more SDL crap for you to look at!

rkitover:
We might have to do axes mappings, but I will save that for later.

src/wx/widgets/sdljoy.cpp
```diff
@@ -312,19 +309,37 @@ void wxSDLJoy::ConnectController(uint8_t joy)
...
         if (dev) {
-            joystate[joy].dev = dev;
-
+            name   = SDL_GameControllerName(dev);
             js_dev = SDL_GameControllerGetJoystick(dev);
+
+            std::string extra_msg;
+
+            if (contains(JOYSTICK_OVERRIDES, name)) {
+                is_gc             = false;
```
briansrls, line 324: What does this logic mean?  Open as a gamecontroller, and then treat it as a joystick.

briansrls: Why is it necessary for these types of controllers?

rkitover: As I said in the commit message, they only send joystick events.

briansrls: Could you just open it as a joystick then?  For example:

If !contains(JOYSTICK_OVERRIDES, name) && IsGameController

else, open as a joystick

seems like the logic will be simpler

briansrls: I think I answered my own question, you're getting the name after it's open.  Maybe it's not possible to do what I'm saying.

briansrls (commented):
You would have to use this function to do what I'm proposing:

SDL_GameControllerNameForIndex(i)

src/wx/widgets/sdljoy.cpp
briansrls: For some reason it put this in another comment.  You would need to instead use this function, so you don't have to open the controller just to find the name:

SDL_GameControllerNameForIndex(i)

But, I'll leave that decision up to you.

rkitover: For some reason I did not get emails for your comments, I will make this change.

rkitover:
The MayFlash works on Linux without the joystick override under a different name, the Switch Pro does not work at all.

ZachBacon:
@rkitover I can test joycons and my 8bitdo sn30+ in switch mode to be sure if you want, add some extra validation.

rkitover:
@briansrls I tried using `SDL_JoystickNameForIndex()` and opening it as a joystick, but it did not work. It looks like I have to use my method of opening it as a GC for it to work at all.

There is another problem with connecting it without the input dialog when it's already configured, I need to look at that before I think of merging this.

ZachBacon:
@rkitover @briansrls any update on this? I'm going to try and clean up a few things in the Pull Request cue before I get back into the run of things, especially now that I got my hands on a lease for an m1 macbook

PR #742, opened 2020-09-06
Refactor wxSDLJoy

Steelskin:
This is a major refactor of the wxSDLJoy class.
* Move handling of SDL objects creation and destruction to its own
  class. This simplifies the lifespan of SDL-related objects.
* Re-add handling of HATs. This is necessary for DirectInput
  controllers.
* Fix a crash on joystick disconnect.
* Rename the public API for wxSDLJoy to be clearer.
* Add documentation for every class related to wxSDLJoy.
* Change some tabs to white spaces in wxvbam.cpp.

### DIFF
--- a/src/wx/widgets/sdljoy.cpp
+++ b/src/wx/widgets/sdljoy.cpp
@@ -1,35 +1,36 @@
-#include <cstddef>
-#include "wxvbam.h"
 #include "wx/sdljoy.h"
-#include "SDL.h"
-#include <SDL_events.h>
-#include "../common/range.hpp"
+
+#include <algorithm>
+#include <wx/timer.h>
+#include <SDL.h>
+
 #include "../common/contains.h"
+#include "../wxvbam.h"
 
-using namespace Range;
+namespace {
 
-// For testing a GameController as a Joystick:
-//#define SDL_IsGameController(x) false
-
-DEFINE_EVENT_TYPE(wxEVT_SDLJOY)
-
-wxSDLJoy::wxSDLJoy()
-    : wxTimer()
-{
-    // Start up joystick if not already started
-    // FIXME: check for errors
-    SDL_Init(SDL_INIT_JOYSTICK|SDL_INIT_GAMECONTROLLER);
-    SDL_GameControllerEventState(SDL_ENABLE);
-    SDL_JoystickEventState(SDL_ENABLE);
+const std::string SDLEventTypeToDebugString(int32_t event_type) {
+    switch (event_type) {
+    case SDL_CONTROLLERBUTTONDOWN:
+    case SDL_CONTROLLERBUTTONUP:
+        return "SDL_CONTROLLERBUTTON";
+    case SDL_CONTROLLERAXISMOTION:
+        return "SDL_CONTROLLERAXISMOTION";
+    case SDL_JOYBUTTONDOWN:
+    case SDL_JOYBUTTONUP:
+        return "SDL_JOYBUTTON";
+    case SDL_JOYAXISMOTION:
+        return "SDL_JOYAXISMOTION";
+    case SDL_JOYHATMOTION:
+        return "SDL_JOYHATMOTION";
+    default:
+        assert(false);
+        return "UNKNOWN SDL EVENT";
+    }
 }
 
-wxSDLJoy::~wxSDLJoy()
-{
-    // SDL_QuitSubSystem(SDL_INIT_JOYSTICK);
-}
-
-static int16_t axisval(int16_t x)
-{
+// Converts an axis value to a direction since we do not need analog controls.
+static int16_t AxisValueToDirection(int16_t x) {
     if (x > 0x1fff)
         return 1;
     else if (x < -0x1fff)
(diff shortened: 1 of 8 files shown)


Steelskin:
@rkitover @briansrls 
Please take a look. I have spent some time trying to clean up the code so it is less error-prone in the future. The main change in this pull request is that SDL objects are encapsulated in their own class, ensuring better handling of these.
@rkitover Sorry, it looks like this is going to conflict with your pull request adding GC and Switch controllers support.

src/wx/widgets/sdljoy.cpp
```diff
@@ -38,420 +36,455 @@ static int16_t axisval(int16_t x)
...
-                }
+    if (value_changed) {
+        wxLogDebug("GOT %s: joy:%d ctrl_idx:%d val:%d prev_val:%d",
+                   SDLEventTypeToDebugString(event_type), index_,
+                   control_index, control_value, previous_value);
 
-                got_event = true;
+        auto handler = wxGetApp().frame->GetJoyEventHandler();
+        if (!handler) {
+            std::cerr << "NO HANDLER" << std::endl;
```
rkitover, line 202: Remove this.

src/wx/widgets/sdljoy.cpp
```diff
@@ -38,420 +36,455 @@ static int16_t axisval(int16_t x)
...
+        control = wxSDLControl::WXSDLJOY_HAT;
+        previous_value = hats_[control_index];
+        if (previous_value != control_value) {
+            hats_[control_index] = control_value;
+            value_changed = true;
+        }
+        break;
+
+    // We should only receive button, hat or axis events so just default here.
+    default:
```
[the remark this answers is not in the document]
rkitover: Never mind, I confused this with the later poll method, but still I'd check explicitly.

src/wx/widgets/sdljoy.cpp
```diff
@@ -38,420 +36,455 @@ static int16_t axisval(int16_t x)
...
-                    auto val      = e.jbutton.state;
-                    auto prev_val = state.button[but];
+void wxSDLJoyState::SetRumble(bool activate_rumble) {
+    if (!IsGameController())
+        return;
 
-                    if (val != prev_val) {
-                        CreateAndSendEvent(state.index, WXSDLJOY_BUTTON, but, val, prev_val);
+    if (activate_rumble)
+        SDL_GameControllerRumble(game_controller_, 0xFFFF, 0xFFFF, 300);
```
rkitover, line 269: You need to check SDL version here, this is why the CI is failing.

src/wx/widgets/sdljoy.cpp
```diff
@@ -38,420 +36,455 @@ static int16_t axisval(int16_t x)
...
-    if (dev && SDL_IsGameController(0)) {
-        if (rumbling) {
-            SDL_GameControllerRumble(dev, 0xFFFF, 0xFFFF, 300);
-            if (!IsRunning())
-                Start(150);
-        }
-        else {
-            SDL_GameControllerRumble(dev, 0, 0, 0);
+    // Do rumble only on the first device.
+    joystick_states_.begin()->second->SetRumble(activate_rumble);
```
rkitover, line 478: Please do the check for the first device being a GC here, we don't want to run the timer unnecessarily.

src/wx/widgets/wx/sdljoy.h
```diff
@@ -1,138 +1,102 @@
...
-enum {
-    // The types of supported controls
-    // values are signed-16 for axis, 0/1 for button
-    // hat is bitmask NESW/URDL
-    WXSDLJOY_AXIS,
-    WXSDLJOY_HAT,
-    WXSDLJOY_BUTTON
-};
+    // Set of requested SDL joystick indexes.
+    std::unordered_set<int> requested_indexes_;
```
rkitover, line 91: This should be an unsigned type.

rkitover (changes requested):
This is excellent, your modern C++ is much better than mine, and I would love to have you on our team. I sent you an invite.

I left some comments on the code, some other more general things:

- The crash on disconnect was fixed, you may have been testing an earlier version of my PR. I believe the crash was caused by `SDL_GameControllerGetAttached()`.

- This suffers the same problem as an earlier version of my PR that @briansrls pointed out, SDL may change the joystick index on device disconnects. On disconnect events you should destroy and reconnect all requested indexes, this should be trivial.

- Don't worry about my PR, I will rebase it on top of this once this is merged.

src/wx/widgets/sdljoy.cpp
```diff
@@ -38,420 +36,455 @@ static int16_t axisval(int16_t x)
...
-        }
-        else {
-            SDL_JoystickClose(dev);
-        }
-
-        dev = nullptr;
+void wxSDLJoy::ConnectController(int index) {
+    std::unique_ptr<wxSDLJoyState> joy_state(new wxSDLJoyState(index));
+    if (joy_state->IsValid()) {
+        joystick_states_.emplace(
```
rkitover, line 444: This should be done by first erasing the index and then inserting the new state.

But since SDL likely handle multiple open calls just fine, just replace the `emplace()` call with an `erase()` followed by `insert()`.

rkitover: Actually no, that won't work either, if we do that the device will be closed and the new state will no longer be valid.

You'll need to keep a map of indexes to states and erase the connecting index and then insert it.

briansrls:
Hi @Steelskin , I also think your code is excellent.  I will make some comments regarding the usage of SDL libraries, if any.

I suggest you read [this thread](https://bugzilla.libsdl.org/show_bug.cgi?id=5264) I had with someone who works on the SDL joystick library.  I think you'll get a good idea of the following:

1.  Instance ID
2.  Index ID
3.  Player Index

Note, this is also a very important point they made:

```
2. Joystick and controller are ref counted. You need to call close for each call to open to actually free their memory.
```

We should consider this, or we will have small memory leaks.

src/wx/widgets/sdljoy.cpp
briansrls: Hi @rkitover, also consider this point made by SDL:

```
2. Joystick and controller are ref counted. You need to call close for each call to open to actually free their memory.
```

We should avoid calling open/close multiple times.  I think this is used for a publish/subscriber type of behavior.  For example, let's say multiple modules want to keep track of the Joystick independently.  It should only be freed when they are all finished.  I don't think we have this need, so we should only call open once.

src/wx/widgets/sdljoy.cpp
Steelskin: Done.

src/wx/widgets/sdljoy.cpp
briansrls: I was thinking about my comment more, and I think it this strategy would work if we also use player_index instead of index, like you are now.  

So, my vote is for:

Not closing/reopening a controller on every state change
Using player_index, as it isn't as volatile as index

src/wx/widgets/sdljoy.cpp
Steelskin: I changed this so it is handled in the `wxSDLJoyState` class. There is no need to do this in `wxSDLJoy` class anymore.

src/wx/widgets/sdljoy.cpp
Steelskin: Yes, the SDL Check was for the API, silly me :)

src/wx/widgets/sdljoy.cpp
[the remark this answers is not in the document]
Steelskin: I did that and added asserts to make sure we're doing the right thing.

src/wx/widgets/sdljoy.cpp
Steelskin: I spent some time toying around with SDL and indexes and everything. Unfortunately, there is no easy fix for this. I dhave done some changes to deal with this:
* The `wxSDLJoy` public API should only deal with *player indexes*. The fact that we also use *SDL indexes* internally is an implementation details that is of no use to the caller.
* As a result, `wxSDLJoy` should take care of converting between player indexes and SDL indexes.
* The code should be clear on what it is actually using. I have renamed variables and parameters to that end. Player indexes are unsigned, SDL indexes are signed because that is what the SDL API uses. Unfortunately, I have not figured out a good way to map between the 2 other than adding or substracting 1, which is ugly and probably error-prone down the line. However, as a result, I no longer care about the SDL index after joystick initialization, which is a slight improvement.
* We basically have 2 modes for connecting joysticks: During configuration, where we poll all joysticks and otherwise, where we only poll a select number of joysticks. I have changed the public API to reflect this. We now have a `PollJoysticks()` API to poll a select number of player indexes. and `PollAllJoysticks()` where we remap on every connect/disconnect. The reason for this is that we do not want the active joystick to disconnect during gameplay because another, unrelated joystick has disconnected. For instance, if you are playing with joystick player index 2 (SDL index 1), you do not want to be disconnected because joystick player index 1 (SDL index 0) has disconnected.
* SDK is *weird* with joystick ID handling. Connect 2 controllers (SDL indexes 0 and 1), disconnect SDL index 0, you are left with one controller, that ends up having SDL index 0. Reconnect the other controller, both controllers are back to their original SDL index position, meaning the previous controller is back to SDL index 1.

Please let me know what you think. I'm not very happy with my solution but I can't figure out a better way to do this without losing the joystick connection during gameplay in some cases - like multiple VBA-M instances running at the same time.

src/wx/widgets/wx/sdljoy.h
Steelskin: I changed the API here so it's an int since this is an SDL index.

Steelskin:
Thanks for the comments!

I have tried to clean up SDL and joystick indexes usage but I am not very happy with my solution. I still think it is an improvement overall but we really should have a better way of dealing with that. I toyed a bit with SDL GUIDs, unfortunately, they are not unique per controller. I have a number of 8bitdo controllers and they all report the same GUID. It is still a promising thing to look at but we need to figure out what the API should look like.

Steelskin:
Ah, looks like `SDL_JoystickGetDeviceInstanceID` is too recent for the older Ubuntu versions. I am going to add a workaround for that.

src/wx/widgets/sdljoy.cpp
briansrls: I think I am on the same page that you are.  I tried a few things, similar to you, and found that there is no *perfect* solution here.  Of course we are at the mercy of the SDL library, so it looks like the best we can do is to agree that some solution is our best possible effort.  Let me just respond to your points with some of my thoughts:

>The wxSDLJoy public API should only deal with player indexes. The fact that we also use SDL indexes internally is an implementation details that is of no use to the caller.

This is correct.  I would actually argue we should not use the SDL joystick index at all past the initially connection.  I don't think storing them is valuable to us.

For example, in my implementation I was thinking the player_index should be managed as follows:
We have a maximum of N players (say, 4 players).  We have a simple boolean array of length 4:

[false, false, false, false]

This indicates to us if we have any players active on that player_index.  Now, we connect the first controller, it has SDL index 0, instance ID 1.  We assign it to player_index 0 using SDL_GameControllerSetPlayerIndex.

[true, false, false, false]

Now, we accept another controller connection.  Our second player is SDL index 1, instance ID 2.  We look at our map, and determine that player_index 1 is available.  We set this controller to player_index 1 using SDL_GameControllerSetPlayerIndex.

[true, true, false, false]

Now, say player 1 disconnects.  Sure, the SDL indexes will be sliding around.  Player 2's SDL index will slide from index 1 to index 0, like you described.  However, it will still have player_index 1.  This should be the source of truth, along with its instance ID.

After player 1 disconnects, we clear that hole in our table:

[false, true, false, false]

Now, whatever controller connects next, will get player_index 0, and fill the hole we created.

[true, true, false, false]

So, my idea is to ditch SDL index immediately, set a self determined player index via SDL_GameControllerSetPlayerIndex and SDL_GameControllerGetPlayerIndex to avoid sliding controllers.  This is basically a very rudimentary controller port system, like on a console.  The only downside to this is that we have no concept of tracking controllers through disconnects, so technically if both player 3 and player 1 disconnect, they could swap depending on the order they reconnect in.

I just read now that you recently introduced player_index instead of index, which is basically what I'm describing here.  However, I see it is simply set to SDL index + 1.  Can you explain what your idea is now?

> As a result, wxSDLJoy should take care of converting between player indexes and SDL indexes.

Right, this is basically the system I described above.  It's simply one idea, there are other ways of doing this like you've probably thought of.

> The code should be clear on what it is actually using. I have renamed variables and parameters to that end. Player indexes are unsigned, SDL indexes are signed because that is what the SDL API uses. Unfortunately, I have not figured out a good way to map between the 2 other than adding or substracting 1, which is ugly and probably error-prone down the line. However, as a result, I no longer care about the SDL index after joystick initialization, which is a slight improvement.

Great, I am a fan of all of this.  I think not caring about it is the right way to go.  And the more accurate naming will help us kill the bugs more easily, because we should never confused the context in which they are used.

>  We basically have 2 modes for connecting joysticks: During configuration, where we poll all joysticks and otherwise, where we only poll a select number of joysticks. I have changed the public API to reflect this. We now have a PollJoysticks() API to poll a select number of player indexes. and PollAllJoysticks() where we remap on every connect/disconnect. The reason for this is that we do not want the active joystick to disconnect during gameplay because another, unrelated joystick has disconnected. For instance, if you are playing with joystick player index 2 (SDL index 1), you do not want to be disconnected because joystick player index 1 (SDL index 0) has disconnected.

I was a little confused by this, how could this happen?  Would we just receive an event that index 2 disconnected, and then close that connection?  How could we accidentally disconnect player index 2 in this case?

> SDK is weird with joystick ID handling. Connect 2 controllers (SDL indexes 0 and 1), disconnect SDL index 0, you are left with one controller, that ends up having SDL index 0. Reconnect the other controller, both controllers are back to their original SDL index position, meaning the previous controller is back to SDL index 1.

Yes, I'm assuming you're referring to SDL index and not player index, this is also kind of why I had my idea to track them internally using some data structure (I've been calling this behavior "sliding joysticks" in this post).

>  Please let me know what you think. I'm not very happy with my solution but I can't figure out a better way to do this without losing the joystick connection during gameplay in some cases - like multiple VBA-M instances running at the same time.

I will also think on this a bit.  I have not fully reviewed your code yet, but I suggest we just think about it for a day or two and just decide what is the best course of action.  It doesn't need to be perfect, but since you already did a great job, let's make it as great as we can.

src/wx/wxvbam.cpp
```diff
@@ -927,23 +928,26 @@ void MainFrame::SetJoystick()
...
+    joy.StopPolling();
 
     set_global_accels();
 
     if (!emulating)
         return;
 
-    for (int i = 0; i < 4; i++)
+    std::unordered_set<unsigned> needed_joysticks;
+    for (int i = 0; i < 4; i++) {
```
briansrls, line 939: Request to make this "4" a macro like NUM_JOYSTICKS or something, there are other parts of the code that use this value I believe.

rkitover (changes requested):
One thing you may not have known, is that the SDL API wiki is not up to date, if you click on the header files there is more stuff there.

Now to @briansrls's point that we should use `player_index`, we can probably do that.

This is the API for `player_index`:

```c
extern DECLSPEC SDL_GameController *SDLCALL SDL_GameControllerFromPlayerIndex(int player_index);
extern DECLSPEC int SDLCALL SDL_GameControllerGetPlayerIndex(SDL_GameController *gamecontroller);
extern DECLSPEC void SDLCALL SDL_GameControllerSetPlayerIndex(SDL_GameController *gamecontroller, int player_index);

extern DECLSPEC int SDLCALL SDL_JoystickGetDevicePlayerIndex(int device_index);
extern DECLSPEC SDL_Joystick *SDLCALL SDL_JoystickFromPlayerIndex(int player_index);
extern DECLSPEC int SDLCALL SDL_JoystickGetPlayerIndex(SDL_Joystick * joystick);
extern DECLSPEC void SDLCALL SDL_JoystickSetPlayerIndex(SDL_Joystick * joystick, int player_index);
```

I'm not sure what version of SDL is required for this, but we may be able to fall back to using `device_index+1` on older versions.

Here's my proposal for the design:

- On connect events, call `SDL_JoystickGetDevicePlayerIndex(int device_index)`, if it's `-1`, check what is the first available index in the set, and if it matches a requested index from the GUI call `SDL_JoystickSetPlayerIndex()` with the new index.

- The `requested_sdl_indexes_` map would become `requested_player_indexes_`. We wouldn't use the `device_index` at all, except for getting the `player_index`.

- `RemapControllers()` would no longer be necessary, since we would be managing the `player_index` through the API.

src/wx/wxvbam.cpp
rkitover: This is not related to this PR, but feel free to clean up some magic numbers as a separate PR in the core and the core interface, I did a very little bit of that but we need to do a lot more.

rkitover:
@briansrls described exactly the design I was thinking of above for `player_index`, except we can't limit the map to 4 controllers because during input configuration or accel configuration we have to accept input from all available controllers, and we store the `player_index` in the config for one of the 4 logical GBA controllers.

I think we're on the same page here.

briansrls:
> @briansrls described exactly the design I was thinking of above for `player_index`, except we can't limit the map to 4 controllers because during input configuration or accel configuration we have to accept input from all available controllers, and we store the `player_index` in the config for one of the 4 logical GBA controllers.
> 
> I think we're on the same page here.

Maybe this even could all be done internally?  For example, to avoid issues with compatibility, we could just store a private variable called "virtual_player_index".

We never call SetPlayerIndex, or GetPlayerIndex, and just ignore index and player_index.  We only deal with instance_id and virtual_player_index.

We then instead of maintaining a bool array, we maintain an array like:

vector<wxSDLJoy> player_index_array, where the index corresponds to our virtual_player_index.

I guess I'm even questioning the need to mess with the SDL api.  I think we need a reliable way to preserve our ordering internally, and to map that ordering to instance IDs.  I think we can do that even without the API.

Steelskin:
Thanks for your thoughts on the matter. I think we can get something nicer if we make some form of explicit player index type. I will dig through the SDL API for that tomorrow and see if it is better to use it or do our own thing.

src/wx/widgets/sdljoy.cpp
```diff
@@ -38,420 +39,536 @@ static int16_t axisval(int16_t x)
...
-    if (joy_n < 0) {
-        for (auto&& joy : joystate)
-            DisconnectController(joy.second);
-
-        joystate.clear();
+void wxSDLJoy::DisconnectController(SDL_JoystickID joystick_index) {
+    assert(!add_all_joysticks_);
 
+    auto iter = joystick_states_.find(joystick_index);
+    if (iter == joystick_states_.end()) {
```
rkitover, line 519: Offtopic, but I really don't like this particular idiom in modern C++. Is there something like a convenient optionals header we could use for these operations?

briansrls: Would you prefer

if(map.count(key) == 0)

briansrls: Oh, I think someone already implemented a contains(map, key) function in our code.

rkitover: Yes, I did that, it's a very simple template function.

The issue is doing a lookup that may not exist, and not doing it twice. Languages like Swift use optionals for this.

rkitover:
@Steelskin any objections if I merge my Nintendo controllers PR for now, until we get this work sorted out. I can later resolve conflicts and merge this. I'd just like the GameCube/Switch controller support in the nightlies.

src/wx/widgets/sdljoy.cpp
briansrls: I understand now, I didn't know this was called an "optional".  I was going to say I use this in python a lot, although I'm not sure if those would be considered optionals (like None or inf), but maybe just a natural feature of the duck typing.

Anyway, I found this very recent library https://en.cppreference.com/w/cpp/utility/optional

Steelskin: Unfortunately, `std::optional` is C++17.
In this case, however, since we re-use the iterator to delete the element later on, it is faster to keep track of the iterator. I do agree the STL container syntax is less than ideal but it's what we have to deal with.

Steelskin:
> @Steelskin any objections if I merge my Nintendo controllers PR for now, until we get this work sorted out. I can later resolve conflicts and merge this. I'd just like the GameCube/Switch controller support in the nightlies.

@rkitover Yes, please, go ahead, I will rebase on top.

Steelskin:
I spent some time trying to improve this today. Unfortunately, at this point, I think it is best to leave this PR in this state. This is why:
* I believe what we really want out of all this is an abstract `wxSDLJoyKey`, which should be the API that `wxSDLJoyEvent` uses to identify a unique joystick device via this key. Internally, this is basically an integer but this is opaque to the API user.
* We could use an integer directly with the code in the current state - this is basically what I do right now but it's unwieldy and you have to be careful about what exact type you are using.
* The problem with changing `wxSDLJoyEvent` is that all the callers for this assume the joystick index (that I renamed to player index) is an integer. I believe this could easily be changed but it's a lot of changes throughout the rest of the codebase and this pull request is already large enough as it is.

> vector<wxSDLJoy*> player_index_array, where the index corresponds to our virtual_player_index. We can set the values to NULL or to a valid SDLJoy object depending on who owns that index.

@briansrls I really like this but I think we should go one step further and use an abstraction like I described above, using a `map<wxSDLJoyKey, zxSDLJoy*>`. It is a few extra things to keep track of but it should work. The logic around reconnecting a joystick gets a bit tricky but it is *much* easier to handle that with the `*PlayerIndex` SDL APIs. So I think a better first step would be to link statically against SDL on every platform, if this is possible. Then we don't have to worry about SDL versions.

For now, I suggest that we rebase this as needed and land it as is. We're certainly no worse than the current situation and this is something we can build on.

briansrls:
@Steelskin this plan is totally fine by me, could you please rebase it and then I will test it.  We could track this feature as an enhancement, for now as long as there are no bugs/crashes and basic functionality works I will still be very happy.

Steelskin:
@briansrls I believe this is already up-to-date with the master branch, you should be fine to test. I'm filing a bug for the next improvements.

briansrls:
> @briansrls I believe this is already up-to-date with the master branch, you should be fine to test. I'm filing a bug for the next improvements.

@rkitover are you alright with this plan as well?  I think steelskin makes a fine point, the PR is pretty big already

briansrls:
I am looking into one issue I found.  Here is the scenario:  Two controllers connected, a PS4 and Xbox One controller.  

I disconnect both controllers, in the below log I disconnect the xbox one s and then ps4 controller.  I don't think it matters which order.

I then connect the ps4 controller back.  We get the relevant events and see the success log, Connected joystick 1: PS4 Controller

I then connect the xbox one controller, we see the relevant events, but the controller wasn't actually added back for some reason.

I'm going to look into it more later and get more logs.  I don't see this issue in the nightly.

If I am in the state where both joysticks are plugged in, I can workaround it by entering say the Input menu.  This triggers a remapping of the controllers, which seems to forceably connect the xbox one controller.  I will try to see why the reconnect is failing.

```
Controller Device Removed  //ps4 controller receives CONTROLLERDEVICEREMOVED                                                                                     
Disconnected joystick 2  //Disconnect ps4 controller
Joydevice Removed //ps4 receives JOYDEVICEREMOVED                   
Controller Device Removed //same for xbox one
Disconnected joystick 1 //same for xbox one                                                         
Joydevice Removed //same for xbox one                                                               
Controller Device Added //ps4 controller gets CONTROLLERDEVICEREMOVED                                                                      
Joystick ID = 2 //Added with ID 2    
Connected joystick 1: PS4 Controller //Connected with index 1
joy_state->IsValid == TRUE                                      
Joydevice Added //JOYDEVICE ADDED for ps4                                
Controller Device Added //Controller device added for ps4                                                               
Joystick ID = 3 //Added with Instance ID 3                                  
Only add the joysticks we care about: 0 //Why is it hitting this check here?  Shouldn't we still care about this index?
Joydevice Added 
```

Here is the diff of my logs so you can relate it to the code:

```
diff --git a/src/wx/widgets/sdljoy.cpp b/src/wx/widgets/sdljoy.cpp
index ea210d9d..8ea6e28a 100644
--- a/src/wx/widgets/sdljoy.cpp
+++ b/src/wx/widgets/sdljoy.cpp
@@ -352,6 +352,7 @@ void wxSDLJoy::Poll() {
         case SDL_CONTROLLERDEVICEADDED:
         {
             int sdl_index = e.cdevice.which;
+            std::cout << "Controller Device Added" << std::endl;
             if (add_all_joysticks_)
                 RemapControllers();
             else
@@ -363,6 +364,7 @@ void wxSDLJoy::Poll() {
         case SDL_CONTROLLERDEVICEREMOVED:
         {
             SDL_JoystickID joy_id = e.cdevice.which;
+            std::cout << "Controller Device Removed" << std::endl;
             if (add_all_joysticks_)
                 RemapControllers();
             else
@@ -427,7 +429,7 @@ void wxSDLJoy::Poll() {
         case SDL_JOYDEVICEADDED:
         {
             int sdl_index = e.jdevice.which;
-
+            std::cout << "Joydevice Added" << std::endl;
             // Do not process joystick events for game controllers.
             if (SDL_IsGameController(sdl_index))
                 return;
@@ -443,6 +445,7 @@ void wxSDLJoy::Poll() {
         case SDL_JOYDEVICEREMOVED:
         {
             SDL_JoystickID joy_id = e.jdevice.which;
+            std::cout << "Joydevice Removed" << std::endl;
             if (add_all_joysticks_)
                 RemapControllers();
             else
@@ -491,21 +494,28 @@ void wxSDLJoy::ConnectController(int sdl_index) {
     SDL_JoystickClose(joystick);
 #endif
 
+    std::cout << "Joystick ID = " << joystick_id << std::endl;
+
     if(contains(joystick_states_, joystick_id)) {
         // Do not reconnect the same device twice. This can happen because
         // SDL indexes are not stable and change whenever a device is
         // connected or disconnected.
+
+        std::cout << "Do not connect the same device twice" << std::endl;
         return;
     }
 
     auto iter = needed_sdl_indexes_.find(sdl_index);
     if (iter == needed_sdl_indexes_.end()) {
+
+        std::cout << "Only add the joysticks we care about: " << sdl_index << std::endl;
         // Only add the joysticks we care about.
         return;
     }
 
     std::unique_ptr<wxSDLJoyState> joy_state(new wxSDLJoyState(sdl_index));
     if (joy_state->IsValid()) {
+        std::cout << "joy_state->IsValid == TRUE" << std::endl;
         joystick_states_.emplace(
             joy_state->joystick_id(), std::move(joy_state));
         needed_sdl_indexes_.erase(iter);
```

briansrls:
I will look into it again when I have some more time.  Basically, I want to scrutinize needed_sdl_indexes_ and see if we are missing a readd.  I just put some more logs and it looks like we add that index when the xbox controller disconnects, but then we are surprised when it tries to reconnect saying it doesn't exist.

Steelskin:
Ah I am guessing this is because I do not disconnect the remaining controller, which no longer has the same sdl_index.
I also noticed a bug, we should not handle added/removed events for game controllers because we are going to get the corresponding joystick event anyway.

Steelskin:
@briansrls I tested your scenario, this is happening due to the instability of SDL indexes over reconnects. The xbox one controller is being added as SDL index 0, but we already connected the PS4 controller as SDL index 0, so we are no longer expecting a new controller at SDL index 0.

The reason why I am reluctant to remap all controllers when we are only polling for some joysticks is to avoid this scenario, which happens on master branch:
* Configure joystick at SDL index 1 (player index 2) as the player 1 joystick.
* Start a game
* Joystick at SDL index 0 disconnects.
* Your joystick at SDL index 1 is no longer active in the emulator, you have to reconnect the other joystick to proceed.

Whereas, with this PR, disconnection of a joystick at a lower SDL index does not trigger the disconnect of your active controller. But, as you pointed out, this breaks if you have 2 active joysticks. The second joystick does not reconnect properly if you reconnect them in a certain order.

Anyway, I solved the issue by handling disconnect and new connections differently. We always remap on a new connection and only remove one joystick on a disconnect. This solves the issue in both of the above cases. Do note that things will still break in scenarios with more controllers being connected and disconnected but I would guess these are much more rare.

briansrls:
@Steelskin that sounds good to me.

src/wx/widgets/sdljoy.cpp
rkitover: What are the implications for compatibility of switching to C++17? So far I have been keeping it at C++11, but I am clueless about C++ and rarely use advanced features. We do not necessarily have to support 2011 compilers in this year of our lord.

This would only even matter for libretro, and we could keep the core at C++11 or even earlier, and I can make the necessary cmake adjustments for that.

I made this comment on the wrong instance of this usage, I actually meant the the instance id lookups etc.. I realize that in this case you are reusing the iterator and this is all fine.

But in those other cases, I'd like something more elegant. We could switch to C++17 for linux/windows, if everyone is OK with that. If not, I'm sure there is some C++11 compatible header someone wrote for this, perhaps in boost or whatever. If using boost we wouldn't have to depend on distro libraries, we can just bundle the individual headers necessary, there is unlikely to be a runtime dependency. @ZachBacon I'd like your opinion on this too, I know you don't like boost.

I actually consider this important. This is not the only place where we'd need optional syntax, it would be used everywhere in new code. `std::optional` looks very nice actually.

ZachBacon: I very much am against the use of boost, the core itself needs to remain c++ compatible with older versions due to libretro core being able to be compiled for certain consoles, but the GUI, if we can avoid having another dependency, I very much would appreciate not using boost. That's my personal matter on the situation. It's not that I don't like boost, it's that I just don't want to introduce yet another dependency to vba-m

rkitover: Like I said, if I use boost headers I wouldn't depend on the library.

ZachBacon: if you think you can get it to work on all platforms the libretro port uses, then go for it. That's the only condition I have.

rkitover: libretro has nothing to do with this, and I said that already.

ZachBacon: I know, but I am still stating it. If it's just the header, go ahead.

rkitover:
Sorry, I'm just getting back to all of this after a very bad month.

> So I think a better first step would be to link statically against SDL on every platform, if this is possible. Then we don't have to worry about SDL versions.

On Windows we use the latest SDL from vcpkg, and it's linked statically. On Linux we only need to support recent distributions which have a recent SDL. I am going to drop support for ancient Ubuntu versions etc.. and switch to C++17 for the GUI.

rkitover:
Before I forget, I have an issue with the reintroduction of hats.

I noticed on my 360 controllers with some drivers I get both button and hat events for the direction pad.

Is there a way we can ignore hat events for all controllers where this is possible, perhaps check for the SDL DirectInput driver directly?

rkitover:
As far as the previously mentioned `SDL_JoystickSetPlayerIndex()` and `SDL_GameControllerSetPlayerIndex()`, these are a 2.0.12 API released on March 11th of this year. To not make things more difficult for backporters to still not ancient distributions, we'd want a version guard around them and some kind of fallback.

rkitover:
I'm not going to merge my nintendo controllers branch yet because I noticed an issue with reconnects that I need to work out.

src/wx/widgets/sdljoy.cpp
```diff
@@ -38,420 +39,489 @@ static int16_t axisval(int16_t x)
...
-    if (joy_n < 0) {
-        for (auto&& joy : joystate)
-            DisconnectController(joy.second);
-
-        joystate.clear();
-
-        return;
+void wxSDLJoy::RemapControllers() {
+    if (!add_all_joysticks_) {
+        // Save the connected SDL indexes first.
```
rkitover, line 471: We shouldn't do this, the wanted player indices should only be set by the API client.

src/wx/widgets/sdljoy.cpp
```diff
@@ -38,420 +39,489 @@ static int16_t axisval(int16_t x)
...
-        if (rumbling) {
-            SDL_GameControllerRumble(dev, 0xFFFF, 0xFFFF, 300);
-            if (!IsRunning())
-                Start(150);
-        }
-        else {
-            SDL_GameControllerRumble(dev, 0, 0, 0);
-            Stop();
+    joystick_states_.clear();
+    for (int i = 0; i < SDL_NumJoysticks(); i++) {
```
rkitover, line 478: This should iterate over wanted SDL indices in the normal case and all available joysticks in the add-all case.

rkitover (changes requested):
I left some comments on the code and in the thread. The main thing is the `RemapControllers()` logic needs a bit more work.

src/wx/widgets/sdljoy.cpp
rkitover: As the code is currently, this is incorrect behavior, since on attach of a controller all available joysticks will be attached, but in the general case we only want the indexes requested from config.

rkitover:
@Steelskin I can merge all of this manually with some changes if you like.

Steelskin:
> @Steelskin I can merge all of this manually with some changes if you like.

Thank you but I can get to it tonight, I have been busy in the past few days. If you want to do it yourself before then, be my guest.

src/wx/widgets/sdljoy.cpp
Steelskin: You are correct, I rewrote this method.

src/wx/widgets/sdljoy.cpp
Steelskin: This was from an older commit when I was not reconnecting all joysticks on a new connection, which led to some issues. It no longer has any purpose so I removed it.

src/wx/widgets/sdljoy.cpp
```diff
@@ -38,420 +39,489 @@ static int16_t axisval(int16_t x)
...
+            RemapControllers();
+            got_event = true;
+            break;
+        }
 
-        instance_map[instance] = &(joystate[joy]);
+        case SDL_JOYDEVICEREMOVED:
+        {
+            SDL_JoystickID joy_id = e.jdevice.which;
+            if (add_all_joysticks_)
```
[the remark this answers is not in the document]
Steelskin: The variable is used to differentiate between adding a select list of joysticks or listening to all joysticks. I renamed it to `poll_all_joysticks_` since it was confusing.

src/wx/widgets/sdljoy.cpp
```diff
@@ -38,420 +39,489 @@ static int16_t axisval(int16_t x)
...
-
-                    if (val != prev_val) {
-                        CreateAndSendEvent(joy.first, WXSDLJOY_AXIS, axis, val, prev_val);
-
-                        joy.second.axis[axis] = val;
+        case SDL_JOYAXISMOTION:
+        {
+            SDL_JoystickID joy_id = e.jaxis.which;
+            auto joy_state = joystick_states_.find(joy_id);
+            if (joy_state != joystick_states_.end()) {
```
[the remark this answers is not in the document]
Steelskin: I was trying to do that but my local checkout had issues finding `<optional>` I am using mingw to build and test on Windows, it might have something to do with that. In any case, feel free to do it as I have no way of testing my code is correct.

Steelskin:
Thank you, I think at this point, we have 2 issues left:
- The one with the 360 controller you mentioned above. Could you let me know what model (wired/wireless) and what OS you were using? These should show up as XInput devices and therefore as game controllers in SDL, so we should not be getting HAT events for them so it is odd.
- Using `std::optional` with `map::find()`. As I mentioned in the comments, I could not get the header to show up for the compiler in a mingw environment but I may very well be holding it wrong.

src/wx/widgets/sdljoy.cpp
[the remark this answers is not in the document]
rkitover: Which version of mingw are you using in what environment? Which version of gcc is it?

rkitover:
> The one with the 360 controller you mentioned above. Could you let me know what model (wired/wireless) and what OS you were using? These should show up as XInput devices and therefore as game controllers in SDL, so we should not be getting HAT events for them so it is odd.

I believe I was seeing this before the initial GameController refactor with xboxdrv on linux, so I guess it should not be an issue anymore.

> Using std::optional with map::find(). As I mentioned in the comments, I could not get the header to show up for the compiler in a mingw environment but I may very well be holding it wrong.

Which mingw environment are you using? I see this in msys2: `./include/c++/10.2.0/optional`

src/wx/widgets/sdljoy.cpp
```diff
@@ -38,420 +39,472 @@ static int16_t axisval(int16_t x)
...
+            got_event = true;
+            break;
         }
-        else {
-            SDL_JoystickClose(dev);
+
+        case SDL_JOYDEVICEREMOVED:
+        {
+            SDL_JoystickID joy_id = e.jdevice.which;
+            if (poll_all_joysticks_)
```
rkitover, line 422: Can this be removed? That is, can we remove the state in the `poll_all_joysticks_` case too, or would it cause issues in the input config dialog? This is my only remaining issue with the code as it is.

rkitover (changes requested):
I left one comment on the code, other than that everything looks great.

The git history is a mess, if you can, can you please clean it up and squash into one or more commits, as you feel would be appropriate (using `rebase -i` etc..) I can also do this myself if you prefer.

briansrls:
> I left one comment on the code, other than that everything looks great.
> 
> The git history is a mess, if you can, can you please clean it up and squash into one or more commits, as you feel would be appropriate (using `rebase -i` etc..) I can also do this myself if you prefer.

Is there a way we can configure auto squashing of commits?  We use this feature at work, it is very convenient.

rkitover:
> Is there a way we can configure auto squashing of commits? We use this feature at work, it is very convenient.
Meaning, squash when you finally merge.

I'd prefer it if people learn to use git properly, `rebase -i` is not difficult, and you should present a clean history for a feature, with good commit messages. This does not have to be one commit, ideally you would have one commit per specific thing that you did. But one commit can also be fine.

This is how work is done in projects much more serious than this one, and I would like to use the same methodology.

Steelskin:
Sorry about the history snafu, GitHub Desktop was being stupid and would not let me force push without creating a merge commit. I usually just squash everything all the time and keep a single commit on my work branches but GitHub does not work nicely with that.

> > Using std::optional with map::find(). As I mentioned in the comments, I could not get the header to show up for the compiler in a mingw environment but I may very well be holding it wrong.
> 
> Which mingw environment are you using? I see this in msys2: `./include/c++/10.2.0/optional`

I am using mxe, let me try switching to msys2, I'll get to the other comments at the same time.

Steelskin:
I accidentally deleted a commit, let me fix this. I really do not like the GitHub flow, I am used to Gerrit.

src/wx/widgets/sdljoy.cpp
Steelskin: I thought about it and the only issue we would have is that we may end up with the wrong player index if a controller gets disconnected while we are in the input config dialog. But in the end, I think that it is very unlikely and if it does happen, all the user has to do is either reconnect the other joystick or re-open the configuration dialog, which is a very minor annoyance. I think most users will intuit that the player index depends on how many controllers they have connected and that it might change if they have more than one.

The `poll_all_joysticks_` variable also serves another purpose, which is to differentiate the polling mode (some joystick vs. all joysticks). To work around that, I have switched to another variable `is_polling_active_`, which is set to true when either mode is active.

Steelskin:
I ended up working around the `optional` issue by playing with the fact that we already have a `unique_ptr` in the map and we can simply use an utility method to return the raw pointer to the joystick state if it is present.

rkitover:
@Steelskin I gathered from your comments that you are using Windows, but compiling with MinGW, I finally finished a draft of my new PowerShell environment setup guide and tutorial, could you please try it:

https://github.com/visualboyadvance-m/visualboyadvance-m/blob/master/DEVELOPER-MANUAL.md#windows-native-development-environment-setup

As for the commits, I think this change can be squashed into one commit with a long description of your changes.

Steelskin:
Thank you, I squashed everything. I have been told in the past to keep a separate commit for each update but I much prefer keeping a single commit when I am working on something so I am very happy to do that here 😃 

Yes, I am working from a Windows laptop and building from a WSL2 ubuntu session mainly because I don't know the Windows command line really well and Visual Studio insists on rebuilding ffmpeg regularly, which takes forever. I also don't know cmake well enough to troubleshoot my issues and this is definitely something I need to work on. This setup is good enough for me for now but it is definitely not ideal.

rkitover (approved):

rkitover:
@Steelskin well I finally merged this, congratulations, this has been quite a project.

Regarding the issue of when to present multiple or a single commit for a work, it can be either, but, you have to remember that the commit message summary are meant to go into master, so commit messages like "make the build pass" should definitely be squashed.

For example, a single commit can be:

```
Refactor FooClass to be more modern.
```

while multiple commits can be something like:

```
Add X support to Foo system.
Add Y support to Foo system.
```

or something like:

```
Rewrite Foo system with better UI design.
Tweak new Foo system based on feedback.
```

it has to make sense when you're reading the git log for master, not the branch. It's ok to have multiple commits for a work if they show how it was developed more clearly or other reasons.

As for Visual Studio builds, yes the ffmpeg builds are annoying, I will hopefully address this soon, perhaps by switching to Conan for packages or setting up a build server for vcpkg packages.

I would still very much appreciate it if you could find some time to go through my tutorial and let me know what your impressions are. I am planning to break that PowerShell tutorial out into a separate project.

If you do, let me know what things you still find confusing about cmake or PowerShell and I will address them in the manual.

PR #779, opened 2020-11-21
Handle allocation quirk on non-null-returning allocators

koachan:
Some allocators return a non-null pointer when calling *alloc with zero size, that will segfault upon access.
Check for the size parameter before allocating anything to ensure that the breakpoint and trace fields will stay null when `size >> 3` or `size >> 1` is zero.


koachan:
Wait, it's even worse. Just checked [the documentation for calloc](https://pubs.opengroup.org/onlinepubs/9699919799/functions/calloc.html) and it seems that `calloc(0, n)` is allowed to return null. So yeah, this should fix that too.

rkitover:
Nice, thank you!

PR #790, opened 2020-12-11
Remove unintended semicolon from CPACK_PACKAGE_VERSION_PATCH

Clownacy:
I noticed this while trying to install a CPack-generated Debian package, since `dpkg` complains about the semicolon being an invalid character.


rkitover:
Thank you very much!

PR #791, opened 2020-12-14
SDL audio fixes

Clownacy:
I've fixed the bug that causes no audio to play on Windows without a hack, and I've fixed the audio delay. See the commit descriptions for more information.


rkitover:
Thank you very much for this, the audio delay was a high priority for me to fix.

PR #792, opened 2020-12-14
SDL frontend fixes and cleanup

Clownacy:
This silences a compiler warning (encountered on GCC 10.2), which complained about `strncpy`'s length parameter being determined by the length of the source string.

I've also fixed some bugged logic on Windows, where the profile path is used instead of the `argv[0]` path because of a shadowed variable. This bug presumably prevents the PATH from ever being searched, and it also seems to have broken the logic for searching the executable directory.

### DIFF
--- a/src/sdl/SDL.cpp
+++ b/src/sdl/SDL.cpp
@@ -323,23 +323,29 @@ bool sdlCheckDirectory(const char* dir)
 
 char* sdlGetFilename(char* name)
 {
-    char path[1024] = ""; // avoid warning about uninitialised value
+    char path[1024];
     char *filename = strrchr(name, FILE_SEP);
     if (filename)
-        strncpy(path, filename + 1, strlen(filename));
+        strcpy(path, filename + 1);
     else
-        sprintf(path, "%s", name);
+        strcpy(path, name);
     return strdup(path);
 }
 
 char* sdlGetFilePath(char* name)
 {
-    char path[1024] = ""; // avoid warning about uninitialised value
+    char path[1024];
     char *filename = strrchr(name, FILE_SEP);
-    if (filename)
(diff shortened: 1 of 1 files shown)


src/sdl/SDL.cpp
```diff
@@ -323,23 +323,29 @@ bool sdlCheckDirectory(const char* dir)
 
 char* sdlGetFilename(char* name)
 {
-    char path[1024] = ""; // avoid warning about uninitialised value
+    char path[1024];
     char *filename = strrchr(name, FILE_SEP);
     if (filename)
-        strncpy(path, filename + 1, strlen(filename));
+        strcpy(path, filename + 1);
```
glneo, line 329: What was wrong with the strncpy/sprintf versions?

src/sdl/SDL.cpp
```diff
@@ -323,23 +323,29 @@ bool sdlCheckDirectory(const char* dir)
...
 {
-    char path[1024] = ""; // avoid warning about uninitialised value
+    char path[1024];
     char *filename = strrchr(name, FILE_SEP);
-    if (filename)
-        strncpy(path, name, strlen(name) - strlen(filename));
-    else
-        sprintf(path, "%c%c", '.', FILE_SEP);
+    if (filename) {
+		size_t length = strlen(name) - strlen(filename);
```
glneo: You removed some whitespace issues in the last patch, then add some more here..

src/sdl/SDL.cpp
Clownacy: They both do the same thing as `strcpy` only with more overhead and, in the `strncpy`'s case, a compiler warning.

src/sdl/SDL.cpp
Clownacy: I corrected them in a later commit.

PR #835, opened 2021-04-24
Add basic VS Code support

Steelskin:
* Add variants for the VS Code CMake extension.
* Add documentation for building with VS Code.
* Remove the "/std:c++17" flag from the C compiler flags.


Steelskin:
I went through the documentation you added and managed to build natively on Windows. I have a preference for VS Code so I added some basic configuration for it. Please, let me know what you think.
Removing the flag from the C compiler also removes some annoying build output warnings.

ZachBacon:
apologies for such a late reply, I've been taking a hiatus due to life, I'm coming back in september, but I figured I'd at least clear up some stuff until I come back again.
I much rather not include preferences for editors like vscode since it would have to support a multitude of different configurations from mingw (either on windows or linux) and linux default gcc, it's something I am not looking to merge in it's current state

Steelskin:
Fair enough. I'll try to find another way to handle VS Code better with the current setup. I think we can get some improvements by doing some modifications to the build system but I need to dig into this more.
In the meantime, I can have my own local changes.

PR #838, opened 2021-04-25
Use constant aliases in MainFrame::FilterEvent()

Steelskin:
Previously, this function returned -1 or "true" but wxWidgets
defines constants for the returned values for FilterEvent()
overrides so use these instead.


Steelskin:
Please take a look, this is a simple no-op fix.

briansrls (approved):
Looks good to me

https://docs.wxwidgets.org/3.0/classwx_event_filter.html

Steelskin:
I did a small change to fix the indentation

ZachBacon:
looks good tyvm again apologies for taking so long to get to this.

PR #866, opened 2021-08-13
in Wx, the Option is Enable MMX, make the variable also enableMMX and…

Squall-Leonhart:
… set its default to on.

The present ini variable is from vbam 1.x and is behaviourally backwards where setting disablemmx to true actually enables mmx, 2.x builds should not be using the ini from 1.x for various reasons.

### DIFF
--- a/src/common/ConfigManager.cpp
+++ b/src/common/ConfigManager.cpp
@@ -166,7 +166,7 @@ int captureFormat = 0;
 int cheatsEnabled = true;
 int cpuDisableSfx = false;
 int cpuSaveType = 0;
-int disableMMX;
+int enableMMX = 1;
 int disableStatusMessages = 0;
 int dsoundDisableHardwareAcceleration;
 int filterHeight;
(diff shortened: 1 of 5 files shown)


Squall-Leonhart:
I'm not sure if i should be setting the variable default under dictionary* preferences; or void LoadConfig()

Squall-Leonhart:
looks like these changes do the trick, also prior inis have the disablemmx and replaced by enablemmx

denisfa:
Hello there! I am not really active here anymore, but I think this change is very straightforward. 

From my very brief search on the source code and the changes, shouldn't the variable `enableMMX` be checked somewhere on the filters/other files code too? `MMX` is a macro,  so you have to compile with `ENABLE_MMX`. From my grep search, if `MMX` is set, then the only variable that toggles functionality is `cpu_mmx`, that is hardcoded to `1` on `./src/filters/2xSaI.cpp`.

I am not sure if this was even toggable before... But then again, it has been so long that I am probably missing something. Leaving this for Zach. Everything else is ok.

denisfa (approved):

ZachBacon:
Apologies for the lateness of this, but it does look good, tvm

PR #882, opened 2021-10-02
Fix debian installdeps

DeenOConnor:


rkitover:
Merged manually with some changes, sorry for the delay.

PR #895, opened 2021-11-13
fix: make vbam build on riscv

XieJiSS:
When building on RISC-V using gcc, `defined(__GNUC__)` will be positive, so we need to add a guard.


XieJiSS:
The head of branch master can build successfully on RISC-V, so closing this PR.

Issue #900, opened 2021-11-21
Lost my save due to unexpectedly destructive keyboard defaults.


raflemakt:
This will not be a technical bug report, but I lost tens of hours of game progress so it might be worth to address.

I use the SDL version (vbam-sdl) where there's no toolbar menu over the game window, so out of curiosity I pressed F10 to see if there was one (since this is a common keybinding for accessing the toolbar menu). Turns out F10 undoes your last save.

Thing is, I never used save states (on purpose), I only saved through the in game menu like you would on an actual GB. I didn't realize at this point that pressing the F-keys could change anything unless I actually saved through the game menu so I pressed several of them to see what they did.

In hindsight it's obvious that I got what I asked for, but I think this could have happened to anyone. My suggestion is that the sort of "power user" utilities like save states should either not have keybindings by default or at least have less accessible keybindings (especially since it's common on laptops to have functions like volume and light levels on the function keys, so it's a catastrophe waiting to happen).

Rom: Pokémon Yellow (GBC)
Version: vbam-sdl 2.1.4
OS: Arch Linux

Squall-Leonhart:
F10 loads the saved state (Ctrl+F10) on slot 10.

raflemakt:
I'm still not sure how I did it, but I ended up reverting to an old state that I had (unpurposely) saved in the beginning of the game.

raflemakt:
#### I found the terminal output from when the overwrite happened. As you can see I pressed quite some F-keys to try to figure out what was going on.

$ vbam  Pokemon\ -\ Yellow\ Version\ \(UE\)\ \[C\]\[\!\].gbc 
VisualBoyAdvance-M 2.1.4
Trying patch Pokemon - Yellow Version (UE) [C][!].ips
Trying patch Pokemon - Yellow Version (UE) [C][!].ups
Trying patch Pokemon - Yellow Version (UE) [C][!].ppf
Loaded battery './/Pokemon - Yellow Version (UE) [C][!].sav'
Color depth: 32
Wrote battery './/Pokemon - Yellow Version (UE) [C][!].sav'
20:11:41 21.11.2021: Last save UNDONE
20:11:47 21.11.2021: Last load UNDONE
20:11:52 21.11.2021: Current state backed up to 9
20:11:52 21.11.2021: Loaded state 7
20:11:53 21.11.2021: Current state backed up to 9
20:11:53 21.11.2021: Loaded state 7
[...]

Issue #781, opened 2020-11-23
Save states apparently destroy “normal” saves


Wuzzy2:
OK, so this is technically not a bug report, but it might be major usability problem, so probably still relevant.

I just lost a TON of game progress with a single key press.
How? I loaded an old savestate by accidentally pressing F2 ...

Before, I saved the game the “normal” (in-game) way all the time, but this is useless here.
Restarting the game doesn't help, because the game will just show the savegame that was present at the time of the savestate.
If I knew that savestates also destroy the “normal” savegames, I would never have used them in the first place.
I can't find the help files either, so I don't know if there is any way to recover.

Anyway, the issue here is that it's possible to destroy your game progress with a single key press, and loading the “normal” way is (apparently) useless. Unless I am mistaken and overlooked something (I can't find the documentation for this game, only some very incomplete README files in `doc`.)

I use visualboyadvance-m, the WX version. 2.1.2. On Arch Linux (package name = `vbam-wx`).

Wuzzy2:
Oh, I'm sorry. I meant to say:

> I use visualboyadvance-m, the WX version. 2.1.4.

rkitover:
The "normal" save games are written to `.sav` files with the same name and in the same directory as your ROM. These are only written when the game writes to the flash, e.g. you save your game the normal way. Save states are written to other files and do not affect this. So if you load an old save state, it will not overwrite the `.sav` file unless the game actually saves the normal way, this depends on the game of course.

Squall-Leonhart:
This is because of the (i  believe default) option where Save states restore the battery, the Load state menu has an option to disable this behavior.

rkitover:
@Squall-Leonhart thanks, I will do this.

Wuzzy2:
Ah, that's a good thing. Yeah, the current default behavior is surprising, it's a good idea to change it to a non-surprising non-destructive behavior.

Because never in a 1000 years would I have expected, as a dumb user (haha!), that "saving a state" implies "overwriting the battery file".

Just to be sure I understand correctly: So what will the new behavior be then, precisely? When I save a state, the battery file is no longer touched? Do save states "belong" to a certain battery file or are they independent? What happens if the battery file was switched between saving and loading a save state? Are there any other things I need to know?

Squall-Leonhart:
For you? there won't be a change, the config file is already written so you would select the option yourself.

I can confirm that through various loads with the option set, i have not had a battery replaced.

[...]

PR #924, opened 2022-03-03
Add checkbox to Key Shortcuts to disable Fn keys

o-i-z-y-s:
Added a checkbox to the Key Shortcuts window that, when checked, will temporarily remove keybinds involving F1 through F10. Once the box is unchecked, the default keybinds are restored.

A user is likely to hit a function key and accidentally load a save state, potentially losing game progress (see #900 and #781), so this addition does the work of manually having to remove these defaults or rebind those keys.

### DIFF
--- a/src/wx/guiinit.cpp
+++ b/src/wx/guiinit.cpp
@@ -11,6 +11,10 @@
 #include <algorithm>
 #include <stdexcept>
 #include <typeinfo>
+#include <tuple>
+using namespace std;
+#include <map>
+using namespace std;
 
 #include <wx/checkedlistctrl.h>
 #include <wx/clrpicker.h>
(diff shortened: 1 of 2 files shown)


rkitover:
Hi @o-i-z-y-s, sorry I have not been following PRs lately, but anyway, this code will need some work, I will leave more detailed comments later.

If the default F key bindings are an issue, we should disable them by default, and you need to hook this system into the configuration.

To do that:
1. Add the option to `src/common/ConfigManager.h` and `src/common/ConfigManager.cpp`.
2. Add the option to `src/wx/opts.cpp`.
3. Make sure the GUI updates the option value.

rkitover:
@o-i-z-y-s Hey, are you still interested in working on this? I can clean this up and merge it myself if you want.

PR #925, opened 2022-03-04
Updating to YML Issue form for Bug, adding Feature request form, link to IRC and disbling blank issues.

Squall-Leonhart:
As title


ZachBacon (changes requested):
Since the colourizing hack isn't template related, this should be in a different pull request.

Squall-Leonhart:
The revert makes it in line with current master, something went wrong with a fetch so i had that line still on my end, it isn't present in master.

Squall-Leonhart:
ok, i squashed the 2 commits so the history of the file won't change, i can't make the commit entry go though, but this only touches the 4 template files.

ZachBacon:
> ok, i squashed the 2 commits so the history of the file won't change, i can't make the commit entry go though, but this only touches the 4 template files.

I'll merge it once you deem it ready then!

Squall-Leonhart:
Ok, i think this is satisfactory

ZachBacon:
lgtvm

Issue #928, opened 2022-03-08
Undefined behavior in panel.cpp (but where) 


Squall-Leonhart:
Delving into the x86 Opengl crashes that arose once @rkitover enabled code optimisations, i've narrowed it down to the block of code between these two locations

https://github.com/visualboyadvance-m/visualboyadvance-m/blob/252b65b13ce12be7234c25c05dac5961d6c39f61/src/wx/panel.cpp#L2217 - https://github.com/visualboyadvance-m/visualboyadvance-m/blob/252b65b13ce12be7234c25c05dac5961d6c39f61/src/wx/panel.cpp#L2335

i don't understand the code at all, so it'd be best to hand this over to someone else.

Squall-Leonhart:
The issue is within this block of code

https://github.com/visualboyadvance-m/visualboyadvance-m/blob/252b65b13ce12be7234c25c05dac5961d6c39f61/src/wx/panel.cpp#L2310

PR #929, opened 2022-03-08
fix x86 builds crashing when /Gs set by Optimizations, turns on full optimizations for x86

Squall-Leonhart:
turns out bools are not BOOLS, and PFNWGLSWAPINTERVALEXTPROC is a BOOL

also glGetExtensionsStringEXT expects a Const Char*

Resolves a case of undefined behavior that allows x86 builds to run as fast as x64.


Squall-Leonhart:
closes #928

ZachBacon:
lgtm

PR #951, opened 2022-05-14
  encourage users to search for existing issues in bug report template

Squall-Leonhart:
encourage users to search for existing issues


ZachBacon:
Lgtvm

PR #962, opened 2022-06-09
Game Recording Bugfix & New Movie Format Support

laqieer:
# Fix bug
## Problem
Menu state doesn't reset correctly after playtrack ended, and `stop playing movie` button doesn't work as expected, so we are trapped here.
image

## Reason
1. When playback ended, `game_playback` is set to false.
2. `systemStopGamePlayback` checks `game_playback`:
- Branch A: if it is false, returns directly <------ trapped
- Branch B: if it is true, sets it to false and resets menu state
3. Therefore, menu state cannot reset correctly.
## Solution
There are 2 solutions:
- Solution A: when playback ended, calls `systemStopGamePlayback`
- Solution B: remove Branch A of `systemStopGamePlayback`

I prefer Solution A because it resets the menu state automatically when playtrack ended so that you needn't to stop it manually.
## Screenshot to prove bug fixed
image

# Support new format
## Diff
- Old format: records time of key input since the begin of the movie: at time 1, press button A; at time 2 press button B; ...
- New format: records time of key input since the previous key input: wait time 1, press button A; wait time 2, press button B; ...
## Reason
- It is easier to edit, concat and cut the movie in the new format in some cases.
- It is common practice to record key input time with wait/delay time in automation.
- Recorded time won't increase infinitely and overflows max int like the old format.
## Implementation
|Format|Key Input|Save State|
|---|---|---|
|Old|`.vmv`|`.vm0`|
|New|`.vmd`|`.vm1`|

image
It adds the new format to the end of the drop-down list and keeps the old format as default.

### DIFF
--- a/po/wxvbam/wxvbam.pot
+++ b/po/wxvbam/wxvbam.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-07-23 17:25-0700\n"
+"POT-Creation-Date: 2022-08-06 20:51+0800\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
@@ -358,7 +358,7 @@ msgid "Text files (*.txt;*.log)|*.txt;*.log|"
 msgstr ""
 
 #: viewers.cpp:562 viewers.cpp:772 gfxviewers.cpp:1600 gfxviewers.cpp:1742
-#: cmdevents.cpp:1157 cmdevents.cpp:1235 cmdevents.cpp:1305 cmdevents.cpp:1332
+#: cmdevents.cpp:1157 cmdevents.cpp:1235 cmdevents.cpp:1305 cmdevents.cpp:1374
 #: viewsupt.cpp:1180
 msgid "Select output file"
 msgstr ""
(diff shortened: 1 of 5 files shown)


rkitover:
@laqieer Hey, I'm sorry for the very long delay in getting back to this. Can you please rebase on master and force push your branch and I'll try to take a look at everything and/or merge it. Thank you.

Also, one question, why is the new format called "vmd - diff format". What does "diff" stand for? Also, would it make more sense to make it the new default format? Could we keep it as `.vmv` extension and auto-detect the new and old formats instead?

rkitover:
I meant to say, life is very difficult for me right now, which is why I haven't looked at this yet. Sorry.

laqieer:
> Can you please rebase on master and force push your branch

Done.

> why is the new format called "vmd - diff format"

I can't come up with a better name to describe it.

> What does "diff" stand for?

Time diff.

> would it make more sense to make it the new default format?

I think it is good but you may want to include it in release note if you change the default format.

> Could we keep it as `.vmv` extension and auto-detect the new and old formats instead?

In fact, both formats are just raw data. If we keep the same extension, we will need to add a file header or something like that to specify the format version. It is easy to add a header in the new format, but the old format doesn't have it.

rkitover:
> In fact, both formats are just raw data. If we keep the same extension, we will need to add something like a file header to specify the format version. It is easy to add a header in the new format, but the old format doesn't have it.

In that case, could we check for the existence of a header, and assume v1 if there is no header?

Then the selection list could be:

- VBA Movie v2, Time Diff Format (*.vmv)
- VBA Movie v1, Old Version for Compatibility (*.vmv)

. What do you think?

Also, please join us on the new [dev discussions](#981) page and if you want our new [bridged discord](https://discord.gg/gQTP3BsK).

laqieer:
LGTM. Just confirm that the data for new header never happens to appear at the head of raw data by any chance.

Thanks for your invitation. I have joined both. The link to the dev discussion page seems to be incorrect, and it seems to be [here](https://github.com/visualboyadvance-m/visualboyadvance-m/discussions/981).

rkitover:
I think if you include some text strings in the header, then the chances of it matching raw movie data would be practically zero.

laqieer:
image
Done.

rkitover:
Thanks, looks good to me, I will just look over the code a bit and merge this. I don't claim to understand movie recording, I barely fixed a minor bug in our ffmpeg engine.

rkitover:
I see you made the header the 32 bit value `2`.

Could you make it something like:

```console
VBA Movie Version 2
```

, or something like that. I think that would be safer and better. Then when we do version 3 we can change that and add other fields if needed etc..

rkitover:
One more thing, can you please use range-based for loops to iterate over vectors and other data structures. for example:

```cpp
for (auto&& elem: vec_var) {
    std::cout << elem << std::endl;
}
```

This also works with your formats array: https://godbolt.org/z/WsEGfn3dc

laqieer:
That header is safe because previous format already has it. There is no special reason to change it. It is my bad to provide the incorrect information. It's been so long that my memory doesn't have to serve me correctly.

laqieer:
It looks nice to implement loop in that way, but it is not consistent with code in other places. I suggest to keep the same code style unless you fix it in all places.

rkitover:
> That header is safe because previous format already has it. There is no special reason to change it. It is my bad to provide the incorrect information before. It's been so long that my memory doesn't serve me correctly.

Ok, works for me then 👍

> It looks nice to implement loop in that way, but it is inconsistent with code in other places. I suggest to keep the same code style unless you fix it in all places.

That code was written by denisfa who did a lot of work for this project. He's a C programmer so he wasn't aware of these modern C++ features, but we are definitely moving towards using modern C++ where possible. This project currently builds against the C++17 standard.

And if you think that code is ugly, look at some of the code in the emulator core.

Changing everything everywhere to nicer code is a very low priority, or even fixing warnings, but at least we can use nicer code in additions/improvements where possible.

laqieer:
Nice. Fixed.

rkitover:
Great, if you don't mind I will merge this manually to fix up the Git history.

In the future please try to write nice Git commit titles and messages describing what you're doing.

I will update our dev manual with all of this now that we actually have devs again.

rkitover:
Can you please check over the [movie-format](https://github.com/visualboyadvance-m/visualboyadvance-m/tree/movie-format) branch, the two commits at the tip are your work.

Please check that I didn't break anything and that everything looks good to you and I will merge to master.

Thanks.

laqieer:
https://github.com/visualboyadvance-m/visualboyadvance-m/pull/962/commits/d92027c8d656b51fa427befc09d95694baba3e6e
Nothing breaks.

rkitover:
Squashed your latest commit into the movie format commit.

Good to merge now?

laqieer:
good

rkitover:
Merged, thank you!

Issue #745, opened 2020-09-08
Improve wxSDLJoy handling of controllers.


Steelskin:
Following #742, there is still room for improvements in the way `wxSDLJoy` handles controllers, especially reconnects.
Currently, we do not have a good way to ensure a reconnected joystick is going to be the same as the previous one. In addition, usage of `wxSDLJoyEvent` could be improved by abstracting what is referred to as the `player_index`. The latter should really be treated as an abstract key that refers to a joystick. The tricky part is matching that key back to a joystick in a future session or in case of a reconnect.

High-level summary of things that were discussed on PR #742:
* Keep a map of joystick keys to active joysticks. Or more simply a vector of joysticks where the index corresponds to a joystick key if we simply use an integer.
* Make use of the `*PlayerIndex` API in SDL. This is going to require updating SDL, maybe linking statically against it on all platforms to avoid weird conditional blocks.
* Abstract the `player_index` value in `wxSDLJoyEvent` into a `wxSDLJoyKey`. This is going to require refactoring of all the classes that use it.

Select quotes:
@briansrls said
> This is correct. I would actually argue we should not use the SDL joystick index at all past the initially connection. I don't think storing them is valuable to us.
>
>For example, in my implementation I was thinking the player_index should be managed as follows:
> We have a maximum of N players (say, 4 players). We have a simple boolean array of length 4:
>
> [false, false, false, false]
>
> This indicates to us if we have any players active on that player_index. Now, we connect the first controller, it has SDL index 0, instance ID 1. We assign it to player_index 0 using SDL_GameControllerSetPlayerIndex.
> 
> [true, false, false, false]
>
> Now, we accept another controller connection. Our second player is SDL index 1, instance ID 2. We look at our map, and determine that player_index 1 is available. We set this controller to player_index 1 using SDL_GameControllerSetPlayerIndex.
>
> [true, true, false, false]
> 
> Now, say player 1 disconnects. Sure, the SDL indexes will be sliding around. Player 2's SDL index will slide from index 1 to index 0, like you described. However, it will still have player_index 1. This should be the source of truth, along with its instance ID.
>
> After player 1 disconnects, we clear that hole in our table:
>
> [false, true, false, false]
>
> Now, whatever controller connects next, will get player_index 0, and fill the hole we created.
> 
> [true, true, false, false]
>
[...]

PR #971, opened 2022-07-10
Introduce wxUserInput class

Steelskin:
This introduces a new abstraction for any user input. The long-term goal
is to replace every usage of "custom" {key, mod, joy} triplets in the
code base with this new class.

This class implements comparison operators, allowing for faster access
in a set or as a key in a map, compared to the vectors currently used.

Issue: #745

### DIFF
--- a/src/wx/CMakeLists.txt
+++ b/src/wx/CMakeLists.txt
@@ -728,6 +728,7 @@ set(
     wxutil.cpp
     widgets/keyedit.cpp
     widgets/joyedit.cpp
+    widgets/userinput.cpp
     widgets/sdljoy.cpp
     widgets/wxmisc.cpp
     # probably ought to be in common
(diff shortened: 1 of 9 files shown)


Steelskin:
After that, it's mostly a matter of converting every usage of joy, key, mod and every call site.

rkitover:
@Steelskin not really in a condition to read this stuff right now, if there are no functionality changes I'll trust your judgement on this.

Steelskin:
The devil is in the details but the only functional change is that the global `keys_pressed` vector has been replaced with a set. This should be a bit faster and functionality should otherwise be identical. I tested a few things (different controllers, keyboard bindings. ...) to ensure there are no regressions.

rkitover:
Awesome, thank you.

PR #974, opened 2022-07-23
Refactor game controls bindings

Steelskin:
This introduces abstractions for a game key (`wxGameKey`), a game
control (`wxGameControl`) and a common global handler for in-game
user input processing (`wxGameControlState`).
User configuration is changed from a vector to a map of `wxGameControl`
to a set of `wxUserInput`, which simplifies input configuration updates.

User input processing for in-game controls is now unified between
keyboard and joypad input, and is much faster in general since access
to game control state is now always logarithmic rather than linear.
This comes at the expense of slightly slower user input configuration
updates. However, in the worst case scenario, this is still done in
O(log(n)).

This removes all uses of `wxJoyKeyBinding`. However, some uses of the
key, mod, joy triplets remain and will be cleaned up in follow-up PRs.

Issue: #745

### DIFF
--- a/po/wxvbam/wxvbam.pot
+++ b/po/wxvbam/wxvbam.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2021-11-24 21:24+0000\n"
+"POT-Creation-Date: 2022-07-23 17:25-0700\n"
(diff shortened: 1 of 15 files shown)


Steelskin:
Sorry this one ended up being rather large. I can try to break it down into smaller chunks if you want.

rkitover:
@Steelskin Thank you very much for working on this stuff, I will try to read all of it when I get a chance.

rkitover:
And sorry this took so long, I was meaning to read these but I have serious problems in life right now. I will try to read all of your changes when I can.

Issue #158, opened 2017-10-23
Full screen f11 does not work after rom loading


wingedonezero:
### Expected behavior

F11 going full screen

### Actual behavior 

Nothing only goes full screen with no rom loaded

### Steps to reproduce the behavior

Hit f11 with rom loaded and without

### ROM details

Which games are affected by the issue (please state if GBA or GB game):

All games gb, gbc and gba

### Option details

Video driver (simple, opengl):

Simple and opengl same

Operating System (Windows, Mac, Linux (state distribution), etc.):

Windows 10

Program version 2.0.1

RhinoKneel:
I've actually found this to be with all keyboard shortcuts after loading. Can't reset, save state, quick switch game, nothing.

wingedonezero:
Yea seems to be the case i didn't try any of the others but i have found that a older version i had that was beta 4 works fine. So it had to be something that changed between then and this release.

BanCrash:
I'd have the same issue. I found out that, for default, the key for screenshot and for full screen it's the same (F11). I change on Options -> Input -> Configure -> Special, the key for screenshots from F11 to F12. Now it seems to work properly.

BanCrash:
Well, if I launch the game from an external application (Launchbox) seems to not work neither.

wingedonezero:
I can confirm what you mentioned clearing all inputs resetting the keys and adding f12 to the screenshot section fixes this issue. It must be a problem with mixed up hot key defaults then with full screen and screenshot.

rkitover:
I had a few weeks ago removed the `F11` mapping for "save state".

Currently, as far as I can tell, the only thing `F11` is mapped to by default is "full screen" on Windows and Linux.

What is probably happening here, is that you guys have your old configs from some time when this mapping existed and it got saved into the config.

I verified that `F11` invokes entry into full screen on a fresh Windows build with no config.

BanCrash:
I try with a fresh installation and can confirm what you said. I never mapped a key for screenshot, that's why I thought this would be a problem with the build.

Btw, sorry for open a duplicate issue, I don't know why I thought I searched for that issue too yesterday when I was searching for this issue. My bad.

rkitover:
Well, how would you suggest we fix this, I was thinking I could check if any configured keymappings conflict with app keymappings and remove them or something.

BanCrash:
That sounds pretty well to me. 
[...]

PR #998, opened 2022-09-07
[options] Refactor INI options

Steelskin:
This change introduces a new class to handle all of the INI options data, VbamOption. A VbamOption represents a single option in the INI file. It is not constructible outside of its implementation, which prevents the initialization of incorrectly formatted options.

Internally, a VbamOption points to a global variable in memory, which holds the value used at runtime. This is because various parts of the codebase edit the global variable rather than going through a central registry. This also means we need to separately update the INI values.

In the future, we may be able to convert all existing reads and writes to the global variable to go through VbamOption. Individual UX elements could link directly to getters/setters for a specific VbamOption rather than have duplicate data.

VbamOption replaces the opt_desc struct and the global opts array. All users of opt_desc and the global opts array have been updated.

This is a necessary preliminary change to better support a refactor of accelerators, through wxUserInput.

Issue: #745, #158

### DIFF
--- a/src/wx/guiinit.cpp
+++ b/src/wx/guiinit.cpp
@@ -2978,10 +2978,14 @@ bool MainFrame::BindControls()
                     checkable_mi.push_back(cmi);
 
                     for (const VbamOption& option : VbamOption::AllOptions()) {
-                        if (option.is_int()) {
-                            MenuOptionIntMask(option.command(), option.GetInt(), (1 << 0));
-                        } else if (option.is_bool()) {
-                            MenuOptionBool(option.command(), option.GetBool());
+                        if (cmdtab[i].cmd == option.command()) {
+                            if (option.is_int()) {
+                                MenuOptionIntMask(
+                                    option.command(), option.GetInt(), (1 << 0));
+                            } else if (option.is_bool()) {
+                                MenuOptionBool(
+                                    option.command(), option.GetBool());
+                            }
                         }
                     }
                 }
@@ -3692,8 +3696,11 @@ bool MainFrame::BindControls()
             getsc("MaxScale", maxScale);
             /// Basic
             getrbi("OutputSimple", gopts.render_method, RND_SIMPLE);
+
+#if defined(__WXMAC__)
             getrbi("OutputQuartz2D", gopts.render_method, RND_QUARTZ2D);
-#if !defined(__WXMAC__)
+#else
+            rb = SafeXRCCTRL<wxRadioButton>(d, "OutputQuartz2D");
             rb->Hide();
 #endif
             getrbi("OutputOpenGL", gopts.render_method, RND_OPENGL);
@@ -3706,10 +3713,11 @@ bool MainFrame::BindControls()
                 rb->Hide();
             }
 #endif
-            getrbi("OutputDirect3D", gopts.render_method, RND_DIRECT3D);
-#if !defined(__WXMSW__) || defined(NO_D3D) || 1 // not implemented
+
+            // Direct3D is not implemented so hide the option on every platform.
+            rb = SafeXRCCTRL<wxRadioButton>(d, "OutputDirect3D");
             rb->Hide();
-#endif
+
             ch = GetValidatedChild<wxChoice, wxGenericValidator>(d, "Filter", wxGenericValidator(&gopts.filter));
 
             // Save the Filters choice control to extract the names from the XRC.
--- a/src/wx/guiinit.cpp
+++ b/src/wx/guiinit.cpp
@@ -24,6 +24,7 @@
 #include <wx/wfstream.h>
 
 #include "opts.h"
+#include "vbam-options.h"
 #include "wx/gamecontrol.h"
 #include "wx/userinput.h"
 #include "../gba/CheatSearch.h"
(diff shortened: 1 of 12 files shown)


Steelskin:
Okay, I apologize but this one is big. The most important part is the conversion of the opt_desc struct into the new one in vbam-option-static.cpp. The rest should be easier to review since most of the complexity was moved to VbamOption.

Also, the description for the options was not properly marked for translation since the `wxTRANSLATE` macro does not actually add to the translation array, this is now fixed. The side effect is that it adds a ton of strings to the translation.

I would appreciate if you could see if macOS is complaining about anything here, in particular about the use of `std::variant`. We had issues with `std::optional` and I hope we won't encounter anything like that here.

ZachBacon:
Keep in mind, the SDL binary also makes use of the same ini format.

src/wx/vbam-options-static.cpp
```diff
@@ -0,0 +1,830 @@
...
+constexpr size_t kNbAudioApis = static_cast<size_t>(AudioApi::kLast);
+
+// These MUST follow the same order as the definitions of the enum above.
+// Adding an option without adding to this array will result in a compiler
+// error since kNbAudioApis is automatically updated.
+static const std::array<wxString, kNbAudioApis> kAudioApiStrings = {
+    "sdl",
+    "openal",
+    "directsound",
+    "xaudio2",
```
rkitover, line 148: directsound and xaudio2 are for Windows only, don't know if that matters for your implementation.

src/wx/vbam-options-static.cpp
```diff
@@ -0,0 +1,830 @@
...
+        { kFilterStrings[13], FilterFunction::kSimple3x }, 
+        { kFilterStrings[14], FilterFunction::kHQ3x }, 
+        { kFilterStrings[15], FilterFunction::kSimple4x }, 
+        { kFilterStrings[16], FilterFunction::kHQ4x }, 
+        { kFilterStrings[17], FilterFunction::kXbrz2x }, 
+        { kFilterStrings[18], FilterFunction::kXbrz3x }, 
+        { kFilterStrings[19], FilterFunction::kXbrz4x }, 
+        { kFilterStrings[20], FilterFunction::kXbrz5x }, 
+        { kFilterStrings[21], FilterFunction::kXbrz6x }, 
+        { kFilterStrings[22], FilterFunction::kPlugin },
```
rkitover, line 663: Any way to build this map iteratively?

src/wx/vbam-options-static.cpp
```diff
@@ -0,0 +1,830 @@
...
+#include "opts.h"
+#include "wx/stringimpl.h"
+
+#define VBAM_OPTIONS_INTERNAL_INCLUDE
+#include "vbam-options-internal.h"
+#undef VBAM_OPTIONS_INTERNAL_INCLUDE
+
+namespace {
+
+// This enum must be kept in sync with the one in wxvbam.h
```
rkitover, line 20: We'll need to unify these, either now or later, duplication is bad.

Perhaps the string list could be taken from the xrc and the enum could be generated from the xrc.

src/wx/vbam-options-static.cpp
```diff
@@ -0,0 +1,830 @@
...
+        return 0;
+    }
+    return static_cast<int>(iter->second);
+}
+
+int StringToInterframe(const wxString& config_name, const wxString& input) {
+    static const std::map<wxString, Interframe> kStringToInterframe = {
+        { kInterframeStrings[0], Interframe::kNone },
+        { kInterframeStrings[1], Interframe::kSmart },
+        { kInterframeStrings[2], Interframe::kMotionBlur },
```
rkitover, line 682: Likewise.

src/wx/vbam-options-static.cpp
```diff
@@ -0,0 +1,830 @@
...
+    }
+    return static_cast<int>(iter->second);
+}
+
+int StringToSoundQuality(const wxString& config_name, const wxString& input) {
+    static const std::map<wxString, SoundQuality> kStringToSoundQuality = {
+        { kSoundQualityStrings[0], SoundQuality::k48kHz },
+        { kSoundQualityStrings[1], SoundQuality::k44kHz },
+        { kSoundQualityStrings[2], SoundQuality::k22kHz },
+        { kSoundQualityStrings[3], SoundQuality::k11kHz },
```
rkitover, line 746: And these three.

src/wx/guiinit.cpp
```diff
@@ -2973,17 +2974,14 @@ bool MainFrame::BindControls()
...
-                    for (int j = 0; j < num_opts; j++) {
-                        wxString menuName = wxString(opts[j].cmd, wxConvUTF8);
-
-                        if (menuName == cmdtab[i].cmd) {
-                            if (opts[j].intopt)
-                                MenuOptionIntMask(opts[j].cmd, *opts[j].intopt, (1 << 0));
-                            else if (opts[j].boolopt)
-                                MenuOptionBool(opts[j].cmd, *opts[j].boolopt);
+                    for (const VbamOption& option : VbamOption::AllOptions()) {
+                        if (option.is_int()) {
```
rkitover, line 2981: What about `double`? Or is that not relevant here?

src/wx/vbam-options.h
```diff
@@ -0,0 +1,272 @@
...
+    kSoundGBEcho,
+    kSoundGBEnableEffects,
+    kSoundGBStereo,
+    kSoundGBSurround,
+    kSoundQuality,
+    kSoundVolume,
+
+    // Do not add anything under here.
+    Last,
+};
```
rkitover, line 146: Could we put the string map here and add an assert for the count matching for sanity checking? I don't really like the duplication between the enum/string-map, but there is probably no better way to do that.

rkitover:
Looks great. I'll review your previous PRs as well when I get a chance, I wasn't able to then because of my shitty situation.

I left some comments.

@ZachBacon This doesn't change anything with regard to `ConfigManager`, we can still add shared config entries there.

rkitover (changes requested):

src/wx/guiinit.cpp
Steelskin: I don't really understand what this code is doing to be honest. The old code was not using `double` so I didn't add it here either.
I will revisit when I have a better understanding of what is happening here.

src/wx/vbam-options-static.cpp
Steelskin: Yes, and that's actually better for long term maintenance, though we lose the const qualifier. Oh well, the one-time runtime cost is very low anyway.

src/wx/vbam-options-static.cpp
Steelskin: Done

src/wx/vbam-options-static.cpp
Steelskin: ditto

src/wx/vbam-options.h
Steelskin: I created the string map iteratively using `kAllOptionsData`, so we're no longer duplicating the data.
I still would like to find a better way of handling this down the line, we are redefining too many things for my tastes too.

src/wx/guiinit.cpp
Steelskin: I separately spent some time on code that makes use of this so I understand what is happening here better. This code initializes the checkable menu items to be checked or not and then spews a warning if we missed one - that's what the `initialized` field handles. So I believe we don't need to handle `double` here.

PR #1002, opened 2022-09-19
Interframe blending changer OSD improvement

rkitover:
This puts the name of the interframe blending algorithm in the OSD instead of the index on activation of the `Change interframe blending` menu entry, as requested by @Squall-Leonhart.

This also sets the default accelerator for this menu entry `CTRL+I`.

I am also doing a minor refactor of my `strutils` library, @Steelskin sanity check please?

### DIFF
--- a/src/wx/cmdevents.cpp
+++ b/src/wx/cmdevents.cpp
@@ -1,4 +1,5 @@
 #include "wxvbam.h"
+
 #include <algorithm>
 #include <wx/aboutdlg.h>
 #include <wx/ffile.h>
@@ -10,6 +11,7 @@
 #include <wx/wfstream.h>
 #include <wx/msgdlg.h>
 
+#include "strutils.h"
 #include "../common/version_cpp.h"
 #include "../gb/gbPrinter.h"
 #include "../gba/agbprint.h"
@@ -2766,7 +2768,7 @@ EVT_HANDLER_MASK(ChangeFilter, "Change Pixel Filter", CMDEN_NREC_ANY)
     }
 
     wxString msg;
-    msg.Printf(_("Using pixel filter %s"), pixel_filters->GetString(gopts.filter));
+    msg.Printf(_("Using pixel filter: %s"), pixel_filters_->GetString(gopts.filter));
     systemScreenMessage(msg);
 }
 
@@ -2781,7 +2783,7 @@ EVT_HANDLER_MASK(ChangeIFB, "Change Interframe Blending", CMDEN_NREC_ANY)
     }
 
     wxString msg;
-    msg.Printf(_("Using interframe blending #%d"), gopts.ifb);
+    msg.Printf(_("Using interframe blending: %s"), interframe_blenders_->GetString(gopts.ifb));
     systemScreenMessage(msg);
 }
 
--- a/src/wx/strutils.cpp
+++ b/src/wx/strutils.cpp
@@ -1,10 +1,13 @@
 #include "strutils.h"
+
 #include <wx/tokenzr.h>
+
+namespace strutils {
 
 // From: https://stackoverflow.com/a/7408245/262458
 //
 // Modified to ignore empty tokens or return sep for them.
-wxArrayString str_split(const wxString& text, const wxString& sep, bool empty_token_is_sep) {
+wxArrayString split(const wxString& text, const wxString& sep, bool empty_token_is_sep) {
     wxArrayString tokens;
     size_t start = 0, end = 0;
 
@@ -29,11 +32,9 @@ wxArrayString str_split(const wxString& text, const wxString& sep, bool empty_to
     return tokens;
 }
 
-wxArrayString str_split_with_sep(const wxString& text, const wxString& sep)
+wxArrayString split_with_sep(const wxString& text, const wxString& sep)
 {
-    return str_split(text, sep, true);
+    return split(text, sep, true);
 }
 
-size_t vec_find(wxArrayString& opts, const wxString& val) {
-    return opts.Index(val);
-}
+} // namespace strutils
--- a/src/wx/strutils.h
+++ b/src/wx/strutils.h
@@ -7,15 +7,16 @@
 #include <wx/string.h>
 #include <wx/arrstr.h>
 
+namespace strutils {
+
 // From: https://stackoverflow.com/a/7408245/262458
-wxArrayString str_split(const wxString& text, const wxString& sep, bool empty_token_is_sep=false);
+wxArrayString split(const wxString& text, const wxString& sep, bool empty_token_is_sep=false);
 
 // Same as above, but it includes the sep dir.
 // If "A,,,B" is the text and "," is sep, then
 // 'A', ',' and 'B' will be in the output.
-wxArrayString str_split_with_sep(const wxString& text, const wxString& sep);
+wxArrayString split_with_sep(const wxString& text, const wxString& sep);
 
-// From: https://stackoverflow.com/a/15099743/262458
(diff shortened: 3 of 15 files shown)


src/wx/cmdevents.cpp
```diff
@@ -2766,10 +2769,26 @@ EVT_HANDLER_MASK(ChangeFilter, "Change Pixel Filter", CMDEN_NREC_ANY)
     }
 
     wxString msg;
-    msg.Printf(_("Using pixel filter %s"), pixel_filters->GetString(gopts.filter));
+    msg.Printf(_("Using pixel filter:") + " %s", pixel_filters->GetString(gopts.filter));
```
Steelskin, line 2772: Not every language will put the value at the end of the sentence, so the formatter should be part of the string to localize.

src/wx/cmdevents.cpp
```diff
@@ -2766,10 +2769,26 @@ EVT_HANDLER_MASK(ChangeFilter, "Change Pixel Filter", CMDEN_NREC_ANY)
...
     systemScreenMessage(msg);
 }
 
+static const wxString&
+ifb_short_name(const wxChoice* ifbs, unsigned num)
+{
+    static std::map<unsigned, wxString> cache;
```
Steelskin, line 2779: I'd suggest `std::unordered_map` instead, we don't care about ordering.

src/wx/strutils.h
```diff
@@ -7,15 +7,23 @@
 #include <wx/string.h>
 #include <wx/arrstr.h>
 
+namespace strutils
+{
```
Steelskin, line 11: I'm used to putting the opening curly brace on the same line, it saves on vertical space. I don't feel very strongly on the matter, though.

src/wx/cmdevents.cpp
```diff
@@ -2780,8 +2799,10 @@ EVT_HANDLER_MASK(ChangeIFB, "Change Interframe Blending", CMDEN_NREC_ANY)
         panel->panel = nullptr;
     }
 
+    auto ifb_name = ifb_short_name(interframe_blenders, gopts.ifb);
+
     wxString msg;
-    msg.Printf(_("Using interframe blending #%d"), gopts.ifb);
+    msg.Printf(_("Using interframe blending:") + " %s", ifb_name);
     systemScreenMessage(msg);
```
Steelskin, line 2806: You can use `wxString::Format` directly rather than using a variable.

src/wx/wxvbam.h
```diff
@@ -382,7 +382,8 @@ class MainFrame : public wxFrame {
     // Load a named wxDialog from the XRC file
     wxDialog* LoadXRCropertySheetDialog(const char* name);
 
-    wxChoice* pixel_filters = nullptr;
+    wxChoice* pixel_filters       = nullptr;
+    wxChoice* interframe_blenders = nullptr;
```
Steelskin, line 386: I'd suggest `interframe_blenders_` with a trailing `_`. It makes it easier to see we are using a private member here and not a variable. I know this is a style change from the rest of the file but I feel like it really helps readability and maintainability.

src/wx/cmdevents.cpp
Steelskin: Also, put the `%s` formatter in the localized string too. As above, some languages don't follow the same word order.

src/wx/cmdevents.cpp
```diff
@@ -2766,10 +2769,26 @@ EVT_HANDLER_MASK(ChangeFilter, "Change Pixel Filter", CMDEN_NREC_ANY)
...
 }
 
+static const wxString&
+ifb_short_name(const wxChoice* ifbs, unsigned num)
+{
+    static std::map<unsigned, wxString> cache;
+
+    if (cache.count(num))
+        return cache[num];
```
Steelskin, line 2782: You end up doing 2 lookups here. Using `find()` is probably what you want, even if it means adding the usual boilerplate to check against `end()`.

src/wx/cmdevents.cpp
```diff
@@ -2766,10 +2769,26 @@ EVT_HANDLER_MASK(ChangeFilter, "Change Pixel Filter", CMDEN_NREC_ANY)
     }
 
     wxString msg;
-    msg.Printf(_("Using pixel filter %s"), pixel_filters->GetString(gopts.filter));
+    msg.Printf(_("Using pixel filter:") + " %s", pixel_filters->GetString(gopts.filter));
     systemScreenMessage(msg);
 }
 
+static const wxString&
+ifb_short_name(const wxChoice* ifbs, unsigned num)
```
Steelskin, line 2777: `assert` that `num` is in the correct range.

src/wx/cmdevents.cpp
```diff
@@ -2766,10 +2769,26 @@ EVT_HANDLER_MASK(ChangeFilter, "Change Pixel Filter", CMDEN_NREC_ANY)
...
+{
+    static std::map<unsigned, wxString> cache;
+
+    if (cache.count(num))
+        return cache[num];
+
+    // Shorten and clean up name.
+    cache[num] = strutils::trim(strutils::titlecase(
+        strutils::remove_word(strutils::remove_word(ifbs->GetString(num).Lower(), _("interframe")), _("blending"))
```
Steelskin, line 2786: There's only so many values. Could we just populate the map on first call and simply return the value on further calls? If we're doing that, we can probably use a simple `vector`.

Steelskin, line 2786: This is going to be confusing for localizers. Without context, it is not clear what these 2 strings are referring to and that they need to be identical to how "interframe blending" was translated elsewhere. In addition, not every language will translate that expression in 2 words, some may use 3 or more.
The case for the remaining word is also different than in its original context, which has different inflections in some languages.
Finally, not every language uses spaces for word delimiters.

Instead, I would suggest hardcoding the strings we want here. This is not duplication in the sense that the difference of context means they are different localizable strings even though they happen to be the same in English.

src/wx/strutils.cpp
```diff
@@ -29,11 +34,44 @@ wxArrayString str_split(const wxString& text, const wxString& sep, bool empty_to
...
 }
 
-size_t vec_find(wxArrayString& opts, const wxString& val) {
-    return opts.Index(val);
+wxString trim(const wxString& in_str)
+{
+    std::wstring res{in_str};
+
+    res = std::regex_replace(res, std::wregex{L"^\\s+"}, L"");
+    res = std::regex_replace(res, std::wregex{L"\\s+$"}, L"");
```
Steelskin, line 47: All of these use regex, which is pretty slow. But see my other comment, we may not need these.

src/wx/cmdevents.cpp
```diff
@@ -1,5 +1,7 @@
 #include "wxvbam.h"
+
 #include <algorithm>
+#include <unordered_map>
```
Steelskin, line 4: nit: this include is no longer needed

src/wx/strutils.h
```diff
@@ -7,15 +7,22 @@
...
 // If "A,,,B" is the text and "," is sep, then
 // 'A', ',' and 'B' will be in the output.
-wxArrayString str_split_with_sep(const wxString& text, const wxString& sep);
+wxArrayString split_with_sep(const wxString& text, const wxString& sep);
+
+wxString trim(const wxString& in_str);
+
+wxString remove_word(const wxString& in_str, const wxString& word);
+
+wxString titlecase(const wxString& in_str);
```
Steelskin, line 24: These are no longer used. We could still keep them around in case we need them down the line but I think we're better off removing dead code.

Steelskin (approved):
Thanks! lgtm with a couple nits.

PR #1010, opened 2022-09-27
[build] Use modern CMake functions for the MSVC build

Steelskin:
* Remove all defaults for CMAKE_C*_FLAGS_*
* Use /W4 on every configuration.
* Use /O1 for every release build, other than MinSizeRel.
* Optimize link for every release build.


CMakeLists.txt
```diff
@@ -29,6 +29,19 @@ if(COMMAND cmake_policy)
...
+    endif()
+
+    if(POLICY CMP0092)
+        cmake_policy(SET CMP0092 NEW) # Do not set /w for MSVC.
+    endif()
+
+    if(POLICY CMP0117)
+        cmake_policy(SET CMP0117 NEW) # Do not set /GR for MSVC.
+    endif()
+
```
Neumann-A, line 44: So much nonsense. Just put the `cmake_minimum_required` at the top of the CMakeList and set the version to something reasonable. This monstrosity of build script is already using at least 3.1 features.

CMakeLists.txt
```diff
@@ -801,43 +812,59 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID STREQUAL Clang)
...
+    set(CMAKE_CXX_FLAGS_DEBUG "")
+    set(CMAKE_CXX_FLAGS_RELEASE "")
+    set(CMAKE_CXX_FLAGS_MINSIZEREL "")
+    set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "")
+
+    # Set the right MSVC Runtime library.
+    set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:Debug>:Debug>")
+    if(NOT VBAM_STATIC)
+        set(CMAKE_MSVC_RUNTIME_LIBRARY "${CMAKE_MSVC_RUNTIME_LIBRARY}Dll")
+    endif()
```
Neumann-A, line 832: Use a toolchain file instead. The CMakeLists.txt shouldn't deal with any flags !

CMakeLists.txt
```diff
@@ -801,43 +812,59 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID STREQUAL Clang)
...
 
-    add_compiler_flags(/std:c++17 -D__STDC_LIMIT_MACROS /fp:fast /Oi)
+    # We use C++17
+    set(CMAKE_CXX_STANDARD 17)
```
Neumann-A, line 835: this is a cmake 3.1 feature.

src/wx/CMakeLists.txt
```diff
@@ -1008,7 +1008,7 @@ if(NOT TRANSLATIONS_ONLY)
         endif()
     elseif(MSVC)
         # the debug lib libcmtd is linked in debug mode, so don't link the normal version
-        set_target_properties(visualboyadvance-m PROPERTIES LINK_FLAGS_DEBUG "/nodefaultlib:libcmt /subsystem:console")
+        target_link_options(visualboyadvance-m PRIVATE "$<$<CONFIG:DEBUG>:/nodefaultlib:libcmt;/subsystem:console>")
```
Neumann-A, line 1011: This (/nodefaultlib) typically is an indicator that something is linking wrongly.

CMakeLists.txt
```diff
@@ -801,43 +812,59 @@ if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID STREQUAL Clang)
...
+    add_link_options("$<$<CONFIG:RELWITHDEBINFO>:/OPT:REF;/OPT:ICF>")
+
+    # MinSizeRel flags
+    add_compile_options("$<$<CONFIG:MINSIZEREL>:/O1;/Ob1;/Oy-;/Oi>")
+    add_compile_definitions("$<$<CONFIG:MINSIZEREL>:NDEBUG>")
+    add_link_options("$<$<CONFIG:MINSIZEREL>:/OPT:REF;/OPT:ICF>")
+
+    if(ENABLE_LTO)
+        add_compile_options(/GL)
+        add_link_options(/LTCG)
```
Neumann-A, line 867: This is not modern CMake! The only thing modern here is the use of generator expressions. Again my adivce: Move this stuff into a toolchain file and set `CMAKE_C(XX)_FLAGS_<CONFIG>` explicitly

src/wx/CMakeLists.txt
Squall-Leonhart: its required to to make use of Xaudio from the DXSDK rather than Windows SDK, as the former uses Xaudio2.7, rather than 2.8 which is only available in Windows 8+.

The alternative is to use Xaudio2.9 redist but that adds loose files to the package.

Steelskin:
I am abandoning this PR since all of this is landing separately.

PR #1011, opened 2022-09-27
[hidpi] Add High DPI support for all platforms

Steelskin:
wxWidgets now supports High DPI natively.


Steelskin:
Closing this. The API I was using is only available in wxWidgets 3.1. I will rework that one to use the underlying native APIs.

PR #1017, opened 2022-10-09
[hidpi] Implement full HiDPI support

Steelskin:
This converts the HiDPIAware into a series of helper functions, using wxWidgets native support starting with wxWidgets 3.1.4. Custom implementations are provided for Mac and Windows.

In addition, this fixes a few issues with the Windows build:
* Use the same defines variables in the MSVC and MinGW builds, with g++ and Clang.
* Add a custom manifest for the application, indicating full DPI support and support for recent Windows versions so the application will no longer run in compatibility mode by default.
* Manually spawn a console in debug builds if none is attached and always use the parent console if it is available otherwise, even in release builds. This also removes the /subsystem:console linker argument.


src/wx/CMakeLists.txt
```diff
@@ -89,6 +89,11 @@ if(NOT WIN32 AND NOT APPLE)
     endif()
 endif()
 
+# Win32 definitions common to all toolchains.
+if (WIN32)
+    add_definitions(-D_UNICODE -DUNICODE -DwxUSE_GUI=1 -D__WXMSW__ -DWINVER=0x0A00 -DNTDDI_VERSION=0x0A000007)
+endif()
```
rkitover, line 95: I gather this sets compatibility to win10. Do we really need this for the features that we need? Could we set compatibility to 7 or even Vista?

src/wx/visualboyadvance-m.manifest
```diff
@@ -0,0 +1,42 @@
...
+    </trustInfo>
+    <asmv3:application>
+        <asmv3:windowsSettings>
+            <dpiAwareness xmlns="http://schemas.microsoft.com/SMI/2016/WindowsSettings">PerMonitorV2, system</dpiAwareness>
+            <dpiAware xmlns="http://schemas.microsoft.com/SMI/2005/WindowsSettings">true</dpiAware>
+        </asmv3:windowsSettings>
+    </asmv3:application>
+    <compatibility xmlns="urn:schemas-microsoft-com:compatibility.v1">
+        <application>
+            <!-- Windows Vista -->
```
rkitover, line 30: Are we supporting Vista? I would be in favor of that if it doesn't cause too many problems.

rkitover (changes requested):
Looks awesome, thank you.

Just wanted to make sure `WINVER` and `NTDDI_VERSION` are the lowest we can support.

src/wx/CMakeLists.txt
Steelskin: This is confusing but `WINVER` and `NTDDI_VERSION` are used by the Windows SDK as an indicator for "I want to access APIs for Windows version up to X". There are ways to check at runtime if an API is present or not. Anyway, simply defining them does not result in Windows-10-only binaries, it merely gives access to Windows 10 APIs.
We do need to be careful when we use Windows APIs to provide fallback code paths for older versions but this is an orthogonal issue.

src/wx/visualboyadvance-m.manifest
Steelskin: The GUIDs here tell Windows to not run the program in compatibility mode. I don't know if we actually support Vista, I haven't tested it but nothing I am adding here should break compatibility.

Steelskin: I actually tested it in a VM and I can confirm this is working on Vista, probably XP too.

rkitover: Awesome!

PR #1020, opened 2022-10-13
Revert "Revert DMA:Linger last DMA on bus ..."

rkitover:
This reverts commit 173512f65aa9040fda4e97fd90b5ce4a23f59176.

Include the patch instead with one line removed that @Squall-Leonhart said fixes the bug in Croket.

### DIFF
--- a/src/gba/GBA.cpp
+++ b/src/gba/GBA.cpp
@@ -50,8 +50,9 @@ bool busPrefetchEnable = false;
 uint32_t busPrefetchCount = 0;
 int cpuDmaTicksToUpdate = 0;
 int cpuDmaCount = 0;
-bool cpuDmaHack = false;
+bool cpuDmaRunning = false;
 uint32_t cpuDmaLast = 0;
+uint32_t cpuDmaPC = 0;
 int dummyAddress = 0;
 
 bool cpuBreakLoop = false;
@@ -2343,7 +2344,7 @@ void doDMA(uint32_t& s, uint32_t& d, uint32_t si, uint32_t di, uint32_t c, int t
     int dw = 0;
     int sc = c;
 
-    cpuDmaHack = true;
+    cpuDmaRunning = true;
     cpuDmaCount = c;
     // This is done to get the correct waitstates.
     if (sm > 15)
@@ -2408,7 +2409,7 @@ void doDMA(uint32_t& s, uint32_t& d, uint32_t si, uint32_t di, uint32_t c, int t
     }
 
     cpuDmaTicksToUpdate += totalTicks;
-    cpuDmaHack = false;
+    cpuDmaRunning = false;
 }
 
 void CPUCheckDMA(int reason, int dmamask)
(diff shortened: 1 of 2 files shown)


rkitover:
@Squall-Leonhart @Steelskin please confirm before I merge this.

@negativeExponent Would love your input here.

See: https://github.com/visualboyadvance-m/visualboyadvance-m/commit/173512f65aa9040fda4e97fd90b5ce4a23f59176

Squall-Leonhart:
I would like @Normmatt to take a look at this personally

src/gba/GBAinline.h
```diff
@@ -165,7 +166,7 @@ static inline uint32_t CPUReadMemory(uint32_t address)
                 armMode ? armNextPC - 4 : armNextPC - 2);
         }
 #endif
-        if (cpuDmaHack) {
+        if (cpuDmaRunning || ((reg[15].I - cpuDmaPC) == (armState ? 4 : 2))) {
```
Steelskin, line 169: I am a bit puzzled by what this does. I just treated myself to a crash course on GBA architecture so my understanding may be wrong.
Basically, this checks if the PC register (r15) is 0x0004 in ARM mode or 0x0002 in Thumb mode, since `cpuDmaPC` is always 0. This memory range is in the BIOS, which can't be right.

If we are missing some `cpuDmaPC` initialization - presumably to the PC location when we started the DMA operation, then this check becomes "Is PC 1 instruction ahead of the DMA start?", which makes more sense with the intent of the PR.

However, in this case, if I am reading this page properly: http://problemkaputt.de/gbatek-gba-unpredictable-things.htm, it seems that this should only be done in Thumb mode, not ARM mode, so it should be in the not `armState` branch below (and ditto below).

But we'd need to test that on real hardware to confirm

src/gba/GBA.cpp
```diff
@@ -50,8 +50,9 @@ bool busPrefetchEnable = false;
 uint32_t busPrefetchCount = 0;
 int cpuDmaTicksToUpdate = 0;
 int cpuDmaCount = 0;
-bool cpuDmaHack = false;
+bool cpuDmaRunning = false;
 uint32_t cpuDmaLast = 0;
+uint32_t cpuDmaPC = 0;
```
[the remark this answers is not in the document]
rkitover: This was set by the line Squall asked me to remove, me removing that line essentially disables this whole patch.

The line is [here](https://github.com/visualboyadvance-m/visualboyadvance-m/blob/a9f35645b79356d06a4c7873ccdfebb1a2ef067a/src/gba/GBA.cpp#L2307).

src/gba/GBAinline.h
rkitover: @Squall-Leonhart does this make sense to you?

rkitover:
@Steelskin Can you take over here please, the branch is on this repo so you can write to it.

The game we are trying to not break is called Croket I believe.

@Squall-Leonhart Can you fill him in on the details please.

src/gba/GBAinline.h
Squall-Leonhart: the GBA documentation is not complete, some games take advantage of these undocumented features.

For instance, nogba has the slowdown issue, but not the hang issue that the original version of the downstream libretro patch introduced.

Squall-Leonhart:
Croket 2 actually.

PR #1026, opened 2022-10-16
[config] Add Observers for Options

Steelskin:
This adds a generic Observer interface to config::Option. To demonstrate its intended usage, the Display Configuration dialog and the options it controls have been updated to be entirely handled via the config::Option class.

Implementations for wxValidator are used to validate the flow between the UI and the underlying Option. In turn, modifying the Option value triggers all of its observers that should then do what they need to do.

Rather than explicitly calling all of the needed methods after modifying a global option value, the UI elements that need to be notified when an Option value is modified will be notified via their observers. Runtime assert checks are put in place to prevent infinite recursion if an observer attempts to modify an Option while handling the observer callback.

Once all uses of Options have been updated, we should get into a state where the following will be true:
* cmdevents.cpp will no longer rely on the application state.
* All dialogs will have been moved to specific implementations, reducing the size of guiinit.cpp
* update_opts() will no longer need to be called and will be removed.

This will then make it easier to update accelerator handling to be done with config::UserInput.

Bug: #745

### DIFF
--- a/po/wxvbam/wxvbam.pot
+++ b/po/wxvbam/wxvbam.pot
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-10-21 19:17+0000\n"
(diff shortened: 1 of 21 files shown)


rkitover (approved):
Looks good to merge after rebase.

PR #1030, opened 2022-10-22
Backport 800d6ed69b from wx to fix Wayland EGL pos

rkitover:
Backport 800d6ed69b from wxWidgets git to fix resizing the `GLDrawingPanel` in Wayland using the EGL `wxGLCanvas`.

Also rename `IsItWayland()` to `IsWayland()`.

### DIFF
--- a/src/wx/guiinit.cpp
+++ b/src/wx/guiinit.cpp
@@ -2672,7 +2672,7 @@ bool MainFrame::BindControls()
 
             if (cmdtab[i].cmd_id == XRCID("AllowKeyboardBackgroundInput")
 #if defined(__WXGTK__)
-                && IsItWayland()
+                && IsWayland()
 #endif
                ) {
                 if (mi)
--- a/src/wx/panel.cpp
+++ b/src/wx/panel.cpp
@@ -1302,7 +1302,7 @@ static void process_keyboard_event(const wxKeyEvent& ev, bool down)
     int kc = ev.GetKeyCode();
 
     // Under Wayland or if the key is unicode, we can't use wxGetKeyState().
-    if (!IsItWayland() && kc != WXK_NONE) {
+    if (!IsWayland() && kc != WXK_NONE) {
         // Check if the key state corresponds to the event.
         if (down != wxGetKeyState(static_cast<wxKeyCode>(kc))) {
             return;
@@ -2281,7 +2281,7 @@ void GLDrawingPanel::DrawingPanelInit()
     glClearColor(0.0, 0.0, 0.0, 1.0);
 // non-portable vsync code
 #if defined(__WXGTK__)
-    if (IsItWayland()) {
+    if (IsWayland()) {
 #ifdef HAVE_EGL
         if (vsync)
             wxLogDebug(_("Enabling EGL VSync."));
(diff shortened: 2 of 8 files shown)


src/wx/wayland.h
```diff
@@ -1 +1,15 @@
...
+
+#include <wx/version.h>
+#include <wx/window.h>
+#include <wx/glcanvas.h>
+
+bool IsWayland();
+
+// Temporary hack to backport 800d6ed69b from wxWidgets until 3.2.2 is released.
+#if defined(__WXGTK__) && wxCHECK_VERSION(3, 2, 0) && !wxCHECK_VERSION(3, 2, 2)
+void MoveWaylandSubsurface(wxGLCanvas* win);
```
Steelskin, line 12: I'd suggest leaving the definition in the header and making it a no-op on non-wayland platforms to fix the build issue.

src/wx/wayland.cpp
```diff
@@ -1,11 +1,31 @@
+#define private public
```
Steelskin, line 1: Please, limit this to being behind the `WX_HAS_BROKEN_WAYLAND_OPENGL` check and do `#undef private` after the include. The include could be limited to `#include <wx/unit/glegl.h>`

src/wx/wayland.cpp
```diff
@@ -1,11 +1,31 @@
...
 #ifdef GDK_WINDOWING_WAYLAND
 #include <gdk/gdkwayland.h>
-        bool IsItWayland() { return GDK_IS_WAYLAND_DISPLAY(gdk_display_get_default()); }
-#else
-        bool IsItWayland() { return false; }
+
+bool IsWayland() { return GDK_IS_WAYLAND_DISPLAY(gdk_display_get_default()); }
+
+// Temporary hack to backport 800d6ed69b from wxWidgets until 3.2.2 is released.
+#if !wxCHECK_VERSION(3, 2, 2)
```
Steelskin, line 12: The check should be:
```c++
defined(__WXGTK__) && wxCHECK_VERSION(3, 2, 0) && !wxCHECK_VERSION(3, 2, 2)
```

Though I would suggest doing this instead before any other include:
```c++
#include <wx/version.h> 

#if defined(__WXGTK__)

#if wxCHECK_VERSION(3, 2, 2)
#define WX_HAS_WAYLAND_OPENGL_SUPPORT 1
#define WX_HAS_BROKEN_WAYLAND_OPENGL 0
#elif wxCHECK_VERSION(3,2,0)
#define WX_HAS_WAYLAND_OPENGL_SUPPORT 1
#define WX_HAS_BROKEN_WAYLAND_OPENGL 1
#else  // wxCHECK_VERSION
#define WX_HAS_WAYLAND_OPENGL_SUPPORT 0
#define WX_HAS_BROKEN_WAYLAND_OPENGL 0
#endif  // wxCHECK_VERSION

#else  // defined(__WXGTK__)

#define WX_HAS_WAYLAND_OPENGL_SUPPORT 0
#define WX_HAS_BROKEN_WAYLAND_OPENGL 0

#endif  // defined(__WXGTK__)
```

Then you can simply refer to the macros in the rest of the file, including for the includes.

src/wx/panel.cpp
```diff
@@ -2363,6 +2362,11 @@ void GLDrawingPanel::OnSize(wxSizeEvent& ev)
 {
     AdjustViewport();
 
+// Temporary hack to backport 800d6ed69b from wxWidgets until 3.2.2 is released.
+#if wxCHECK_VERSION(3, 2, 0) && !wxCHECK_VERSION(3, 2, 2)
+    MoveWaylandSubsurface(this);
```
Steelskin, line 2367: You can invoke this in every case and make it a no-op on other platforms (see other comments) - the linker will probably optimize it out, see other comments.

src/wx/guiinit.cpp
```diff
@@ -2672,7 +2672,7 @@ bool MainFrame::BindControls()
 
             if (cmdtab[i].cmd_id == XRCID("AllowKeyboardBackgroundInput")
 #if defined(__WXGTK__)
-                && IsItWayland()
+                && IsWayland()
```
Steelskin, line 2675: We can call this on every platform (see other comment)

rkitover:
@Steelskin Your window options PR made all of this much nicer. Also, I redid this to be nicer too, can you take a look again please?

PR #1049, opened 2022-12-29
Arm64 msvc cross

rkitover:
@Steelskin I'm building with `-DCMAKE_CXX_FLAGS=-DNO_OGL` until I add your wxwidgets port patch.

### DIFF
--- a/cmake/HostCompile.cmake
+++ b/cmake/HostCompile.cmake
@@ -1,23 +1,34 @@
 function(host_compile src dst_cmd)
     if(CMAKE_CROSSCOMPILING)
         unset(link_flags)
+        set(dst "${dst_cmd}")
 
         if(CMAKE_HOST_WIN32)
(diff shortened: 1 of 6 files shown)


cmake/HostCompile.cmake
```diff
@@ -1,23 +1,32 @@
...
-
             if(CMAKE_COMPILER_IS_GNUCXX)
                 set(link_flags -Wl,--subsystem,console)
             endif()
+        endif()
+
+        if(MSVC)
+            add_custom_command(
+                OUTPUT ${dst}
+                DEPENDS ${src}
```
Steelskin, line 15: We also depend on `${CMAKE_SOURCE_DIR}/cmake/MSVC_x86_Host_Compile.cmake` here.

Steelskin (commented):
Thanks for doing most of the heavy lifting!
I have a few comments, mostly to make the output less verbose.

PR #1054, opened 2023-01-04
[WIP] Take whatever you like

Neumann-A:
So i only ran the windows build
- use vcpkg for all deps (manifest mode)
- delete vendored deps
- modernize/simplify CMake
- remove dead CMake stuff?
- removed CI logic from build scripts (these belong into the CI scripts not the CMake script [it can however be a CI])
- update to xaudio 2.9
- require cxx17 because some constexpr stuff (compiler complaint)

So I am probably not going to spend the time making this mergeable so take whatever changes you like.


ZachBacon:
The changes I don't mind as much, but the xaudio2 code I can't agree with just yet. It will fail the CI tests that use mingw-w64. Right now their xaudio2 code mainly targets 2.7. But, on the other hand support for it could most likely be added without much effort in mingw. And the STB headers, I'd like to ask why you removed them? Is it in regards to the vpkg changes? If so, we kinda need those to remain, again for CI purposes and for those that develop on non msvc toolchains.

Neumann-A:
>  I'd like to ask why you removed them?

They are not latest (2.25) and just pulling them in from vcpkg was straightforward (2.27).

Maybe you want to consider switching completely to vcpkg. It should work on your targets leading to a more consistent cmake build. From what I see in your vcpkg usage I would assume you looked into it before manifests/binarycaching were introduced.

ZachBacon:
The main issue with vcpkg on mingw is that it's not a tested target, many compile issues are had last I checked and not a lot of linux distributions have those headers either so it makes packaging for linux headache prone, buuut not impossible, I know at least archlinux carries those headers for sure. Worst case scenario we could pull in from the stb git repo if it's being compiled by mingw. Other than that, there's a lot of good cleanup regardless.

Neumann-A:
> Worst case scenario we could pull in from the stb git repo if it's being compiled by mingw.

Since I upped CMake to 3.11 you can use FetchContent but since stb is header only you could also use vcpkg to pull it in. Nothing to compile so it should always work.

> The main issue with vcpkg on mingw is that it's not a tested target, many compile issues are had last I checked

Yeah MinGW is not an official tested target by ms but there are users using it with MinGW. Since both `sdl2` and `ffmpeg` have `VCPKG_TARGET_IS_MINGW` in their portfile I would assume that at least those will work if your MinGW setup is correct.

ZachBacon:
To a varying degree, I typically just use the binary packages from MSYS2 since it's an active project with binary packages that can be downloaded and updated in that regard and their mingw headers typically follow mingw-w64's git repository, so some advancements in the upstream project trickles down, which is why I said prior eventually it could get xaudio 2.9 support most likely soon enough.

ZachBacon:
At any rate, I'll let the others chime in with their review as well.

Steelskin:
I have cherry-picked many of these changes in #1238 and #1244. There is still more to do to improve the build but we are already in a much better state. Next on my slate is cleaning up more of the global include stuff and narrowing down the scope of dependencies. I also would like to use vcpkg in manifest mode at some point but I need to read more of the documentation for that.

I truly thank you for your contribution, this draft PR has been an invaluable help in getting there.

PR #1051, opened 2022-12-30
Correctly port over DoDMA from the Libretro fork and partially revert Downcasting in bios.cpp

Squall-Leonhart:
This patch ports over DoDMA from the libretro fork, removes uInt Downcasting on the sine table and changes the bios instructions to signed.

Makes #1020 redundant
Fixes #1050

Issue #1050, opened 2022-12-29
Animation error on Pokemon 3 gen [RS/FRLG/E]


Unknown326:
So far I was using VBA-M 2.1.4 x64 bits or its respective nightly version and I didn't find any errors. My operating system is LINUX MINT 21 Vanessa [Cinnamon] x64 bits and I run it with WINE 8.0 - rc2. What caught my attention is the animation of the pokeball that the trainer throws when entering combat since in version 2.1.4 it happens normally.

With the current version 2.1.5, the animation flickers and the pokeball doesn't spin. I tested it on all 3 gen games. Although it does not affect the game, it caught my attention and that is why I report it. In advance, thank you very much for your time and thanks for so much effort :)

Test in GBA 3 gen Pokemon games [RS/FRLG/E]

Settings

Video driver (simple, opengl): opengl
Sound driver (directsound, xaudio, openal, SDL): No
Scale Filter Used?: xBRZ 4x

Specifications and Build details

VBA-M: 2.1.5 x64 bits
OS: Linux Mint Vannesa Cinnamon x64 bits

Squall-Leonhart:
turn off frame skipping.

Issue #1003, opened 2022-09-25
[Build] Build with -Wall -Werror


Steelskin:
## Let's build with -Wall -Werror!

This is a fun project anyone can join in by picking a file and making it build with no warning.

To get started:
* Build visualboyadvance-m
* Pick one file at random in the build output.
* Change it until it builds with no warning
* Send a PR

Please try to limit your PRs to one per file, it will make it easier to review. Some warnings come from headers and will probably require multiple file changes, this is fine.

### Old C function definitions

This is an older and soon to be removed from the C standard way of defining functions in C:
```c
int old_style(a , b) int a;
int b;
{
        // [...]
}
```

This corresponds to the following modern syntax:
```c
int old_style(int a, int b)
{
        // [...]
}
```

### Declaration of `var` hides class member

Rename the class member to have a trailing underscore and make sure to replace all of the occurrences of that class member. If you have clangd configured and a working build, you should be able to find all references and even use clangd refactoring tools. Do double-check every use before submitting. I recommend keeping these to one PR per class member renaming.

### Unreferenced formal parameter

C++ allows to define an argument without giving it a name. Since there can be a good reason to keep the parameter name (e.g. an override or an event handler), I recommend the following syntax:

```c++
int function(int /* unused_parameter */, int other_parameter)
```

### Reason box

N/A

### Examples

N/A

PR #1060, opened 2023-01-14
Remove all warnings in gba/bios.cpp

Steelskin:
A regression from #1008 was identified as the root cause for #1050 and fixed in #1051. This is a follow-up that changes the constants used for `sineTable` to be unsigned hex, which fixes warnings in MSVC without breaking changes.

Bug: #1003


Steelskin:
@Squall-Leonhart Can you double-check this doesn't break anything this time around? I tested it locally but I'd like a second pair of eyes.

PR #1062, opened 2023-01-20
Rename appdata to metainfo

Polynomial-C:
See https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html chapter 2.1.2 "Filesystem locations"

Signed-off-by: Lars Wendler <polynomial-c@gmx.de>


rkitover:
Thank you!

PR #1065, opened 2023-01-31
Add stdint includes

Mystro256:
I believe this is needed on newer GCC's

I noticed this when trying to compile on latest Fedora.

### DIFF
--- a/src/wx/config/option.h
+++ b/src/wx/config/option.h
@@ -7,6 +7,7 @@
 #include <functional>
 #include <unordered_set>
 
+#include <stdint.h>
 #include <wx/string.h>
 
 #include "config/option-id.h"
--- a/src/wx/widgets/wx/wxmisc.h
+++ b/src/wx/widgets/wx/wxmisc.h
@@ -2,6 +2,7 @@
 #define WX_MISC_H
 // utility widgets
 
+#include <stdint.h>
 #include <wx/checkbox.h>
 #include <wx/valgen.h>
 


rkitover:
The correct way to do this in C++ is:

```cpp
#include <cstdint>
```

, can you check if this works for you?

rkitover:
Do you mind also adding a github action for building on latest fedora? It should be similar to the ubuntu one.

Clownacy:
Apparently the `<cstdxxx>` headers [are not guaranteed](https://stackoverflow.com/questions/13642827/cstdint-vs-stdint-h) to place any of their definitions in the global namespace, while the `<stdxxx.h>` ones are. Since there are [lines like this](https://github.com/visualboyadvance-m/visualboyadvance-m/blob/425e7a0aa0e6c46b2986d9017566fa0ef3b04681/src/wx/config/option.h#L166) that use `uint32_t` and not `std::uint32_t`, `<stdint.h>` is the correct header to use.

rkitover:
> Since there are [lines like this](https://github.com/visualboyadvance-m/visualboyadvance-m/blob/425e7a0aa0e6c46b2986d9017566fa0ef3b04681/src/wx/config/option.h#L166) that use uint32_t and not std::uint32_t, <stdint.h> is the correct header to use.

I'd rather fix the variable declarations then, if that's the case. Can you tell me how to reproduce the build failure?

rkitover:
I reproduced the error on rawhide, I'll see what the best way to fix this is.

PR #1070, opened 2023-02-05
[Option] Add Proxy template class for Option

Steelskin:
This removes the global accessors for individual options in favor of a proxy template class with a simpler per-type API. This adds compile-time checks to validate that the right option is used with the right type when accessing it directly.


Steelskin:
I know I've been iterating a lot on the Option class but I think I know where I want to end up now. This should make it much easier to update the values down the line and we get compile-time type checks to boot!

src/wx/cmdevents.cpp
```diff
@@ -2558,32 +2559,32 @@ EVT_HANDLER(GameBoyConfigure, "Game Boy options...")
 
 EVT_HANDLER(SetSize1x, "1x")
 {
-    config::OptDispScale()->SetDouble(1);
+    config::Proxy<config::OptionID::kDispScale>().Set(1);
```
rkitover, line 2562: This is great, but I'd like some much simpler accessors for the main GUI code, this is a bit too long just to set an option.

rkitover (changes requested):

src/wx/cmdevents.cpp
Steelskin: I added a macro and you can now treat the object more like a variable. The setter operator overloads still return `bool` rather than the object itself. Let me know if you'd like to change that so they're more variable-like. The `OptionProxy` object is still accessible directly for these needs.

PR #1081, opened 2023-02-15
Fix automatic dynamic frame skipping

Steelskin:
The previously used algorithm was too aggressive. This fixes the issue by using the last second data and using a more gentle adjustment curve to prevent huge changes in frame skipping.

### DIFF
--- a/src/System.h
+++ b/src/System.h
@@ -76,7 +76,7 @@ extern void systemPossibleCartridgeRumble(bool);
 extern void updateRumbleFrame();
 extern bool systemCanChangeSoundQuality();
 extern void systemShowSpeed(int);
-extern void system10Frames(int);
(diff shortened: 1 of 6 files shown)


Felipefpl:
I was having a weird situation playing golden sun with the hq2x filter where the FPS would go above the limit of 60, does this fix it?

Steelskin:
> I was having a weird situation playing golden sun with the hq2x filter where the FPS would go above the limit of 60, does this fix it?

I doubt it. The only way to activate this code is by manually setting the frameskip value to -1 in the INI file.

Felipefpl:
I see, thanks for the info.

Squall-Leonhart:
there is an auto frameskip option in the menus that isn't hooked up i guess.

PR #1098, opened 2023-03-17
optional using xss to suspend screensaver

multiSnow:
suspend screensaver when game is started/resumed.
restart screensaver when game is unloaded/paused.
(screensaver will be restarted automatically if the whole program disconnected from the X server)

use ENABLE_XSS cmake option to enable xss.
by default, ENABLE_XSS is OFF thus nothing changed (using `XResetScreenSaver` to reset screensaver everytime the joystick botton pressed).


rkitover:
Hi, sorry I missed this, our server was down last week.

Please join our discord if you'd like to discuss this there.

I would prefer this to be a runtime option, like a checkbox in display setting for "suspend screensaver", we'd do any necessary adjustments for other platforms as well.

multiSnow:
The checkbox is added in 'UI Settings...'.

CMakeLists.txt
```diff
@@ -103,6 +103,9 @@ set(EXECUTABLE_OUTPUT_PATH ${PROJECT_BINARY_DIR})
 
 option(ENABLE_SDL "Build the SDL port" OFF)
 option(ENABLE_WX "Build the wxWidgets port" ON)
+if(ENABLE_WX)
+    option(ENABLE_XSS "Check XSS support in the wxWidgets port" OFF)
```
rkitover, line 107: I don't think this is necessary, we can just automatically use it when available.

rkitover (changes requested):
This looks fantastic, thank you, I am requesting one minor change and then I will merge it.

src/wx/guiinit.cpp
```diff
@@ -3188,6 +3188,7 @@ bool MainFrame::BindControls()
 
         d = LoadXRCDialog("UIConfig");
         { getcbb("HideMenuBar", gopts.hide_menu_bar); }
+        { getcbb("SuspendScreenSaver", gopts.suspend_screensaver); }
```
rkitover, line 3191: We should hide the checkbox on Windows and mac until we have an implementation for those platforms. But you don't have to do this, I can do this after I merge your change.

rkitover:
This is merged, thank you for your contribution!

I made one minor change to also hide the checkbox under Wayland.

PR #1103, opened 2023-03-27
Move GB/GBC header parsing to  gbCartData

Steelskin:
This adds a new class, gbCartData to do all of the GB/GBC header parsing in a single location.

Breaking change: ROMs advertising a 2 KiB RAM size now properly only have 2 KiB allocated for RAM, whereas it used to be 8 KiB. This was tested with the one aftermarket ROM making use of this RAM size (Quartet), with no issue. Save files and save state files should still be compatible since the RAM was saved with the correct 2 KiB size.


Squall-Leonhart:
Looks good to me.

Squall-Leonhart (approved):

PR #1106, opened 2023-04-03
Refactor the save/load state and battery code

Steelskin:
* Simplifies most of the GB save/load state code and changes many of the global variable uses to go through gbCartData instead.
* Changes all battery reads to use gzFiles. This seems to have been an oversight in an earlier refactor of the code.
* Removes some unused variables and stops exporting internal-only functions in GB.cpp/gb.h.

Breaking change: Cartridges with no mapper, but a save battery should now properly save and load changes. However, this type of cartridge has not been used in any commercially released software for Game Boy so the exact intended behavior is speculative.

### DIFF
--- a/src/gb/GB.cpp
+++ b/src/gb/GB.cpp
@@ -6,6 +6,7 @@
 #include <cstdio>
 #include <cstdlib>
 #include <cstring>
+#include <vector>
 
 #include "../NLS.h"
 #include "../System.h"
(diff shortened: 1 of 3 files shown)


n-a-c-h:
Overall these changes look great. You eliminated a bunch of magic numbers and got rid of a lot of unnecessary overhead, vastly simplifying the code.

However, that MBC7 change looks problematic. One of the MBC7 games uses 512 bytes of RAM, not 256.

Also, looking over this code, I think we can do a better job handling TAMA5, as what's happening right now for saving/loading is inconsistent with how times are handled internally. I also question why we're saving time both as a unix timestamp and as broken down components, which is redundant. We should use just one or the other. Although that's a pre-existing issue.

n-a-c-h (approved):
We need to do something about 512 byte RAM MBC7 support. Other than that, this change looks good.

Steelskin:
> We need to do something about 512 byte RAM MBC7 support. Other than that, this change looks good.

I changed the reported size to 512 bytes, but left saving and loading to 256 bytes for compatibility reasons. I left TODOs to deal with this properly down the line.

PR #1111, opened 2023-04-10
Refactor the network link dialogs

Steelskin:
* Unifies the network link options into a single dialog by removing the separate dialog for the timeout option.
* Cleans up the logic related to starting the network link.
* Moves more options to g_owned_opts.
* Separates the server and client port into 2 separate options.

Test: Locally with Mario Kart with 2 instances.


src/wx/guiinit.cpp
```diff
@@ -2528,7 +2412,7 @@ bool MainFrame::BindControls()
 #endif
 #ifdef NO_LINK
 
-            if (cmdtab[i].cmd_id == XRCID("LanLink") || cmdtab[i].cmd_id == XRCID("LinkType0Nothing") || cmdtab[i].cmd_id == XRCID("LinkType1Cable") || cmdtab[i].cmd_id == XRCID("LinkType2Wireless") || cmdtab[i].cmd_id == XRCID("LinkType3GameCube") || cmdtab[i].cmd_id == XRCID("LinkType4Gameboy") || cmdtab[i].cmd_id == XRCID("LinkAuto") || cmdtab[i].cmd_id == XRCID("SpeedOn") || cmdtab[i].cmd_id == XRCID("LinkProto") || cmdtab[i].cmd_id == XRCID("LinkConfigure")) {
+            if (cmdtab[i].cmd_id == XRCID("LanLink") || cmdtab[i].cmd_id == XRCID("LinkType0Nothing") || cmdtab[i].cmd_id == XRCID("LinkType1Cable") || cmdtab[i].cmd_id == XRCID("LinkType2Wireless") || cmdtab[i].cmd_id == XRCID("LinkType3GameCube") || cmdtab[i].cmd_id == XRCID("LinkType4Gameboy") || cmdtab[i].cmd_id == XRCID("LinkAuto") || cmdtab[i].cmd_id == XRCID("SpeedOn") || cmdtab[i].cmd_id == XRCID("LinkProto")) {
```
rkitover, line 2415: Could we do something like apply `XRCID()` to the list of strings into a temporary set and check that it contains the item?

rkitover (changes requested):
I think it would be nicer if the client/server panels were shown by themselves on radio click with the other hidden. `Hide()` possibly followed by `Layout()` should do it.

Nice work on the option validator.

Issue #1170, opened 2023-08-15
[Libretro] [Bug]: Compiled builds of VBA-M for Wii/Wii U (libretro) show only a white screen on all emulated games.


saulfabregwiivc:
### Expected behavior?

I expected in the compiled builds for Wii / Wii U (Libretro/RetroArch) to display the screens of the emulated games (title screens, intros, actual gameplay, etc.)

### Actual behavior?

On all tested games, it just display a white screen. (Video driver issues on VBA-M for these consoles? Wii / Wii U?)

GB/GBC:

![image](https://github.com/visualboyadvance-m/visualboyadvance-m/assets/66485640/7e0592b7-bc3f-4b2f-a2be-253e31be6ef7)

GBA:

![image](https://github.com/visualboyadvance-m/visualboyadvance-m/assets/66485640/edc60c9d-4f73-4f97-8028-03d7b5c45e9e)

### Steps to reproduce the problem

1. Compile a build of VBA-M as a Libretro core for Wii or Wii U as usual (enter to **VBA-M_source/src/libretro** folder and run **make platform=wii** or **make platform=wiiu**)
2. Run the compiled build from RetroArch on Wii or Wii U (depending of the system you have built the core)
3. Load any game (GB, SGB, GBC, GBA)
4. The game will only display a white screen

### Error messages?

None. Just a white screen, no error messages.

### What game were you playing

All the tested games have the same issue.

### What settings are you using?

Video driver (simple, opengl): Don't know cause i was trying on Wii (gx)
Sound driver (directsound, xaudio, openal, SDL): Don't know cause i was trying on Wii (gx)
Scale Filter Used?: none

### Specifications and Build details

VBA-M: latest VBA-M until git https://github.com/visualboyadvance-m/visualboyadvance-m/commit/558317555970b5fd6805b8d5e14ea259c9a16360
OS: Wii / Wii U (Libretro / RetroArch)

### Copyright Compliance

- [x] There are no infringing files attached to this bug report.

Steelskin:
My guess is that we have some issue with big endian architectures. The libretro frontend is also not very well tested in-tree.

saulfabregwiivc:
> My guess is that we have some issue with big endian architectures. The libretro frontend is also not very well tested in-tree.

[...]

PR #1171, opened 2023-08-18
Libretro: Fix GameCube/Wii/Wii U builds. (Ploggy)

saulfabregwiivc:
Since GameCube, Wii and Wii U are all big-endian systems, we need to call in the Makefile for Libretro cores that these must use big endian code for be able to work, otherwise the emulation will just give a white screen.
Fix by Ploggy, many thanks to him.

Fixes https://github.com/visualboyadvance-m/visualboyadvance-m/issues/1170.


rkitover:
Thank you!

PR #1178, opened 2023-09-05
Porting to the upcoming SDL3 release

ZachBacon:
SDL3 is still a bit of a ways away. But I figured I'd get a head start since they have changed a lot of things
Been following their migration guide (link below)
https://github.com/libsdl-org/SDL/blob/main/docs/README-migration.md

What's being changed isn't the final form, even after things work since SDL3 is still a moving target and since I'm sure there will be improvements made. Also until SDL3 is released, this will not pass any CI stuff.


rkitover:
test

PR #1179, opened 2023-09-06
Disable mirroring for ROMs > 32MB

### LINKED ISSUE #1046: Cannot play GBA ROM (games) greater than 32 MB
### Expected behavior?

All types of GBA games should work fine

### Actual behavior?

As compare to other emulators, even as compare from VBA 1.8.0 beta 3, it doesn't load ROMS greater then 32 MB.

1. Visual boy advance does not play GBA ROMS greater than 32 MB.

### Error messages?

No issues on original ROM, but I am facing this issue on some specific games (actually modified ROM) which is greater than 32 MB, big name is "Pokemon Adventure Red Chapter beta 15 + Expansion Fix C" which is not loading only in VBA-M, but I am not facing this issue on another emulators such as mGBA, and even VBA v1.8.0 beta 3.

Screenshot (2)
Screenshot (1)

Video driver (simple, opengl): Intel HD Graphics 530
Sound driver (directsound, xaudio, openal, SDL): Realtek HD Audio
Scale Filter Used?:

VBA-M: 2.1.5
OS: (include version and OS build number using WinVer if on Windows): Windows 11 and Windows 10 (64 Bit)
If using Linux, specify if using xorg or Wayland:
CPU: Intel Core i5- 6500T
GPU: Intel HD Graphics 530
RAM: 8GB

### Copyright Compliance

rkitover:
In doMirroring() do nothing if the ROM size is greater than 32MB. This fixes these ROMs automatically working without disabling mirroring in settings.

Fix #1046

### DIFF
--- a/src/common/sizes.h
+++ b/src/common/sizes.h
@@ -20,6 +20,8 @@ constexpr size_t k1MiB = 2 * k512KiB;
 constexpr size_t k2MiB = 2 * k1MiB;
 constexpr size_t k4MiB = 2 * k2MiB;
 constexpr size_t k8MiB = 2 * k4MiB;
+constexpr size_t k16MiB = 2 * k8MiB;
+constexpr size_t k32MiB = 2 * k16MiB;
 
 // Game Boy BIOS Sizes.
 constexpr size_t kGBBiosSize = k256B;
--- a/src/gba/GBA.cpp
+++ b/src/gba/GBA.cpp
@@ -12,6 +12,7 @@
 #include "../System.h"
 #include "../Util.h"
 #include "../common/Port.h"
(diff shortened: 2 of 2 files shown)


src/gba/Globals.h
```diff
@@ -118,4 +118,7 @@ extern uint16_t IE;
 extern uint16_t IF;
 extern uint16_t IME;
 
+// Let's try for fewer magic numbers, define them here.
+#define SIZE_32MB 0x2000000
```
Steelskin, line 122: Could you add it to `sizes.h` instead?

Steelskin (approved):

PR #1210, opened 2023-12-24
faudio.cpp: Begin rework of faudio implementation

ZachBacon:
This begins the rework of the faudio code begin with removing a good deal of the imme code that the driver used to poll audio interfaces

Removed the use of HRESULT code since that's more windowisms.

Also adapted more of the faudio stuff like WaveFormatEx is FAudioWaveFormatEx WAVE_FORMAT_PCM is FAUDIO_FORMAT_PCM.

Also changed S_OK into R_OK but I'm pretty sure that isn't the way to go. That said and done, this is still a work in progress, but it compiles. But it doesn't initilize yet, I might need an extra set of eyes for that.

### DIFF
--- a/src/wx/faudio.cpp
+++ b/src/wx/faudio.cpp
@@ -8,42 +8,44 @@
 #include <vector>
 
 // FAudio
-#include <faudio.h>
+#include <FAudio.h>
 
-// MMDevice API
 #include <mmdeviceapi.h>
 
 #include <wx/arrstr.h>
+#include <wx/log.h>
+#include <wx/translation.h>
 
 #include "core/base/sound_driver.h"
 #include "core/base/system.h"
 #include "core/gba/gbaGlobals.h"
 #include "wx/config/option-proxy.h"
 
+namespace {
+
 int GetFADevices(FAudio* fa, wxArrayString* names, wxArrayString* ids,
     const wxString* match)
 {
-    HRESULT hr;
-    UINT32 dev_count = 0;
-
+    uint32_t hr;
+    uint32_t dev_count = 0;
     hr = FAudio_GetDeviceCount(fa, &dev_count);
 
-    if (hr != S_OK) {
+    if (hr != 0) {
         wxLogError(_("FAudio: Enumerating devices failed!"));
-        return true;
+        return -1;
     } else {
         FAudioDeviceDetails dd;
 
-        for (UINT32 i = 0; i < dev_count; i++) {
+        for (uint32_t i = 0; i < dev_count; i++) {
             hr = FAudio_GetDeviceDetails(fa, i, &dd);
 
-            if (hr != S_OK) {
+            if (hr != 0) {
                 continue;
             } else {
                 if (ids) {
-                    ids->push_back((wchar_t*) dd.DeviceID); //FAudio is an interesting beast, but not that hard to adapt... once you get used to it, XAudio2 wouldn't need this, but FAudio declares FAudioDeviceDetails as int32_t
+                    ids->push_back((wchar_t*) dd.DeviceID);
                     names->push_back((wchar_t*) dd.DisplayName);
-                } else if (*match == dd.DeviceID)
+                } else if (*match == wxString((wchar_t*) dd.DeviceID))
                     return i;
             }
         }
@@ -52,29 +54,7 @@ int GetFADevices(FAudio* fa, wxArrayString* names, wxArrayString* ids,
     return -1;
 }
 
-bool GetFADevices(wxArrayString& names, wxArrayString& ids)
-{
-    HRESULT hr;
-    FAudio* fa = NULL;
-    UINT32 flags = 0;
-#ifdef _DEBUG
-    flags = FAUDIO_DEBUG_ENGINE;
-#endif
-
-    hr = FAudioCreate(&fa, flags, FAUDIO_DEFAULT_PROCESSOR); //Apparently this needs 3 parameters, the processor.
-
-    if (hr != S_OK) {
-        wxLogError(_("The FAudio interface failed to initialize!"));
-        return false;
-    }
-
-    GetFADevices(fa, &names, &ids, NULL);
-    //fa->Release();
-    FAudio_Release(fa);
-    return true;
-}
-
-static int FAGetDev(FAudio* fa)
+int FAGetDev(FAudio* fa)
 {
     const wxString& audio_device = OPTION(kSoundAudioDevice);
     if (audio_device.empty())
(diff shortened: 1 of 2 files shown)


src/wx/faudio.cpp
```diff
@@ -194,43 +190,44 @@ class FAudio_Device_Notifier : public IMMNotificationClient {
 // Synchronization Event
 class FAudio_BufferNotify : public FAudioVoiceCallback {
 public:
-    HANDLE hBufferEndEvent;
+    void *hBufferEndEvent;
 
     FAudio_BufferNotify()
     {
         hBufferEndEvent = NULL;
         hBufferEndEvent = CreateEvent(NULL, FALSE, FALSE, NULL);
```
Steelskin, line 198: Double initialization is not required here. Instead, use the initializer list:
```c++
FAudio_BufferNotify() : buffer_end_event_(CreateEvent(nullptr, FALSE, FALSE, nullptr) {
    assert(buffer_end_event_);
    // rest of initialization
}
```

src/wx/faudio.cpp
```diff
@@ -339,15 +335,15 @@ bool FAudio_Output::init(long sampleRate)
     if (failed || initialized)
         return false;
```
Steelskin, line 336: Would it make sense to assert here instead?

src/wx/faudio.cpp
```diff
@@ -359,38 +355,36 @@ bool FAudio_Output::init(long sampleRate)
     soundBufferLen = (freq / 60) * 4;
     // create own buffers to store sound data because it must not be
     // manipulated while the voice plays from it
-    buffers = (BYTE*)malloc((bufferCount + 1) * soundBufferLen);
+    buffers = (uint8_t*)malloc((bufferCount + 1) * soundBufferLen);
```
Steelskin, line 358: Would it make sense to switch `buffers` to be a `std::vector<yunt8_t>` instead?
You can use `resize(bufferCount + 1)` to resize it here, which will also zero out the memory at the same time. You can access the underlying C-array if you need to via the `data()` API.

src/wx/faudio.cpp
```diff
@@ -359,38 +355,36 @@ bool FAudio_Output::init(long sampleRate)
...
         wxLogError(_("FAudio: Creating mastering voice failed!"));
         failed = true;
         return false;
     }
 
     // create sound emitter
     //This should be  FAudio_CreateSourceVoice()
-    //hr = faud->CreateSourceVoice(&sVoice, &wfx, 0, 4.0f, &notify);
-     hr = FAudio_CreateSourceVoice(faud, &sVoice, (const FAudioWaveFormatEx*)&wfx, 0, 4.0f, &notify, NULL, NULL);
+    hr = FAudio_CreateSourceVoice(faud, &sVoice, &wfx, 0, 4.0f, &notify, NULL, NULL);
```
Steelskin, line 385: Could you document the magic values `0` and `4.0f`? It's fine to leave them as-is if you don't know why they're here.

src/wx/faudio.cpp
```diff
@@ -399,9 +393,8 @@ bool FAudio_Output::init(long sampleRate)
     if (gopts.upmix) {
         // set up stereo upmixing
         FAudioDeviceDetails dd;
-        ZeroMemory(&dd, sizeof(dd));
-        hr = FAudio_GetDeviceDetails(faud, 0, &dd);
-        assert(hr == S_OK);
+        memset(&dd, NULL, sizeof(dd));
```
Steelskin, line 396: In C++, you can just value-initialize the struct to zero it out like this:
`FAudioDeviceDetails dd {};`

src/wx/faudio.cpp
```diff
@@ -399,9 +393,8 @@ bool FAudio_Output::init(long sampleRate)
...
+        assert(FAudio_GetDeviceDetails(faud, 0, &dd) == 0);
         float* matrix = NULL;
```
Steelskin, line 398: Same as above, you can declare `matrix` this way: `std::vector<float> matrix(2 * dd.OutputFormat.Format.nChannels);`
This way you don't have to deal with memory manually.

src/wx/faudio.cpp
ZachBacon: That I wouldn't know, that's something that is even in the original xaudio source file which this is largely based off of.

src/wx/faudio.cpp
ZachBacon: I really don't know, I think this has been the default for even long before I joined the project. The XAudio2 implementation was largely based off of the work on what mudlord did for the original MFC interface https://github.com/visualboyadvance-m/visualboyadvance-m/blob/mfc/src/win32/XAudio2.cpp

src/wx/faudio.cpp
```diff
@@ -194,43 +190,44 @@ class FAudio_Device_Notifier : public IMMNotificationClient {
 // Synchronization Event
 class FAudio_BufferNotify : public FAudioVoiceCallback {
 public:
-    HANDLE hBufferEndEvent;
+    void *hBufferEndEvent;
```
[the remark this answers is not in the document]
ZachBacon: Wouldn't that run afoul for it's use later, like if (WaitForSingleObject(notify.hBufferEndEvent, 10000) == WAIT_TIMEOUT), wouldn't it complain about it being declared private? I guess I'm just not really understanding the thought process behind the suggestion.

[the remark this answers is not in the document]
Steelskin: Private members of a class are only accessible by the methods of said class. Since the `WaitForSingleObject()` call is done within the class itself, it will work just fine. At the end of the day, this is just data that is passed to the `WaitForSingleObject()` call. The point of private members is to prevent non-member functions from accidentally accessing them.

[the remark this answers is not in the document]
Steelskin: Oh, nevermind, I just re-read the code and realized it's used by a different object.
I would suggest turning the `WaitForSingleObject()` call into an utilitly method in `FAudio_BufferNotify`, like this:
```c++
// Waits for this object to be notified for 10 seconds. Returns true if this object was signaled.
bool WaitForSignal() {
    return WaitForSingleObject(hBufferEvent, 10000) != WAIT_TIMEOUT;
}
```

It will also make it easier to port to non-Windows down the line.

src/wx/faudio.cpp
```diff
@@ -194,43 +190,44 @@ class FAudio_Device_Notifier : public IMMNotificationClient {
...
         assert(hBufferEndEvent != NULL);
-    }
 
+        OnBufferEnd = &FAudio_BufferNotify::StaticOnBufferEnd;
+        OnVoiceProcessingPassStart = &FAudio_BufferNotify::StaticOnVoiceProcessingPassStart;
+        OnVoiceProcessingPassEnd = &FAudio_BufferNotify::StaticOnVoiceProcessingPassEnd;
+        OnStreamEnd = &FAudio_BufferNotify::StaticOnStreamEnd;
+        OnBufferStart = &FAudio_BufferNotify::StaticOnBufferStart;
+        OnLoopEnd = &FAudio_BufferNotify::StaticOnLoopEnd;
+        OnVoiceError = &FAudio_BufferNotify::StaticOnVoiceError;
```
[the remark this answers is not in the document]
ZachBacon: Since FAudioVoiceCallback is a struct in FAudio.h

couldn't I do something like

```
struct FAudio_BufferNotify {
  void OnBufferEnd(void* pBufferContext) {}
  FAudioVoiceCallback cb = {
   /* OnBufferEnd */ [](void* thiz, void* pBufferContext) { reinterpret_cast<FAudio_BufferNotify *>(thiz)->OnBufferEnd(pBufferContext); },
  // ... and the rest ...
  };
};
```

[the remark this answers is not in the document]
Steelskin: Making the `FAudio_BufferNotify` the `FAudioVoiceCallback` object is a common pattern that allows you to more easily manage lifetimes. Basically, the idea is that the callback object always have to outlive the caller object. By making the callback object owned by the caller object, you have a strong guarantee from C++ semantics that the callback object is always going to outlive the caller object.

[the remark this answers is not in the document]
Steelskin: Couldn't figure out how to do it differently, that will have to do for now.

src/wx/faudio.cpp
Steelskin: Ack.

src/wx/faudio.cpp
Steelskin: I'll follow up separately on these.

src/wx/faudio.cpp
Steelskin: Ack.

src/wx/faudio.cpp
Steelskin: As above, I'll follow up separately.

Steelskin (approved):

PR #1216, opened 2024-01-24
ffmpeg: Use `av_write_trailer()` on stop recording

### LINKED ISSUE #1214: [Bug]: WAV audio-only recording is broken
### Expected behavior?

On using `File->Record->Start sound recording...` and selecting the WAV format, a WAV file is created to capture the audio.

### Actual behavior?

Everything works as expected, but the RIFF and data chunk headers are broken (the sizes are set to FFFFFFFFh). I believe this is caused by [this line](https://github.com/visualboyadvance-m/visualboyadvance-m/blob/f915ec5972eb517f0972570bd1f017b01083c1a4/src/common/ffmpeg.cpp#L417), which only calls `av_write_trailer()` when _not_ recording in audio-only mode.

Inspecting the ffmpeg source, the call to `av_write_trailer()` should invoke `wav_write_trailer()` in libavformat/wavenc.c, which then writes the correct chunk sizes, as they'd been set to FFFFFFFFh when `avformat_write_header()` invoked `wav_write_header()`. So in other words: the call to `av_write_trailer()` should _always_ be performed, not only when doing a video recording.

1. Start audio recording with WAV output.
2. Stop recording.
3. Inspect the resulting file.

### Error messages?

_No response_

### What game were you playing

_No response_

### What settings are you using?

Default settings, but N/A; this is an ffmpeg interfacing issue.

VBA-M 2.1.8 (64-bit). Should be present in all current builds, though. Not sure how far back this issue goes, since it seems to be related to ffmpeg's API changing through the years.

### Copyright Compliance

Aikku93:
This fixes #1214. I also had to make some modifications so that `av_write_trailer()` isn't called during a failure in initialization, as that would cause the program to crash from the potentially-invalid encoder state.

### DIFF
--- a/src/common/ffmpeg.cpp
+++ b/src/common/ffmpeg.cpp
@@ -339,34 +339,34 @@ recording::MediaRet recording::MediaRecorder::Record(const char *fname, int widt
     ret = setup_common(fname);
     if (ret != MRET_OK)
     {
-        Stop();
+        Stop(false);
         return ret;
     }
     // video stream
     ret = setup_video_stream_info(width, height, depth);
     if (ret != MRET_OK)
     {
-        Stop();
+        Stop(false);
         return ret;
     }
     ret = setup_video_stream(width, height);
     if (ret != MRET_OK)
     {
-        Stop();
+        Stop(false);
         return ret;
     }
     // audio stream
     ret = setup_audio_stream();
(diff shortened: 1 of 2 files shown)


rkitover:
Merged, thank you very much!

PR #1224, opened 2024-02-19
build: fix gentoo dependency namespaces

HWXLR8:
Change sys-devel to dev-build to match current gentoo package naming convention. Also add --ask flag to prompt for confirmation before merging packages.


rkitover:
Thank you very much!

PR #1230, opened 2024-03-05
Make menu more reasonably organized

wuweiran:
- The current menu experience is kind of weird -- there's a UI Settings menu item that pops up a dialog and in the dialog there's only one checkbox option. It is more proper looking to have the only option become a checkable menu item. 
- Furthermore, I find "status bar" is reasonable to be part of UI Setting not part of Video, because I think video is about the canvas while UI Setting is about other part of the window (menu, button, etc.). 
- Upon testing, I also find "enable status bar" and "on-screen status messages" is coupled in a confusing way. It's good to make them perpendicular.


ZachBacon:
This will be probably on hold. One of our developers are already working on some UI stuff that this might interfere with.

src/wx/config/option-id.h
```diff
@@ -112,6 +111,7 @@ enum class OptionID {
     kGeomWindowY,
 
     /// UI
+    kUIEnableStatusBar,
```
Steelskin, line 114: While I agree that this is a better place for it, this has the unfortunate side effect of breaking the setting for existing users.
There is no currently no mechanism to rename an existing INI option. It is something I would like to do down the line and do a (hopefully one-time) rename of options to have more consistent naming. In the meantime, I'm going to ask you leave this to `kGenStatusBar` so we don't break existing users.

Steelskin (commented):
Thanks for the PR!
I just have one comment, lgtm otherwise.

src/wx/config/option-id.h
wuweiran: Thanks for the comment, Steelskin. I'll change `kUIEnableStatusBar` and `EnableStatusBar` back to `kGenStatusBar` and `StatusBar`.

wuweiran:
@Steelskin , name and order changes has been reverted. Please take another look at it, thanks.

Steelskin (approved):

PR #1244, opened 2024-03-12
[Build] Add toolchain-specific files

Steelskin:
This moves most toolchain setup to their own specific filies and removes a couple of unused cmake functions.


CMakeLists.txt
```diff
@@ -56,7 +59,7 @@ set(CMAKE_CXX_STANDARD 17)
 
 project(VBA-M C CXX)
 
-find_package(Git)
+find_package(Git REQUIRED)
```
rkitover, line 62: If git is not actually required for linux tarball builds, then I wouldn't make it be, but I will check this myself.

cmake/Toolchain-gcc-clang.cmake
```diff
@@ -0,0 +1,216 @@
...
+    set(MY_C_FLAGS -I/usr/local/include ${MY_C_FLAGS})
+endif()
+
+foreach(C_COMPILE_FLAG ${MY_C_FLAGS})
+    add_compile_options(${C_COMPILE_FLAG})
+endforeach()
+
+include(CheckCXXCompilerFlag)
+
+check_cxx_compiler_flag(-std=gnu++17 GNUPP17_FLAG)
```
rkitover, line 170: I'm not sure if we should do this anymore because we're now setting the C++ standard with cmake.

rkitover (changes requested):
Everything looks good.

CMakeLists.txt
Steelskin: We depend on git for a lot of things and fail in various places if we don't. I think it's better to fail explicitly early in the build process.

cmake/Toolchain-msvc.cmake
```diff
@@ -0,0 +1,54 @@
...
+
+if(CMAKE_CXX_COMPILER_ID STREQUAL MSVC AND NOT CMAKE_GENERATOR MATCHES "Ninja")
+    string(APPEND ADDITIONAL_FLAGS " /MP")
+endif()
+
+set(CMAKE_MSVC_RUNTIME_LIBRARY "MultiThreaded$<$<CONFIG:DEBUG>:Debug>" CACHE INTERNAL "")
+
+set(PREPROCESSOR_FLAGS "/D_FORCENAMELESSUNION /DWIN32_LEAN_AND_MEAN /DWIN32 /D_WINDOWS /D__STDC_LIMIT_MACROS /D__STDC_CONSTANT_MACROS /D_CRT_SECURE_NO_WARNINGS")
+
+set(CMAKE_CXX_FLAGS "/nologo ${PREPROCESSOR_FLAGS} /GR /EHsc /W4 /std:c++17 ${ADDITIONAL_FLAGS}" CACHE STRING "" FORCE)
```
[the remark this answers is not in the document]
Steelskin: Ditto here.

cmake/Toolchain-gcc-clang.cmake
Steelskin: I removed it. In fact, I rewrote the entire file.

CMakeLists.txt
rkitover: We have to support building a tarball without git like any proper linux software. So if git is being used somewhere in that process, it needs to skip that part for that particular use-case. We don't have to support not having git for e.g. `-DUPSTREAM_RELEASE=TRUE` because that's our stuff and not for people to use.

Steelskin: I reverted this and cleaned up some more stuff.

PR #1275, opened 2024-04-06
[Audio] Remove manual memory allocations

Steelskin:
* Remove explicit calls to new or malloc in favor of owned objects.
* Move AudioSdl to the sdl frontend, it is no longer used by the wx frontend.


rkitover (approved):
Looks good to me!

ZachBacon (approved):
Definitely seeing some new stuff that is definitely making me read up on. But overall, LGTM

PR #1306, opened 2024-05-10
[Test] Change assert to custom CHECK macros

Steelskin:
This changes the way we handle asserts to use a set of custom macros. This greatly speeds up crashing, especially on Windows, and provides assertions in release mode too.


Steelskin:
@rkitover I'd like your opinion on this change. The biggest change here is that the checks are now going to be done on release builds too. This will add a bit of binary size because of the strings added but the effect seems minimal from some very casual local testing. The runtime effect is negligible, these are very cheap checks.

PR #1314, opened 2024-06-22
Add option to mute sound during speedup

### LINKED ISSUE #865: [Feature Request] Option to mute sound when turbo is active
Would be awesome if you could toggle an option that would mute the game when you are fast forwarding. The sound gets annoying after a while and it's a cumbersome to hit two buttons to fast forward and mute since I toggle turbo on and off often. This can be worked around if we were able to set two commands to one hotkey, but currently it just overwrites the key with the new command. really awesome and smooth Emulator, just wish it had the old VBA feature to mute when in turbo.

rkitover:
Add core option to mute sound during speedup, implemented in the gba core. Link the option to the SDL port as well.

Add the new Mute Sound option as a checkbox to the xrc for the Speedup/Turbo dialog and improve the dialog appearance.

Fix the max values for the options speedup_throttle and speedup_frame_skip.

Add base class for numeric OptionProxy types implementing the +=, -= operators, increment/decrement operators and operator T().

Move Speedup/Turbo dialog code from guiinit.cpp to a dialog class.

Remove the "GL VIEWPORT" debug prints from panel.cpp.

Fix #865


src/wx/dialogs/speedup-config.h
```diff
@@ -0,0 +1,32 @@
...
+    void ToggleSpeedupMute();
+    void SetSpeedupThrottle(wxCommandEvent& evt);
+
+private:
+    // The constructor is private so initialization has to be done via the
+    // static method. This is because this class is destroyed when its
+    // owner, `parent` is destroyed. This prevents accidental deletion.
+    SpeedupConfig(wxWindow* parent);
+
+    bool prev_frame_skip_cb_ = OPTION(kPrefSpeedupThrottleFrameSkip);
```
Steelskin, line 26: Please, move the instantiation to the constructor implementation. It's fine to use them in the definition for simple constructors but this is a complex instantiation.

src/wx/dialogs/joypad-config.cpp
```diff
@@ -80,6 +80,7 @@ JoypadConfig* JoypadConfig::NewInstance(wxWindow* parent,
 
 JoypadConfig::JoypadConfig(wxWindow* parent, const config::BindingsProvider bindings_provider)
     : BaseDialog(parent, "JoypadConfig"), bindings_provider_(bindings_provider) {
+
```
Steelskin, line 83: nit: revert this

src/wx/dialogs/speedup-config.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+    this->Bind(wxEVT_SPIN_UP,   std::bind(&SpeedupConfig::SetSpeedupThrottle, this, _1), XRCID("SpeedupThrottleSpin"));
+    this->Bind(wxEVT_SPIN_DOWN, std::bind(&SpeedupConfig::SetSpeedupThrottle, this, _1), XRCID("SpeedupThrottleSpin"));
+    this->Bind(wxEVT_SPIN,      std::bind(&SpeedupConfig::SetSpeedupThrottle, this, _1), XRCID("SpeedupThrottleSpin"));
+    this->Bind(wxEVT_TEXT,      std::bind(&SpeedupConfig::SetSpeedupThrottle, this, _1), XRCID("SpeedupThrottleSpin"));
+
+    wxSpinCtrl* speedup_throttle_spin = GetValidatedChild<wxSpinCtrl>("SpeedupThrottleSpin");
+    wxCheckBox* frame_skip_cb         = GetValidatedChild<wxCheckBox>("SpeedupThrottleFrameSkip");
+
+    if (OPTION(kPrefSpeedupFrameSkip) != 0) {
```
Steelskin, line 39: nit: Here, and below, store the value when you first access it or you end up doing a lookup every time.

src/wx/dialogs/speedup-config.cpp
```diff
@@ -0,0 +1,133 @@
...
+        speedup_throttle_spin->SetValue(OPTION(kPrefSpeedupFrameSkip) * 100);
+        prev_throttle_spin_ = OPTION(kPrefSpeedupFrameSkip) * 100;
+
+        frame_skip_cb->SetValue(true);
+        frame_skip_cb->Disable();
+    }
+    else {
```
Steelskin, line 46: My style is to put the else on the same line as the closing bracket above, but I don't feel very strongly about it :)

src/wx/dialogs/speedup-config.cpp
```diff
@@ -0,0 +1,133 @@
...
+    this->Bind(wxEVT_SPIN_UP,   std::bind(&SpeedupConfig::SetSpeedupThrottle, this, _1), XRCID("SpeedupThrottleSpin"));
+    this->Bind(wxEVT_SPIN_DOWN, std::bind(&SpeedupConfig::SetSpeedupThrottle, this, _1), XRCID("SpeedupThrottleSpin"));
+    this->Bind(wxEVT_SPIN,      std::bind(&SpeedupConfig::SetSpeedupThrottle, this, _1), XRCID("SpeedupThrottleSpin"));
+    this->Bind(wxEVT_TEXT,      std::bind(&SpeedupConfig::SetSpeedupThrottle, this, _1), XRCID("SpeedupThrottleSpin"));
+
+    wxSpinCtrl* speedup_throttle_spin = GetValidatedChild<wxSpinCtrl>("SpeedupThrottleSpin");
+    wxCheckBox* frame_skip_cb         = GetValidatedChild<wxCheckBox>("SpeedupThrottleFrameSkip");
+
+    if (OPTION(kPrefSpeedupFrameSkip) != 0) {
+        speedup_throttle_spin->SetValue(OPTION(kPrefSpeedupFrameSkip) * 100);
```
Steelskin, line 40: Please, add a comment at the top of this block explaining what `OPTION(kPrefSpeedupFrameSkip)` implies. I am guessing it means speedup frame skip disabled?

src/wx/guiinit.cpp
```diff
@@ -1483,92 +1484,6 @@ static class ThrottleCtrl_t : public wxEvtHandler {
     }
 } throttle_ctrl;
 
-static class SpeedupThrottleCtrl_t : public wxEvtHandler {
```
Steelskin, line 1486: Thanks for moving more of the dialog initialization code!

src/wx/dialogs/speedup-config.cpp
```diff
@@ -0,0 +1,133 @@
...
+    speedup_throttle_spin->SetWindowStyleFlag(wxTE_PROCESS_ENTER);
+#endif
+
+    speedup_throttle_spin->Refresh();
+
+    this->Fit();
+    this->Refresh();
+}
+
+void SpeedupConfig::SetSpeedupThrottle(wxCommandEvent& evt)
```
Steelskin, line 70: I would convert this to a validator but it can wait. The issue here is that the changes are going to take effect immediately, before the user hits "OK"

Steelskin (commented):
Thank you, this looks good to me. I skimmed over the core changes, they look fine. We'll probably change these down the line but it's a huge improvement on the UI side!

src/wx/config/internal/option-internal.cpp
```diff
@@ -328,9 +328,10 @@ std::array<Option, kNbOptions>& Option::All() {
         Option(OptionID::kPrefSkipSaveGameCheats, &coreOptions.skipSaveGameCheats, 0, 1),
         Option(OptionID::kPrefSkipSaveGameBattery, &coreOptions.skipSaveGameBattery, 0, 1),
         Option(OptionID::kPrefThrottle, &coreOptions.throttle, 0, 450),
-        Option(OptionID::kPrefSpeedupThrottle, &coreOptions.speedup_throttle, 0, 3000),
-        Option(OptionID::kPrefSpeedupFrameSkip, &coreOptions.speedup_frame_skip, 0, 300),
+        Option(OptionID::kPrefSpeedupThrottle, &coreOptions.speedup_throttle, 0, 450),
+        Option(OptionID::kPrefSpeedupFrameSkip, &coreOptions.speedup_frame_skip, 0, 40),
```
Steelskin, line 332: This is minor and I doubt anyone is going to be hit by this but since you're lowering the higher bound here, could you also silently force-update the value if it's above the higher bound here: https://github.com/visualboyadvance-m/visualboyadvance-m/blob/7a0826a60ccec45bfdd0d92070bb1d8fb783a0e6/src/wx/opts.cpp#L255?
We need a better way of handling this but this will do for now.

Steelskin (approved):
Thanks, lgtm!
I have a minor comment but feel free to discard it. Maybe it's better that the user be informed that their value changed? I'll let you decide what you feel is best.

src/wx/dialogs/speedup-config.cpp
```diff
@@ -0,0 +1,154 @@
...
+// The config option SpeedupThrottle is the throttle value, if frame skip is
+// used it is 0. The config option SpeedupFrameSkip is the number of frames to
+// skip, if throttle is used it is 0. The config option SpeedupThrottleFrameSkip
+// represents the Frame Skip checkbox for values under 450, that is if they are
+// interpreted as a frame skip or a throttle.
+
+namespace dialogs {
+
+bool SpeedupConfigValidator::TransferToWindow() {
+    unsigned opt_frame_skip           = OPTION(kPrefSpeedupFrameSkip);
```
Steelskin, line 28: nit: uint32_t (and the line below too)

Steelskin (approved):
Thanks, lgtm!

PR #1330, opened 2024-09-07
Add convenience functions for UTF16 conversions

Steelskin:


src/wx/config/strutils.cpp
```diff
@@ -41,40 +60,53 @@ wxArrayString str_split_with_sep(const wxString& text, const wxString& sep)
...
-            out.push_back(c);
-        } else if (c < 0x800) {
-            out.push_back(0xC0 | (c >> 6));
-            out.push_back(0x80 | (c & 0x3F));
-        } else if (c < 0xD800 || c >= 0xE000) {
-            // Regular 3-byte UTF-8 character.
-            out.push_back(0xE0 | (c >> 12));
-            out.push_back(0x80 | ((c >> 6) & 0x3F));
-            out.push_back(0x80 | (c & 0x3F));
+std::vector<uint8_t> utf16_to_utf8_vector(char16_t* utf16) {
```
rkitover, line 63: Make the return type `std::string` instead of `std::vector`, it can do the same things you are doing here and that makes more sense and is easier to use, and will make the code here a bit cleaner.

I see why you said you don't like the iconv API very much, we could have used the simple version of the API without buffer resizing (because we only need small strings) and error handling, but since you already did all of this we may as well keep it!

src/wx/config/strutils-test.cpp
```diff
@@ -118,10 +118,25 @@ TEST(UTF16ToUTF8Test, MultiByte) {
...
+    auto vec = config::utf16_to_utf8_vector(utf16.data());
+
+    ASSERT_EQ(vec.size(), 6);
+
+    EXPECT_EQ(vec[0], 0xE2);
+    EXPECT_EQ(vec[1], 0x82);
+    EXPECT_EQ(vec[2], 0xAC);
+    EXPECT_EQ(vec[3], 0xE2);
+    EXPECT_EQ(vec[4], 0x82);
+    EXPECT_EQ(vec[5], 0xAC);
```
rkitover, line 133: Assuming you made the return type `std::string` (see below), you can just do this here:

```cpp
auto str = config::utf16_to_utf8_str(utf16);

ASSERT_EQ(str.size(), 6);

EXPECT_EQ(str, "LITERAL-UTF8-HERE");
```
. No need to even add a BOM to the source file, this will work fine with everything.

rkitover (changes requested):
I left some comments on the commit, nice work!

src/wx/audio/internal/faudio.cpp
```diff
@@ -48,10 +48,7 @@ int FAGetDev(FAudio* fa) {
         if (hr != 0) {
             continue;
         }
-        const std::vector<uint8_t> device_id_u8 =
-            config::utf16_to_utf8(reinterpret_cast<uint16_t*>(dd.DeviceID));
-        const wxString device_id = wxString::FromUTF8(
-            reinterpret_cast<const char*>(device_id_u8.data()), device_id_u8.size());
+        const wxString device_id = config::utf16_to_utf8(reinterpret_cast<char16_t*>(dd.DeviceID));
```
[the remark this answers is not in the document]
Steelskin: Unfortunately, this type of constraint requires C++20. However, since there are really only 2 types, I added 2 functions..

src/wx/config/strutils.cpp
Steelskin: This is a bit pedantic but bear with me. With this function, I want to be able to check the individual code units of the output for testing purposes. While we can use a `std::string` for that purpose, there is some degree of semantic abuse as a string really stores some `char`, which are not necessarily 8-bits and whose signedness is implementation defined (thank you very much, C standard). Furthermore, after double-checking, `char16_t` is not guaranteed to be 16-bits (thank you very much, C++ standard, AGAIN).
This function only really exists for testing purposes and here what we are testing is "are the code units valid?". So since we are exclusively considering code units here, I believe this function should really be taking a `uint16_t*` and return a vector of `uint8_t`.

src/wx/config/strutils-test.cpp
```diff
@@ -108,8 +108,8 @@ TEST(UTF16ToUTF8Test, Basic) {
 
 TEST(UTF16ToUTF8Test, MultiByte) {
     // U+20AC EURO SIGN.
-    std::vector<uint16_t> utf16 = {0x20AC, 0};
-    auto vec = config::utf16_to_utf8(utf16.data());
+    std::vector<char16_t> utf16 = {0x20AC, 0};
```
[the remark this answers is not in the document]
Steelskin: Sadly, we can not guarantee that `char16_t` is 16 bits long (see other comment)

rkitover (approved):

PR #1351, opened 2024-09-29
Ignore illegal opcodes used by Wii U VC

Steelskin:
The Wii U VC emulator incorrectly allows for armv6 thumb instructions (like LSRS). These cause an undefined instruction error on real hardware but the Wii U VC emulator just ignores them. This change mimicks the Wii U VC emulator behavior.


Squall-Leonhart:
Does the trick, no more freeze

Squall-Leonhart (approved):

PR #1370, opened 2024-11-14
[GB] Add support for per-game overrides

### LINKED ISSUE #1368: [Bug]: alleyway not work (key left, right not work)
### Expected behavior?

the left button had to move the sprite to the left, and the right button to the right.

### Actual behavior?

the button left and right not work.

in the emulator (mgba ) : alleyway is work fine

1.start alleyway
2. the start button and A  work, left and right not work
3. 
4.
5.
...

### Error messages?

no message errors

### What game were you playing

title : Alleyway

### What settings are you using?

nvidia 535.183.01

VBA-M: version 2.1.9-67e4944c
OS: linux
If using Linux, specify if using xorg or Wayland: xorg
CPU: Intel(R) Core(TM) i5-10300H CPU @ 2.50GHz
GPU:  nvidia 535.183.01 geforce gtx1650ti
RAM: 32 go

### Copyright Compliance

Steelskin:
Alleyway is a buggy game that does not work properly when a Game Boy Printer is connected. In order to work around this issue, this adds upport for built-in per-software Game Boy overrides. In addition, this renames various variables to make their meaning clearer.

* This only supports built-in overrides. External INI files are not supported.
* Only the Game Boy Printer option is supported, this only takes effect on game startup and the general configuration option is restored when the game is unloaded.
* As a result, it is possible to override the per-game override by manually setting the option while the game is running, this is working as intended.
* Future refactors of the option handling will manage overrides better.

Fixes #1368

### DIFF
--- a/src/core/base/system.h
+++ b/src/core/base/system.h
@@ -70,7 +70,7 @@ extern struct CoreOptions {
     int skipSaveGameBattery = 1;
     int skipSaveGameCheats = 0;
     int useBios = 0;
-    int winGbPrinterEnabled = 1;
+    int gbPrinterEnabled = 1;
     uint32_t speedup_throttle = 100;
     uint32_t speedup_frame_skip = 9;
     uint32_t throttle = 100;
--- a/src/core/base/system.h
+++ b/src/core/base/system.h
@@ -70,7 +70,7 @@ extern struct CoreOptions {
     int skipSaveGameBattery = 1;
     int skipSaveGameCheats = 0;
     int useBios = 0;
-    int gbPrinterEnabled = 1;
+    int gbPrinterEnabled = 0;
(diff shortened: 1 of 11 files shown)


rkitover:
This is great, but can you also please set the default to off and also set the configured value to off once based on a version check for previous versions.

Steelskin:
> This is great, but can you also please set the default to off and also set the configured value to off once based on a version check for previous versions.

I changed the default back to off but I don't think we should automatically toggle the existing value. For the vast majority of software, there is no downside to having the GB Printer option enabled and it is possible that many of our users turned it on on purpose

rkitover:
Remember when we switched LCD filter to off, you made the config versioned so that it only switched off on config version upgrade. That's what I meant.

PR #1371, opened 2024-11-14
[build] Disable link on MinGW

Steelskin:
A recent libc++ update started enforcing correct type use for char_traits, breaking the SFML headers. Temporarily disable the Link option by default on MinGW to work around this issue.


ZachBacon:
Is this in regards to llvm's libc++ or libstdc++?

Steelskin:
> Is this in regards to llvm's libc++ or libstdc++?

It's libc++. Anyway it looks like there is another issue. I'll take another look this weekend.

PR #1373, opened 2024-11-24
[gtest] Use googletest as a submodule

Steelskin:
Some Linux distributions use a hermetic build environment to build. This resulted in issues with `FetchContent` attempting to download googletest at configure time.
This fixes the issue by integrating googletest as a git submodule instead of downloading it at configure time.


rkitover:
Why is 3eea90af not sufficient for this? I had already fixed this.

PR #1222, opened 2024-02-14
Update visualboyadvance-m.metainfo.xml

jhonny-oliveira:
Closes #1221.

PR #1397, opened 2025-01-09
Re-add developer to metainfo

PunkPangolin:
That seems to have gotten accidentally deleted by #1222


rkitover:
Merged manually, thanks!

Issue #1407, opened 2025-02-02
Sound Volume Reset Bug in Version 2.1.11


tsuyaqndii:
### Expected behavior?

The sound volume should remain at 5% as set before closing the emulator.

### Actual behavior?

The sound volume setting does not persist correctly after restarting the emulator. The volume appears to reset to the default level. Although when using the "Increase Volume" option, the volume jumps to 10% instead of increasing from the default. This indicates that the emulator internally remembered the 5% setting, but did not display it correctly. This bug does not occur in version 2.1.5, where the volume settings persist as expected.

### Steps to reproduce the problem

1.Open VisualBoyAdvance-M 2.1.11.
2.Load any ROM. (I used Pokemon Sapphire and Emerald)
3.Adjust the sound volume to 5% using the volume settings.
4.Close the emulator.
5.Reopen the emulator and load the ROM again.

### Error messages?

If this is a crash bug, and you have the time and inclination, please provide a symbolic backtrace.
Instructions for doing so are [here](https://github.com/visualboyadvance-m/visualboyadvance-m/blob/master/README.md#reporting-crash-bugs).

### What game were you playing

<Some of this information can be provided with a screenshot of Rom Information, SAVE TYPE for GBA titles can be obtained from Options > Game Boy Advance > Config>
Title: Pokemon Sapphire and Emerald
GB, GBC or GBA: GBA
ROM CODE: 
BIOS Used?: (Share crc hash)
SAVE TYPE: (Attach Battery or Savestate here)
Clean, Intro'd or Hacked?:(Attach Patch here)

### What settings are you using?

Video driver (simple, opengl):
Sound driver (directsound, xaudio, openal, SDL):
Scale Filter Used?:

### Specifications and Build details

VBA-M: (include commit version number and whether you are using the 32-bit or 64-bit version)
OS: (include version and OS build number using WinVer if on Windows):
If using Linux, specify if using xorg or Wayland:
CPU:
GPU:
RAM:

### Copyright Compliance

- [x] There are no infringing files attached to this bug report.

PR #1408, opened 2025-02-06
Set volume level after closing SoundConfigure dialog

jaysn:
tried to fix this issue (#1407) . the soundSetVolume() got removed (accidentally?) - this should fix the issue at least. even though it could use the soundVolume variable directly.
haven't been in c for some time and can't compile on my windows machine but it compiled on my debian 12.9 x86_64 system using the provided deps from installdeps.

### DIFF
--- a/src/wx/cmdevents.cpp
+++ b/src/wx/cmdevents.cpp
@@ -2161,6 +2161,8 @@ EVT_HANDLER_MASK(SoundConfigure, "Sound options...", CMDEN_NREC_ANY)
     gb_effects_config.echo = (float)OPTION(kSoundGBEcho) / 100.0;
     gb_effects_config.stereo = (float)OPTION(kSoundGBStereo) / 100.0;
     soundFiltering = (float)OPTION(kSoundGBAFiltering) / 100.0f;
+    // Missing Setup for volume here
+    soundSetVolume((float)OPTION(kSoundVolume) / 100.0f);
 }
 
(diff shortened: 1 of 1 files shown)


rkitover:
Can you write a [commit message](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html) please?

jaysn:
Sure 👍🏻 - please let me know if this is good and explaining enough 😄

Steelskin:
I believe this is not fixing the issue you linked. This will explicitly call `soundSetVolume()` after closing the `SoundConfigure` dialog, which is already done via that dialog's validator [here](https://github.com/visualboyadvance-m/visualboyadvance-m/blob/2ab707c0ca68a558a57436dc901867b797d79b2e/src/wx/dialogs/sound-config.cpp#L147). This is then handled by the volume observer set up [here](https://github.com/visualboyadvance-m/visualboyadvance-m/blob/2ab707c0ca68a558a57436dc901867b797d79b2e/src/wx/panel.cpp#L163), the handler for that observer is [here](https://github.com/visualboyadvance-m/visualboyadvance-m/blob/2ab707c0ca68a558a57436dc901867b797d79b2e/src/wx/panel.cpp#L2703).

It seems what we are missing here is a call to `soundSetVolume()` on startup. I am going to look into the issue.

Steelskin:
I am closing this in favor of #1410, which should fix the issue. Thanks a lot for bringing this to our attention, though!

PR #1422, opened 2025-03-13
reallocate the updated size instead of the constant size

### LINKED ISSUE #1421: Is it a bug here? It reallocates with a wrong size.
As title.

https://github.com/visualboyadvance-m/visualboyadvance-m/blob/5b4559f4ee241a6fc64af21d69e75445675243a7/src/core/gba/gba.cpp#L484

@Steelskin @negativeExponent

Squall-Leonhart:
closes #1421

### DIFF
--- a/src/core/gba/gba.cpp
+++ b/src/core/gba/gba.cpp
@@ -481,7 +481,7 @@ void gbaUpdateRomSize(int size)
     if (size > romSize) {
         romSize = size;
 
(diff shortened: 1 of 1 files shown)


Steelskin (approved):
LGTM on principle but please give me a few days to double-check the calling sites.

Steelskin:
Looking at the one call site, it looks like the resizing is already done in `applyPatch` but this can't hurt.

PR #1431, opened 2025-03-29
SDL3 port of VisualBoyAdvance-m

andyvand:
I've fixed the deprecations.
I've ported it to SDL3.

### DIFF
--- a/src/core/base/file_util_desktop.cpp
+++ b/src/core/base/file_util_desktop.cpp
@@ -155,7 +155,7 @@ IMAGE_TYPE utilFindType(const char* file) {
 
 void utilStripDoubleExtension(const char* file, char* buffer) {
     if (buffer != file)  // allows conversion in place
-        strcpy(buffer, file);
+        strncpy(buffer, file, strlen(file));
 
     if (utilIsGzipFile(file)) {
         char* p = strrchr(buffer, '.');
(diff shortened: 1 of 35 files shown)


src/core/base/file_util_desktop.cpp
```diff
@@ -155,7 +155,7 @@ IMAGE_TYPE utilFindType(const char* file) {
 
 void utilStripDoubleExtension(const char* file, char* buffer) {
     if (buffer != file)  // allows conversion in place
-        strcpy(buffer, file);
+        strncpy(buffer, file, strlen(file));
```
rofl0r, line 158: this change is totally bogus. it doesn't make sure that file fits into buffer, since the buffer size isn't passed in. instead, it just makes it slower and NOT zero terminated. because strncpy doesn't do what you think it does.
it pads the buffer with 0s, and when the strlen equals buffer size, the resulting string won't get zero-terminated (see man 3p strncpy). that means functions working upon the so-filled string will run out of bounds when looking for the zero terminator.
muting these MSVC warnings is actively harmful in this case.

ZachBacon: Honestly I would leave the deprecations alone for another pull request. Just do the sdl3 port. @andyvand are you able to redo this and just leave with the deprecations for a separate pull request?

andyvand: I'll use my second GitHub account.
I can do that.

PR #1436, opened 2025-04-04
build: support building ARM64 binary for Mac

rkitover:
Make some adjustments to the Mac build scripts to support building an ARM64 binary for Apple Silicon.

Also support cross-compiling for Intel on Apple Silicon.

Also prefer using Ninja for CMake ports.


tools/builder/MINGW.sh
```diff
@@ -263,10 +263,10 @@ zlib_dist=$(table_line DISTS zlib-target)
 
 table_line_remove  DISTS zlib-target
 
-table_insert_after DISTS cmake "zlib-target $zlib_dist"
+table_insert_after DISTS curl "zlib-target $zlib_dist"
```
Steelskin, line 266: Are these changes related?

tools/macOS/builder
```diff
@@ -5,45 +5,79 @@ set -e
...
+        target_cpu=i386
         shift
         ;;
 esac
 
+if [ -n "$APPLE_SILICON" ]; then
+    if [ -n "$intel_target" ]; then
+        target_build_arch='-target x86_64-apple-macos10.15 -march=core2 -mtune=coffeelake'
+        target_cpu=x86_64
+#        export MACOSX_DEPLOYMENT_TARGET=10.10 # Yosemite
```
Steelskin, line 59: nit: Please, remove the commented line.

tools/macOS/builder
```diff
@@ -71,19 +105,19 @@ best_llvm=$(
...
 [ -n "$best_llvm" ] && export PATH="$best_llvm/bin:\$PATH"
 
 export CC=clang
 export CXX=clang++
 export CPPFLAGS="-DICONV_CONST="
-export CFLAGS="-m${target_bits} -framework Carbon -framework Foundation -framework CoreServices -Wno-unused-command-line-argument -DICONV_CONST= -Wl,-no_compact_unwind"
-export CXXFLAGS="-m${target_bits} -framework Carbon -framework Foundation -framework CoreServices -Wno-unused-command-line-argument -DICONV_CONST= -Wl,-no_compact_unwind"
-export OBJCXXFLAGS="-m${target_bits} -framework Carbon -framework Foundation -framework CoreServices -Wno-unused-command-line-argument -DICONV_CONST= -Wl,-no_compact_unwind"
-export LDFLAGS="-m${target_bits} -framework Carbon -framework Foundation -framework CoreServices -Wno-unused-command-line-argument -Wl,-no_compact_unwind"
+export CFLAGS="$target_build_arch -framework Carbon -framework Foundation -framework CoreServices -framework SystemConfiguration -Wno-unused-command-line-argument -DICONV_CONST= -Wl,-no_compact_unwind"
```
Steelskin, line 117: I think a better way of doing this would be to set the [`CMAKE_OSX_ARCHITECTURES`](https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_ARCHITECTURES.html) to `arm64` or `x86_64` and let CMake handle the rest. WDYT?

tools/builder/MINGW.sh
rkitover: Yes, because I reordered the base dists list.

tools/macOS/builder
rkitover: Done.

PR #1443, opened 2025-04-23
BUILDER: Updated core.sh source packages required to build vba-m

ZachBacon:
This just updates a good portion of the core packages that the build scripts use to build against.

### DIFF
--- a/tools/builder/core.sh
+++ b/tools/builder/core.sh
@@ -109,62 +109,62 @@ PRE_BUILD_DISTS="$PRE_BUILD_DISTS bzip2 xz unzip"
 
 DISTS=$DISTS'
     bzip2           ftp://sourceware.org/pub/bzip2/bzip2-1.0.8.tar.gz                                           lib/libbz2.a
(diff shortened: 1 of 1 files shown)


tools/macOS/builder
```diff
@@ -222,6 +222,7 @@ table_line_append DIST_PRE_BUILD ffmpeg "sed -i.bak 's/-lstdc++/-lc++/g' configu
 # For a 10.7 target:
 #table_line_replace DISTS wxwidgets "https://github.com/wxWidgets/wxWidgets/releases/download/v3.0.5.1/wxWidgets-3.0.5.1.tar.bz2 lib/libwx_baseu-3.0*.a"
 
+table_line_append DIST_ARGS sdl3  "-DCMAKE_C_FLAGS=-fno-fast-math"
```
rkitover, line 225: This would be better in `DIST_EXTRA_CFLAGS` near `DIST_EXTRA_CXXFLAGS` in `core.sh`.

ZachBacon: I'll apply that after my nap

PR #1478, opened 2025-07-15
Fix JP Translate

shotadft:
I changed my name to my current name because my real name was registered by mistake.


rkitover:
The problem is that these files are downloaded from Transifex, so your change would get overwritten.

Is there any way you can make this change on Transifex?

shotadft:
My account in the past is no longer accessible due to the termination of the facebock linkage service!

rkitover:
Ok, I will look into this and get back to you.

rkitover:
So, we already have `ja` which is 60% translated, while `ja_JP` is 0% translated, so I deleted it from transifex and the repo.

PR #1479, opened 2025-07-15
Several fixes for windows

Squall-Leonhart:
improve on the fex type check, so it works on windows x86/64 
move BindAppIcon from BindControls in guiinit.cpp to wxvbam.cpp to avoid a timing issue that see's the icon not populated on windows till late. 
adds 3 more icon sizes to the icon on windows.


src/wx/wxvbam.cpp
```diff
@@ -614,6 +614,8 @@ bool wxvbamApp::OnInit() {
 
     frame->Show(true);
 
+    frame->BindAppIcon();
```
Steelskin, line 617: Please, add a comment to document why this is done here.

Steelskin (approved):

PR #1492, opened 2025-09-17
Apply bilinear option for SDL texture

rkitover:
Set the texture option for bilinear or nearest on the SDL texture depending on the value of the bilinear checkbox.


Squall-Leonhart:
looks good to me!

Bilinear on

image

Bilinear off
image

PR #1496, opened 2025-09-25
Fixes for BSD build

brad0:
- patch fixes for BSD handling
- OpenBSD/aarch64 has elf_aux_info()
- unrar build fix for FreeBSD/NetBSD/OpenBSD
- OpenBSD does not have librt


Squall-Leonhart:
looks ok to me

rkitover:
Merged manually with a minor change, no need to interpolate CMake variables in an `IF()` statement.

Thanks!

brad0:
> no need to interpolate CMake variables in an `IF()` statement.

Most projects using CMake (after looking at a couple hundred) I see like this and wasn't sure why or which is appropriate.

Issue #784, opened 2020-11-29
Outstanding ELF vuln reported several months ago still crashes


endrift:
I made this ELF in January, reported it privately a number of times, and it's still been ignored endlessly. Responsible disclosure periods or anything of the sort expired months ago.

### Expected behavior

VBA-M should not crash when you open a file.

### Actual behavior

VBA-M crashes when running this ELF file I crafted months ago.

### Steps to reproduce the behavior

Open the attached ELF file.

### ROM details

[vbam-vuln.zip](https://github.com/visualboyadvance-m/visualboyadvance-m/files/5611723/vbam-vuln.zip)

### Option details

N/A

### Build details

Operating System (Windows, Mac, Linux (state distribution), etc.): Doesn't matter

If using Linux, specify if using xorg or Wayland: Doesn't matter

Version of code (pre-built binary version, or commit ref, or just "master"): dfa97d8a916fddd4a0fe0b34a07c44abec29d78a, but this issue has existed for a very long time.

rkitover:
We did not purposefully ignore you, it's just that no one has fixed this yet.

Your work on this is very much appreciated.

I will make no claim that this software is secure and people should treat it as such.

To fix these issues will require a thorough refactor of the ELF loader etc., and I just haven't gotten around to it yet, and it's by far not the highest priority.

endrift:
Pretty sure you could fix this with better bounds checking without needing refactoring.

rkitover:
Sure, but there are lots of places that need bounds checking, so it would be changing a lot of code, and the code is horrible and would benefit from a complete rewrite.

easyaspi314:
Unfortunately, @rkitover is right. This code reeks, and needs a rewrite desperately.

I haven't tried that file yet (I'd be more willing to try it if it was a more minimal example), but looking at `elf.cpp`, it seems that the patch for #255

https://github.com/visualboyadvance-m/visualboyadvance-m/blob/bb903e5e08846363c07eb59a7021f4285ec5d372/src/gba/elf.cpp#L2555-L2590

Is missing from this very similar loop:

https://github.com/visualboyadvance-m/visualboyadvance-m/blob/bb903e5e08846363c07eb59a7021f4285ec5d372/src/gba/elf.cpp#L2620-L2641

However, that patch is also bugged: It ignores the error instead of rejecting the file.

[...]

PR #1518, opened 2025-11-25
gbaElf.cpp: Bound checks

ZachBacon:
This should hopefully resolve some of the elf loader vulnerabilities worked on this with help with claude, so another set of eyes should be nice to look at this.

Should solve [#784](https://github.com/visualboyadvance-m/visualboyadvance-m/issues/784)

### DIFF
--- a/src/core/gba/gbaElf.cpp
+++ b/src/core/gba/gbaElf.cpp
@@ -12,6 +12,51 @@
 #define elfReadMemory(addr) \
     READ32LE((&map[(addr) >> 24].address[(addr)&map[(addr) >> 24].mask]))
 
+// Global bounds tracking for ELF file data validation
+static uint8_t* g_elfFileStart = NULL;
+static uint8_t* g_elfFileEnd = NULL;
+
+// Helper function to check if a pointer is within valid ELF file bounds
+static inline bool elfIsValidPtr(uint8_t* ptr, size_t size) {
+    if (g_elfFileStart == NULL || g_elfFileEnd == NULL)
+        return true; // Bounds checking not initialized yet
+    if (ptr+0 < g_elfFileStart || ptr+size >= g_elfFileEnd)
+        return false;
+    return true;
+}
+
+// Forward declarations for read functions
+uint32_t elfRead4Bytes(uint8_t*);
+uint16_t elfRead2Bytes(uint8_t*);
+
+// Safe read wrappers with automatic bounds checking
+static inline bool elfSafeRead4Bytes(uint8_t* data, uint32_t* out) {
+    if (!elfIsValidPtr(data, 4)) {
+        fprintf(stderr, "Error: 4-byte read out of bounds\n");
+        return false;
+    }
+    *out = elfRead4Bytes(data);
+    return true;
+}
+
+static inline bool elfSafeRead2Bytes(uint8_t* data, uint16_t* out) {
+    if (!elfIsValidPtr(data, 2)) {
+        fprintf(stderr, "Error: 2-byte read out of bounds\n");
+        return false;
+    }
+    *out = elfRead2Bytes(data);
+    return true;
+}
+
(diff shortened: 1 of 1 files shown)


rkitover:
Sorry I broke the CI, for some reason I did not get that error, I will fix it and rerun.

I think it would be better to use the library from elfutils to load ELF files instead of reimplementing it, this is a lot of code.

But I'm fine with merging this if it works for now.

As @Squall-Leonhart explained to me, we are using ELF to load GBA ELF executables, not ROMs.

Can you please do some testing with various ELF files that you can find and make sure this works correctly?

Squall-Leonhart:
Nach had some nitpicks about this PR, if you can check discord for them Zach.

ZachBacon:
> Nach had some nitpicks about this PR, if you can check discord for them Zach.

I already took them to heart and applied them. Right before you commented

ZachBacon:
Sadly this still doesn't quite solve the issue currently in 728

ZachBacon:
So as is, with that carefully crafted elf file @endrift made, it segfaults vba-m. I'm working on some code at the moment, haven't commited it. It still segfaults, BUT, it still sadly closes. So I'm going to draft this for now so I can look into this further.

src/core/gba/gbaElf.cpp
```diff
@@ -12,6 +12,51 @@
...
     READ32LE((&map[(addr) >> 24].address[(addr)&map[(addr) >> 24].mask]))
 
+// Global bounds tracking for ELF file data validation
+static uint8_t* g_elfFileStart = NULL;
+static uint8_t* g_elfFileEnd = NULL;
+
+// Helper function to check if a pointer is within valid ELF file bounds
+static inline bool elfIsValidPtr(uint8_t* ptr, size_t size) {
+    if (g_elfFileStart == NULL || g_elfFileEnd == NULL)
+        return true; // Bounds checking not initialized yet
```
Steelskin: Any reason not to return false here? It feels like this should be invalid.

src/core/gba/gbaElf.cpp
```diff
@@ -2639,18 +3386,38 @@ bool elfReadProgram(ELFHeader* eh, uint8_t* data, unsigned long data_size, int&
...
         //   sh[_i]->link, sh[_i]->info);
         if (READ32LE(&sh[_i]->flags) & 2) { // load section
+            uint32_t sect_addr = READ32LE(&sh[_i]->addr);
+            uint32_t sect_offset = READ32LE(&sh[_i]->offset);
+            uint32_t sect_size = READ32LE(&sh[_i]->size);
+            
+            // Validate section data is within file bounds
+            if (sect_offset > data_size || sect_size > data_size || 
+                sect_offset + sect_size > data_size) {
+                fprintf(stderr, "Warning: Section %d data exceeds file bounds, skipping\n", _i);
```
Steelskin: Should we bail out?

src/core/gba/gbaElf.cpp
```diff
@@ -2044,14 +2594,35 @@ uint8_t* elfParseBlock(uint8_t* data, ELFAbbrev* abbrev, CompileUnit* unit,
...
     }
     return data;
 }
 
 void elfGetFunctionAttributes(CompileUnit* unit, uint32_t offset, Function* func)
 {
+    // Validate offset is within compile unit
+    if (offset > unit->length) {
+        fprintf(stderr, "Error: Function attribute offset %u exceeds unit length %u\n", offset, unit->length);
+        return;
```
Steelskin: For these kind of functions, I wonder if we should have a global `bool g_elfHasError`, and set that to true so we don't continue at the call site.

Steelskin (commented):
These changes all look fine to me, though I did not look too closely at the individual checks. We really should have an external library handling this, in addition to having a class to set an internal state.
Anyway, I just have a couple of suggestions as-is.

rkitover:
Andy fixed ELF support and the vuln with ELFIO.

PR #1543, opened 2026-02-02
Use serialization buffer for state management

Squall-Leonhart:
non-windows builds (actually, windows too it seems) don't play nice with the way this was originally done, a suggested change was to use the heap but the way it was written seemed like a possible performance issue when rewind was being used, so rework it as a static buffer instead that is allocated once and used as needed, then delete on de-init.


Squall-Leonhart:
@negativeExponent is this adequate?

PR #1548, opened 2026-03-12
Build system: allow building without wayland library installed #1547

mazes-80:
I think this patch should fix mandatory dependency on wayland.
It need way less work than the local patch I used for version 2.2.3

Maybe useful until GTK drops X11 support


rkitover:
Hi, sorry for taking a long time to get back to you.

This looks fine, but we need CMake support for this.

We need to detect when GTK is missing Wayland support and define this macro appropriately.

PR #1565, opened 2026-06-24
Add Game Boy Player SIO rumble support

illinx:
Emulate the Game Boy Player's SIO handshake and rumble protocol so that GBP-aware GBA titles can drive controller rumble based off of mGBA's implementation

- Automatically turn off when something else is using the link port to match real GBP behavior
- Use existing rumble implementation for playback

### DIFF
--- a/src/core/CMakeLists.txt
+++ b/src/core/CMakeLists.txt
@@ -124,6 +124,8 @@ if(ENABLE_LINK)
     target_sources(vbam-core
         PRIVATE
         gba/gbaLink.cpp
+        gba/gbpSio.cpp
+        gba/gbpSio.h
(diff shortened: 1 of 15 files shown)


Squall-Leonhart:
i don't think we should take this at the present time, SIO needs a thorough rework.

illinx:
yeah, I noticed that the big cycle accurate implementation came in while I was working on this--right now I just worked around it but one question I had was if I should think about basing it off the new implementation

Squall-Leonhart:
I started with basic scheduling but incorporated toggleable tracing in order to work towards removing cycle padding which currently exist between CPU IRQ's and the SIO, its about 7 cycles difference that vba-m currently fakes.

PR #1571, opened 2026-07-17
Validate SGB packet counts

acts-1631:
A malformed SGB packet stream with a zero packet count could continue past
the seven-packet receive buffer and corrupt adjacent emulator state.

This rejects zero counts and resets invalid packet receiver state before a
packet-buffer access can exceed its fixed capacity.


Squall-Leonhart:
This is not correct, you are resetting things that should not be.

Squall-Leonhart:
image
corrupted pokemon palette for instance

PR #1573, opened 2026-07-19
Bound RFU broadcast payloads

acts-1631:
Wireless Adapter command 0x16 stores game-controlled payload words in a six-word payload area without checking the count. Oversized transfers overwrite adjacent RFU state.

Store only the available payload words while continuing to consume the full declared transfer.


Squall-Leonhart:
if you're looking at RFU stuff, can you work out why we manage to expose the digimon racing EU wireless option which is supposed to be disabled/hidden on that region.

PR #1581, opened 2026-09-16
Dockerize The Build Step!

Melnee:
Changes:

In order to build, now you can simply run:

`docker compose up` 

and then execute the generated binary in the build folder.

Screenshot showing proof of succesful build in terminal:
(command run: docker compose up)
image

(End of command, successful exit 0):
image

Running the Binary:
image

Note: I apologize about the misleading branch name -- it's in reference to another feature I'm planning on building. If I should recreate the branch, named differently, let me know! Thanks and I hope y'all enjoy Docker!


rkitover:
I have working `./installdeps` for Ubuntu, what is the advantage of this? And why is it not using it?

rkitover:
I was actually going to work on using Docker/Podman to make a static binary at some point.

Melnee:
> I have working `./installdeps` for Ubuntu, what is the advantage of this? And why is it not using it?

The advantage of docker is avoiding having to install dependencies on the host machine at all : D It aligns with the philosophy behind docker's creation, where the containerized environment is agnostic  to any possible variatoins in the host machine

Melnee:
> I have working `./installdeps` for Ubuntu, what is the advantage of this? And why is it not using it?

oooooo ok I see, yes I'll just call ./installdeps within the dockerfile, that's a very good point

Melnee:
> I was actually going to work on using Docker/Podman to make a static binary at some point.

oh that sounds like a dream. maybe I could make a PR that attempts to do that, and you could take a look and see if it matches what you're thinking of ?

rkitover:
Another thing I didn't mention, the resulting binary from this Docker build only works on Ubuntu and possibly some other distributions that are similar enough, adding a different build path to save a few megabytes of installed dependencies makes this even less useful.

rkitover:
The static binary would use the vcpkg build, which links most of its libraries as static, but there are still some outstanding issues, like the few libraries that are not linked statically except Xorg/Wayland etc. which don't need to be linked statically. Plus embedding the translations. Just haven't gotten around to it yet.

rkitover:
Also, some guy makde an appimage wrapper, it's in the issues somewhere.

Melnee:
> Another thing I didn't mention, the resulting binary from this Docker build only works on Ubuntu and possibly some other distributions that are similar enough, adding a different build path to save a few megabytes of installed dependencies makes this even less useful.

In that case, it looks like the appimage direction is the better move, from what you're describing.

This docker compose logic was useful to me so that I could build and test quickly in a way I'm used to with apps (I love docker)

But the actual feature I want to build involves intercepting the APU channel register state and translating the note/frequency/volume data into MIDI out messages. I have that in a forked repo. If it's not something you/your team is interested in, no problem, and I just keep it as a fork, and use the feature for myself.

Melnee:
> > Another thing I didn't mention, the resulting binary from this Docker build only works on Ubuntu and possibly some other distributions that are similar enough, adding a different build path to save a few megabytes of installed dependencies makes this even less useful.
> 
> In that case, it looks like the appimage direction is the better move, from what you're describing.
> 
> This docker compose logic was useful to me so that I could build and test quickly in a way I'm used to with apps (I love docker)
> 
> But the actual feature I want to build involves intercepting the APU channel register state and translating the note/frequency/volume data into MIDI out messages. I have that in a forked repo. If it's not something you/your team is interested in, no problem, and I just keep it as a fork, and use the feature for myself.

And not to bombard your inbox, but in case you were interested, the goal of the [MIDI-out](https://github.com/Melnee/visualboyadvance-midi-out/pull/2) is this project which I'm demonstrating in this video, filmed yesterday:
https://github.com/user-attachments/assets/8e8779cb-091d-4943-92b3-e065f80b0173

Steinway Model M Grand Piano, year: 1911, satin finish, refurbished with a new sound board & actions, retrofitted with a MIDI-input-based player-piano system. It's playing the "sound" from Oracle of Seasons. Well, bye for now, and thanks for the feedback! :pray:

